PR #133, opened 2025-12-10
Add automatic retry algorithm to client

mcraveiro:
Implement exponential backoff retry algorithm for client connections with automatic reconnection after disconnect detection.

Key changes:
- Add retry_options struct with configurable max_attempts, base_delay, max_delay, jitter_factor, and auto_reconnect settings
- Add connection_state enum (disconnected, connecting, connected, reconnecting) replacing the simple connected_ bool
- Add connect_with_retry() and connect_with_retry_sync() methods for initial connection with retries
- Implement run_reconnect_loop() that spawns automatically when connection is lost and auto_reconnect is enabled
- Use exponential backoff with jitter to prevent thundering herd
- Requests fail immediately during reconnection state
- Disconnect callback only invoked after all retry attempts exhausted
- Sequence numbers and correlation IDs preserved across reconnections


mcraveiro (commented):
Looks good

PR #168, opened 2025-12-17
Add exit confirmation dialog

mcraveiro:
Ask user to confirm before exiting ORE Studio using MessageBoxHelper::question. Also connect ExitAction from the File menu to close and add dismiss icon.


mcraveiro (commented):
Looks good

PR #169, opened 2025-12-17
Improve test failure visibility in build output

mcraveiro:
Replace raw XML dump with a Python script that parses Catch2 XML output and provides a clear summary. The script shows pass/fail/skip counts and detailed information for any failing tests including assertion details. Returns non-zero exit code when tests fail, causing the CI build to fail.


mcraveiro (commented):
Looks good

PR #589, opened 2026-03-29
[qt] Remove badge_colors struct from ColorConstants.hpp (Phase 3 complete)

mcraveiro:
## Summary

- Inline the six `QColor` literals used in `ServiceDashboardMdiWindow` (heartbeat RAG status and grid status colours) — the only remaining consumer of the `badge_colors` struct
- Delete `badge_colors` from `ColorConstants.hpp` — all badge colours are now resolved via `BadgeCache::resolve()` at runtime

This completes Phase 3 of the badge system migration. `ColorConstants.hpp` remains in place for other constants used widely across the codebase.

### DIFF
--- a/projects/ores.qt/src/ServiceDashboardMdiWindow.cpp
+++ b/projects/ores.qt/src/ServiceDashboardMdiWindow.cpp
@@ -25,6 +25,7 @@
 #include <QFutureWatcher>
 #include <QPointer>
 #include <QHeaderView>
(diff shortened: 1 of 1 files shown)


mcraveiro (commented):
Updated in 0bd08e2e2 — the local `kStatus*` constants were unnecessary: `ColorConstants.hpp` already defines `color_constants::level_info/warn/trace/error` with the exact same RGB values. Switched to use those instead and added the missing `#include`.

PR #776, opened 2026-05-19
[doc] Add :eval never to list-presets sh block to fix site build

mcraveiro:
## Summary

- Adds `:eval never` to the `#+begin_src sh` block in `how_do_i_list_available_presets.org`
- The org-mode export pipeline runs Emacs in batch mode; without `:eval never`, Emacs prompts for confirmation before executing the sh block, which hangs CI with "End of file during parsing: Evaluate this sh code block (list-presets) on your system?"
- No other sh blocks in `doc/v2/recipes/` have the same issue (one apparent match is inside an escaped `#+begin_src org` example block and is not evaluated)

### DIFF
--- a/.build-site.el
+++ b/.build-site.el
@@ -85,6 +85,11 @@
 <link rel=\"stylesheet\" href=\"https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.10.0/styles/googlecode.min.css\" />
 <link rel=\"stylesheet\" href=\"https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.0.0-beta3/css/all.min.css\">")
 
+;; Disable babel evaluation during export. ob-shell is still loaded for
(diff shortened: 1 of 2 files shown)


mcraveiro (commented):
Addressed in the latest commit. Root cause was in `.build-site.el`: it loads `ob-shell` but never sets `org-export-use-babel nil`, so org tried to re-evaluate src blocks marked `:exports both` during publish and prompted stdin in batch mode.

Fix: add `(setq org-export-use-babel nil)` to the build script so the site always publishes whatever `#+RESULTS:` is already in git — no evaluation ever happens in CI. The per-block `:eval never` from the first commit is reverted; it was a workaround that would have broken interactive `C-c C-c` and the recipe-refresh workflow.

`:exports both` stays as-is and is intentional: the pre-captured results snapshot is exported alongside the code. Developers refresh it via the org-babel-refresh recipe and commit the updated `#+RESULTS:` back.

PR #796, opened 2026-05-22
[trading] Export parser symbol for Windows DLL build + codegen export.hpp support

mcraveiro:
## Summary

- Fixes Windows link error: `undefined symbol: ores::trading::service::config::parser::parse` when linking `ores.trading.service.tests.exe`
- Root cause: `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` was removed in the symbol visibility migration but `ores.trading.service` was never given `ORES_TRADING_SERVICE_EXPORT` annotations on `parser` and `host`
- Creates `export.hpp`, annotates `config::parser` with `ORES_TRADING_SERVICE_EXPORT`, adds `ORES_TRADING_SERVICE_LIBRARY` compile definition to the lib target
- Wires the same pattern into the codegen system (all four component CMake templates + `cpp_export.hpp.mustache` template + `profiles.json`) so new components are scaffolded correctly
- Updates the symbol visibility knowledge doc to document the service lib export requirement

## Test plan

- [ ] Windows clang-debug-ninja: `ores.trading.service.tests.exe` links without undefined symbol errors
- [ ] Linux/macOS builds unaffected (export macros expand to `__attribute__((visibility("default")))` on GCC/Clang)
- [ ] Codegen: running `component-service` profile generates `export.hpp`, `config/parser.hpp` with export macro, and `src/CMakeLists.txt` with `ORES_TRADING_SERVICE_LIBRARY`

### DIFF
--- a/doc/knowledge/external/shared_library_symbol_visibility.org
+++ b/doc/knowledge/external/shared_library_symbol_visibility.org
@@ -151,9 +151,23 @@ Types shared *across* Qt plugins must live in =ores.qt.api= and be
 exported from there using =ORES_QT_API=.  Plugin-internal types are
 never exported.
 
-** Service executables (=ores.X.service=)
+** Service libraries and executables (=ores.X.service=)
 
-Executables export nothing.  Do not add an export header.
+Each service project produces two targets: a shared library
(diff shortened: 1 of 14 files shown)


mcraveiro:
Recreating from main to get a clean diff.

PR #846, opened 2026-05-25
[client,doc] Instance colour is a window marker, not the name or a theme

mcraveiro:
## Summary

Untangle two conflations around the instance **colour**:

1. `build/scripts/start-client.sh` used the colour to set the **instance
   name** (`red` → "Red", etc.), shadowing the name from `.env`.
2. The user manual described `--colour` as a **light/dark theme** — there
   is no theme flag.

The colour is only a small **coloured circle in the status bar** so you
can tell multiple windows apart; the **name** is separate and comes from
`.env` (or `--name`).

## Changes

**`start-client.sh`**
- Removed the `INSTANCE_NAME="Red"/"Green"/"Blue"/<hex>` assignments from
  the colour cases. The colour now only sets `--instance-color` (and the
  PID/log tag); the instance name comes from `--name`, else
  `ORES_CHECKOUT_LABEL` from `.env`.
- Fixed the `--name` help text (default is now the `.env` label).

**Manual (chapter 1)**
- Rewrote the "Colour theme" section as "Telling windows apart (instance
  colour and name)": `--instance-color HEX` = status-bar circle,
  `--instance-name` = the label; explicitly *not* a theme. Fixed the two
  intro mentions of "colour theme".

(Verified against `CommandLineParser.cpp`: the real flags are
`--instance-name`/`-n` and `--instance-color`; there is no theme flag.)

## Notes

`validate_docs.sh` passes (90/90); `bash -n` clean.

While verifying, I noticed the manual's "Configuration directory" and
"Selecting a connection at startup" sections document `--config-dir` and
`--connection` flags that **the parser does not define** — flagging
separately; not touched here.


mcraveiro:
Closing — will reopen when ready (holding PRs until requested).

PR #850, opened 2026-05-25
[fix] Portfolios and books not visible when a workspace is active

mcraveiro:
## Summary

- Portfolio and book read queries filtered on `ctx.workspace_id()`, but all records are stored under the Live sentinel UUID (`aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa`).
- When a non-Live workspace is active the WHERE clause matched no rows, so the UI showed empty portfolio/book lists.
- Fixed all three read methods (`read_latest()`, `read_latest(id)`, `read_all(id)`) in both `portfolio_repository.cpp` and `book_repository.cpp` to always use `live_workspace_uuid_str` instead of `ctx.workspace_id()`.
- Write and delete operations are unchanged — they remain correctly workspace-scoped.

## Test plan

- [ ] Switch to a non-Live workspace; verify portfolios and books are visible.
- [ ] Verify portfolio/book create/update/delete still work and are scoped to the correct workspace.
- [ ] Verify Live workspace shows portfolios and books as before.

### DIFF
--- a/doc/agile/versions/v0/sprint_18/ore_samples_support/task_fix_workspace_portfolio_book_visibility.org
+++ b/doc/agile/versions/v0/sprint_18/ore_samples_support/task_fix_workspace_portfolio_book_visibility.org
@@ -26,14 +26,14 @@ propagation, UI data-loading path, or missing join); and apply the fix.
 
 * Status
 
-| Field        | Value                                     |
-|--------------+-------------------------------------------|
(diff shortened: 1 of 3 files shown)


mcraveiro (commented):
All six sites updated to `const std::string&` — fixed in 4ea9b2c.

PR #855, opened 2026-05-25
[fix] Dashboard start-services: wrap with setsid to survive shell exit

mcraveiro:
## Summary

- Same root cause as #851 (start-client fix): `compilation-start` runs `start-services.sh` in Emacs's process group. When the script's shell exits after the wait loops complete, SIGHUP kills `nats-server` and `ores.controller.service` before they can serve any requests.
- Fix: wrap the command with `setsid` so the script runs in its own session. Background processes launched with `&` inside the script are orphaned to init on shell exit and survive.
- `stop-services.sh` and `status-services.sh` are unaffected — they don't launch long-running processes.

## Test plan

- [ ] Click "Start services" from the dashboard; verify NATS and controller come up and stay up.
- [ ] Click "Service status" to confirm services are running.
- [ ] Click "Stop services" to confirm clean shutdown still works.

### DIFF
--- a/projects/ores.lisp/src/ores-dashboard.el
+++ b/projects/ores.lisp/src/ores-dashboard.el
@@ -642,7 +642,7 @@ Falls back to two spaces when icons are unavailable, preserving alignment."
             (let ((lbl label)
                   (s   (expand-file-name "build/scripts/start-services.sh" root))
                   (r   root) (db dash-buf))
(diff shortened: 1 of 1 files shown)


mcraveiro (commented):
Good catch. Fixed in 29291dc — using `executable-find` to check availability before prepending `setsid`, so it falls back gracefully on macOS/BSD.

PR #916, opened 2026-05-28
Sprint 18 housekeeping: task record and inbox memory

mcraveiro:
## Summary

- Records PR #914 in the drop-version-field task doc
- Adds memory `capture_goes_to_inbox.org`: new captures always go to `inbox/`, not `near/` or `far/`; wires it into `memory.org`


mcraveiro (commented):
Fixed in 2781f1376 — shortened the memory.org entry to 'use the inbox bucket for new captures and triage to near/far separately', keeping path details only in the canonical memory file.

PR #1159, opened 2026-06-07
[build] Move continuous builds to staggered schedules with relevance guard

mcraveiro:
## Summary

Remove push-to-main triggers from continuous-linux/macos/windows and replace with cron schedules (Linux 2h, macOS 3h, Windows 4h). A guard job diffs HEAD against the last successful run SHA and skips the build when only doc-class paths changed.

## Changes

- Replace push-to-main triggers with cron schedules (2h/3h/4h staggered)
- Add relevance guard job that skips when only doc-class paths changed since last green run
- Set cancel-in-progress false so staggered runs queue rather than cancel

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Retune CI for tens of merges an hour](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/retune_ci_for_merge_rate/story.html) | 3F7752F8-B719-40B4-BA89-09223410999E |
| Task | [Move main builds to staggered schedules with a relevance guard](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/retune_ci_for_merge_rate/task_decouple_main_builds.html) | C320F78B-D900-4CC4-B45D-B2F170DD46D1 |


mcraveiro:

mcraveiro:
Thanks for the thorough reviews. Addressed in 1be28c6.

**Fixed:**
- **Safe default output**: `should_build=true` is now the first line written to `GITHUB_OUTPUT`, so any subsequent failure (transient `gh api` error, network issue) leaves the build enabled rather than silently skipping it.
- **`timeout-minutes: 5`** added to the guard job to bound hung API calls.
- **`|| true` isolation**: each `grep -v` stage is now wrapped in `{ grep ... || true; }` subshells so `git diff` failures propagate instead of being swallowed by a trailing `|| true` on the full pipe.

**Not changed:**
- **`.org` filter scope**: the global `\.(md|org)$` filter is intentional — all org-mode files anywhere in the tree are doc-class (established in PR #1156 which added `*.org` to the PR classifier for the same reason).
- **Cron overlap at midnight**: minor runner-queue concern; happy to revisit if contention is observed.
- **Triplicated guard job**: known maintenance tradeoff; captured as a separate task for a future cleanup.

PR #1183, opened 2026-06-07
[ores.shell] Connect to a TLS broker from the shell

mcraveiro:
## Summary

Lets the shell connect to a secured broker on its own, so a fresh system can be provisioned from the shell before connecting through the Qt UI (whose connect triggers the provisioning wizards — the chicken-and-egg found while testing the script panel). The in-REPL connect command previously rebuilt a bare nats://host:port, dropping TLS and subject prefix. Now: connect reuses the connection the binary was started with (subject prefix + TLS, via a new repl connection template) and accepts --tls-ca/--tls-cert/--tls-key/--subject-prefix overrides; its help lists the flags and a failed plain connect hints them. On the Qt side the Shell action is enabled from startup and the embedded shell always opens a usable REPL: connected, it connects (logging in only if the app is); not connected, it stays a bare REPL and prints a ready-to-paste connect line with the env-resolved TLS paths and prefix. The menu/toolbar entry now reads 'Shell'. Built clean; shell unit tests green; GUI verification needs a display (manual).

## Changes

- connect: reuse startup connection (prefix+TLS) + --tls-*/--subject-prefix overrides; help and failure hint.
- repl: connection_template parameter; standalone app passes options.connection.
- Qt: Shell action always enabled; embedded shell always usable, ready-to-paste connect hint when not connected; 'Console' -> 'Shell'.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Provisioning script library](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/provisioning_script_library/story.html) | F7AC35C9-ADAE-4330-9640-AD2C7A044A55 |
| Task | [Connect to a TLS broker from the shell](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/provisioning_script_library/task_shell_connect_tls.html) | 703E57A2-D7D8-4800-94AF-6EEDF015F8C0 |


mcraveiro:
Thanks for the reviews — addressed in 30a0b78:

**Bug — connect failure falls through to login attempt** (ShellMdiWindow::start_shell): Fixed. The login block is now guarded behind a `shell_connected` flag, so a failed shell connect no longer issues a login `request(...)` on a disconnected session (which produced the confusing second error).

**Bug/observation — TLS hint emits empty path arguments**: Already resolved. The not-connected hint was simplified in an earlier commit and no longer echoes TLS paths at all — it now reads "TLS and namespace are inherited from the application", so there are no empty `--tls-cert`/`--tls-key` placeholders to paste.

**Minor — stale `process_connect` docstring**: Updated — the `@param` block now describes `connection_template` + `args` (and the host/port-or-URL behaviour) instead of the old host/port/identifier.

**Nit — copyright year**: `connection_commands.hpp` bumped to 2025-2026.

**Minor — `nats_options::url` default in REPL template**: Added a comment at the REPL construction noting the template intentionally carries only TLS + subject prefix; `connect` always builds a fresh URL (from host/port or `$ORES_NATS_URL`) and overrides it.

**`setEnabled(true)` in two places / `has_tls` on merged opts / hint ordering**: Acknowledged, no change — correct as flagged. The hint-ordering is cosmetic and left for visual check on a display.

PR #1202, opened 2026-06-08
[ores.shell,ores.qt] Generate the shell script library from recipes

mcraveiro:
## Summary

Make the shell recipes the single source of truth for the ores-shell script library: each doc/recipes/shell/*.org ores-shell block tangles into a committed, category-foldered, self-documenting .ores under projects/ores.shell/scripts/library/, run-able via the Qt panel (Open/Execute) or compass shell -f. Adds provisioning recipes (Barclays/GLEIF, Acme/synthetic, bootstrap, tenant/party), packaging, a drift check, and moves the db dashboard off deprecated scripts onto compass.

## Changes

- New ores-build-recipe-scripts.el + tangle_shell_scripts target: tangle each recipe's ores-shell block into library/<category>/<name>.ores (TARGET-FILE redirect, LANG-RE filter), strip trailing exit, prepend title/description header
- Provisioning recipes via compass add recipe: Barclays (GLEIF), Acme (synthetic), bootstrap smoke, tenant-only, party-only; wired into the shell recipe inventory
- Retire hand-written bootstrap.org/provision_all.org; package library to share/ores/shell-scripts; shell-script-drift CI check
- Qt: standalone script-editor MDI window, folder-tree library panel with Open/Execute context menu, shell+editor geometry persistence
- Dashboard db lifecycle onto compass: recreate_env.sh -> compass db recreate; teardown_database.sh -> compass db drop -k

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Provisioning script library](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/provisioning_script_library/story.html) | F7AC35C9-ADAE-4330-9640-AD2C7A044A55 |
| Task | [Generate the script library from shell recipes](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/provisioning_script_library/task_generate_script_library_from_recipes.html) | BC6E0C19-4E05-419F-B5A8-625A216D3CD8 |


mcraveiro:
Thanks for the thorough reviews — addressed across 98148257 (elisp) and 5dca75d8 (Qt):

**Accepted & fixed**
- **Empty `library/index/` dir** — the generator now skips prose-only recipes (no `ores-shell` block), so the inventory index no longer creates an empty category folder.
- **Multi-block recipe behaviour** — documented in the module commentary (one recipe = one script; multiple blocks would concatenate).
- **`statusChanged` never emitted** — now emitted on Open/Execute so the shell status bar gives feedback.
- **`on_filter_changed` un-hides empty groups** — fixed; a top-level group is now hidden while filtering when it has no match, and only force-shown when unfiltered.
- **Execute has no confirmation** — added a `QMessageBox::question` naming the script before Execute runs it (Execute bypasses the editor and can mutate state).
- **`user_dir()` ignores `mkpath` failure** — now logs a `warn`.
- **Shared geometry key stacks editors** — new editors cascade past those already open (offset by 24px each) instead of restoring on top of one another.

**Clarified (resolution was already correct)**
- **`ores-db.el` teardown path traversal** — `sql-scripts-directory` returns `projects/ores.sql`, so the two `..` steps land exactly at the repo root (identical to `run-compass`). I DRY'd both onto a shared `ores-db/--compass` helper so the resolution lives in one place.

**Acknowledged, no change**
- **Hardcoded `Secure-Password-123`** — intentional local dev/demo default; these scripts are not wired into any CI pipeline against shared infra.
- **`setup-emacs@v8.0` tag vs SHA** — matches the existing `template-drift.yml` policy; left consistent.
- **`drop-at-point` predating the helper** — now also routes through `ores-db/--compass`.

All builds green locally; the library regenerates with zero drift.

mcraveiro:
Round 2 addressed in eaeb294c:

- **`drop-at-point` path traversal** — now uses `ores-db/--compass` like `teardown-at-point`. (My round-1 reply over-claimed this — only `run-compass` and `teardown` had been converted; `drop-at-point` is now done too.)
- **`filter_item` bitwise `|`** — added the inline comment explaining it's intentional (all children must be visited; `||` would short-circuit and leave stale-visible items).
- **`library_dir()` empty return** — now logs a `warn` ("set ORES_SHELL_SCRIPTS_DIR to override") before returning empty, so a dev running from an unexpected directory sees why the panel is empty.

Builds green; elisp byte-compiles clean.

PR #1217, opened 2026-06-09
[build] Use Continuous CDash group for site builds; record revision

mcraveiro:
## Summary

- Site builds were landing under the Experimental CDash group (same as canary) because CTest only recognises Nightly/Continuous/Experimental as model names — "Site" silently fell back to Experimental. Change to Continuous, which is the correct semantic for a build that fires on every merge to main.
- Override CTEST_UPDATE_COMMAND to emit the SHA from ORES_BUILD_COMMIT when available, so CDash shows the commit hash in the Revision column (GitHub's shallow detached-HEAD checkout causes ctest_update to report 0 new commits, leaving the column blank).

## Test plan

- [ ] Merge to main; build-site.yml fires cdash job → site build appears under Continuous in CDash with a revision hash
- [ ] Open a doc-class PR; site job fires → build appears under Continuous too, with its PR revision


mcraveiro:
Closing — re-raising via compass pr create with proper traceability.

PR #1222, opened 2026-06-10
[ores.qt,doc] Fix monospace font ownership; partition shell recipes

mcraveiro:
## Summary

The Qt shell/editor rendered proportional because a global QSS font-family overrode setFont(). Give font family a single owner — the Qt font system — and resolve monospace to an installed, genuinely fixed-pitch family (logged for grep). Also partition doc/recipes/shell into the library's category sub-folders.

## Changes

- Drop font-family/font-size from the global QWidget QSS rule; set the app default font via QApplication::setFont(FontUtils::applicationFont())
- Monospace surfaces use QWidget::setFont(FontUtils::monospace()), now authoritative; removed the per-widget setStyleSheet workaround and monospaceCssFragment
- FontUtils::monospace() resolves the first installed+fixed-pitch family from a fallback chain, skipping a symbols-only 'Fira Code'; resolution logged at ores.qt.font_utils for grepping
- Partition doc/recipes/shell into category sub-folders mirroring scripts/library/; tangler scans recipes recursively; links unaffected (org-roam ids)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Provisioners from the shell — analysis and design](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/run_provisioners_from_shell/story.html) | A19D5B00-BEF3-4718-922D-16F58EEBAAB4 |
| Task | [ores.shell output misaligned: not using a fixed-width font](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/run_provisioners_from_shell/task_ores_shell_fixed_width_font.html) | 9CDEB91C-A0CF-44FF-AB27-654771027CAA |


mcraveiro:
Thanks for the reviews — addressed:

**Accepted & fixed**
- **Dead `DefaultPixelSize` constant** — removed (it was only used by the deleted `monospaceCssFragment()`).
- **`resolvedMonospaceFamily()` could crash if called before `QApplication`** — added a `QCoreApplication::instance()` guard that returns empty (no resolution, no caching) so `monospace()` falls back to the family chain; no current path hits it, but it's now safe.
- **`ComputeConsoleWindow` consistency** — its error view now uses `FontUtils::monospace()` instead of `QFontDatabase::systemFont(FixedFont)`, so it shares the resolved, logged fallback chain (the single source of truth this PR establishes).

**Acknowledged, no change**
- **`applicationFont()` accepts proportional fallbacks** — by design; it *is* the proportional UI default, so any sans-serif fallback is correct (the fixed-pitch verification only matters for `monospace()`).
- **`MonospaceFontFamily` used only as an alias** — kept as a named constant: it documents the preferred face and reads better than a bare string literal at the head of the chain.

Builds green (ores.qt + compute lib).

PR #1224, opened 2026-06-10
[iam,tests] Fix flaky IAM repository tests

mcraveiro:
## Summary

- Change `account_generator.cpp` email suffix from `id_str.substr(0, 8)` (timestamp prefix, same across parallel workers in the same millisecond) to `id_str.substr(24)` (random tail, 48 bits of independent `mt19937_64` state)
- Closes the Qt help system from the user manual story bookkeeping (all tasks DONE, sprint row updated)

## Root cause

UUID v7 places the millisecond timestamp in bytes 0–5. The old `substr(0, 8)` took the first 8 hex chars — bytes 0–3, the upper 32 bits of the timestamp — which are identical for every UUID generated within the same millisecond. CI canary runs with `CTEST_PARALLEL_LEVEL=$(nproc)`, so parallel test processes collided on `accounts_email_uniq_idx`.

## Fix

`substr(24)` takes the last 12 hex chars (bytes 10–15), drawn entirely from the `mt19937_64` engine state seeded by `std::random_device`. Collisions require 48-bit random state agreement across independent processes — astronomically unlikely.

## Test plan

- [ ] CI canary (`linux-gcc-debug-ninja`) passes the full IAM test suite with maximum parallelism
- [ ] No new unique-constraint failures in IAM tests

### DIFF
--- a/doc/agile/versions/v0/sprint_20/fix_flaky_iam_tests/story.org
+++ b/doc/agile/versions/v0/sprint_20/fix_flaky_iam_tests/story.org
@@ -22,12 +22,12 @@ a retry loop or test ordering workaround.
 
 | Field         | Value                                  |
 |---------------+----------------------------------------|
-| State         | BACKLOG                              |
+| State         | DONE                                 |
 | Parent sprint | [[id:D366207F-9E8D-46A2-A7DC-6808DC7FBF31][Sprint 20]] |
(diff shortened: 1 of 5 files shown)


mcraveiro:
Removed the inline comment in 469df6778 — the commit message carries the reasoning, which is where it belongs per CLAUDE.md convention.

PR #1238, opened 2026-06-10
[codegen] Teach entity templates the field-group contract; zero-diff regen

mcraveiro:
## Summary

Phase C of the instrument-models pass (#1103): org_loader and entity templates consume :domain_identity_group:/:domain_audit_group:/:group: annotations so domain class and mapper regeneration is zero-diff for all 9 rates instruments; delete generate_trading_instruments.sh and lift do-not-regenerate warnings.

## Changes

- generator.py: detect group annotations, derive has_identity_group/has_audit_group/is_identity_group_column; auto-inject instrument_identity and audit_record includes
- org_loader.py: fix default_value falsy bug — numeric 0/0.0 silently skipped by pystache; keep as raw string
- cpp_domain_type_class.hpp.mustache: identity/audit conditional blocks replace flat header and audit tail
- cpp_domain_type_mapper.cpp.mustache: both map directions gain identity/audit conditional blocks with prefixed accesses
- 9 instrument org models: trim redundant includes, remove do-not-regenerate warnings, confirm idempotent regeneration
- Guard names corrected from ORES_TRADING_DOMAIN_ to ORES_TRADING_API_DOMAIN_ in 9 hand-written domain headers
- 108 new codegen-managed files added at ores.trading.api/ and ores.trading.core/ for all 9 instruments
- generate_trading_instruments.sh deleted

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Codegen unified model — org-mode migration](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/codegen_org_model_migration/story.html) | D85E2B30-8C64-46DF-819D-B50D2E70B73C |
| Task | [Teach entity templates the field-group contract; zero-diff regen](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_20/codegen_org_model_migration/task_teach_templates_group_contract.html) | 24666C1C-4129-4513-80A1-2F7B69114AA6 |


mcraveiro:
Fixed in 52a97e3b7.

**3-part FQN guard (generator.py):** Added explicit `ValueError` guards for both `domain_identity_group` and `domain_audit_group`. Any annotation not exactly 3 dot-separated parts now raises with a clear message pointing at the bad value.

**Natural-keys + has_identity_group (mapper template):** Added a `{{! ... }}` comment in both map directions (and the org source) documenting that natural keys map flat and combining them with identity-group columns in the identity sub-struct is unsupported. No current model hits this path.

**`load_org_table_model` consistency:** Checked — table model columns use `default` not `default_value`; the two fields don't overlap, no change needed.

**CI drift check:** Was failing because the org literate sources (`cpp_domain_type_class_hpp.org`, `cpp_domain_type_mapper_cpp.org`) hadn't been updated when I modified the mustache files directly in the prior session. Both org sources are now synced with their tangled output.

PR #1257, opened 2026-06-11
Knowledge graph: restore org-roam-ui to external/, fix tag export, add skill filetags

mcraveiro:
## Summary

- Moves `external/org-roam-ui/` (compiled org-roam-ui static build) back from `projects/ores.org-js/graph/` where it was incorrectly placed as if it were source code; updates `.gitignore` and `ores-build-site.el` accordingly
- Fixes `ores-org-roam-export.el` to populate each node's `tags` array from `ALLTAGS` in node properties (the org-roam `tags` table is empty in our DB — all tag data lives in `properties.ALLTAGS`); this enables org-roam-ui's built-in tag whitelist/blacklist filter panel to actually work
- Adds missing `#+filetags: :skill:` to all 65 `SKILL.org` files under `doc/llm/skills/` so skills appear in tag-filtered graph views

After this PR + a graph rebuild, filtering to `:memory: + :runbook:` (~55 nodes) or `:skill:` (65 nodes) produces a readable subgraph. The upstream fork (`mcraveiro/org-roam-ui`) will get a one-click "LLM view" preset button as a follow-up task.

## Test plan

- [ ] Site build deploys graph from `external/org-roam-ui/` without error
- [ ] Rebuild graphdata.json locally; check that node `tags` arrays are populated (e.g. a memory node should have `["memory","llm"]`)
- [ ] Open org-roam-ui graph, open Tweaks panel, add `memory` to the tag allowlist — only memory nodes should remain visible


mcraveiro:
Fixed in 4dcc275:

1. **all-tags-set gap** — added a `dolist` in the node-building lambda body (after `tags-alltags` is computed) to push each extracted tag into `all-tags-set`. The top-level `tags` array in `graphdata.json` is now populated from ALLTAGS, so org-roam-ui's Tweaks panel tag autocomplete works without manual typing.

2. **string-trim simplification** — replaced `(cl-remove-if #'string-empty-p (split-string (string-trim alltags-str ":") ":"))` with `(split-string alltags-str ":" t)`. The `t` omit-nulls argument handles both leading and trailing colons natively, eliminating the need for `string-trim` and `cl-remove-if`.

PR #1282, opened 2026-06-16
Add ORES_ENV_TYPE and propagate to postgres and ores.qt

mcraveiro:
## Summary

- Adds `ORES_ENV_TYPE` (development/staging/production) to the `.env` template via `env_init.py`; existing value is preserved on re-runs
- Propagates it to postgres via `ALTER DATABASE SET ores.env_type` in `compass db setup`
- ores.qt reads `ORES_ENV_TYPE` from `--env-type` CLI arg → env var → `"development"` fallback, and shows a colour-coded chip in the status bar (red=production, orange=staging, neutral=development)
- Also excludes `projects/*/modeling/` paths from CI build triggers (unrelated cleanup bundled in a separate commit)

## Test plan

- [ ] `compass env configure` generates `.env` containing `ORES_ENV_TYPE=development`
- [ ] Re-running `env configure` preserves an existing `ORES_ENV_TYPE` value
- [ ] `compass db setup` sets `ores.env_type` on the database (`SELECT current_setting('ores.env_type')`)
- [ ] ores.qt status bar shows env chip; `--env-type production` colours it red
- [ ] CI passes on Linux


mcraveiro:
All accepted. Fixed in commit 958d604:

**Security fix** — `compass_db.py`: `ORES_ENV_TYPE` is now validated against `{"development", "staging", "production"}` before SQL interpolation; unknown values print a warning to stderr and default to `"development"`.

**Chip visibility** — removed the redundant `setToolTip` and `setVisible(!name.isEmpty())` from `setInstanceInfo`. The intent (per the task notes) is for the chip to always be visible; `setEnvType` already owns both chip visibility and the full tooltip unconditionally, so the guards in `setInstanceInfo` were dead code creating the inconsistency.

**CI pattern** — added `projects/*/*/modeling/*` to the `pr.yml` modeling exclusion glob to match the depth-agnostic regex in `continuous-*.yml`.

**Broken site link** — committed the previously untracked `task_spike_compass_story_done.org`; its ID was already referenced in `story.org`, which caused the site build to fail with 'Unable to resolve link'.

PR #1283, opened 2026-06-16
Implement compass env provision/deprovision with ores_dev_ naming

mcraveiro:
## Summary

- New `compass env provision [name] [--type light|full] [--preset P]`: creates a git worktree at `../ores_dev_<name_underscored>`, assigns isolated ports by scanning sibling `.env` files, writes a skeleton `.env`, and optionally runs `env configure`
- New `compass env deprovision <name> [-y]`: removes the worktree and directory; falls back to legacy `OreStudio.*` naming for backwards compatibility
- Updated `env_init.py`: reads `ORES_ENV_NAME`/`ORES_BASE_PORT`/`ORES_NATS_PORT` from existing `.env` early so re-runs are idempotent; fixes the hyphen-in-DB-name bug for adjective-noun envs (`festive-hawking` → `ores_dev_festive_hawking`); writes `ORES_ENV_NAME` and `ORES_BASE_PORT` to generated `.env`
- Updates story D2 decision: directory naming is now `ores_dev_{name_underscored}` (same as DB name) rather than `OreStudio.{name}`
- Adds provision and deprovision recipes and skills

## Test plan

- [ ] `compass env provision` auto-generates an adjective-noun name and creates `../ores_dev_<name>` with correct skeleton `.env`
- [ ] `compass env provision festive-hawking` creates `../ores_dev_festive_hawking`
- [ ] Ports are distinct from sibling environments
- [ ] `compass env configure` in new worktree uses `ORES_ENV_NAME` and preserved ports
- [ ] `compass env deprovision festive-hawking -y` removes directory and git worktree registration
- [ ] `compass env configure` on legacy `OreStudio.local1` is unaffected
- [ ] CI passes on Linux


mcraveiro:
All accepted. Fixed in two commits:

**08516a3** — Bug fixes:
- NATS print: `{label}` → `{env_name}` so the shown config filename matches what `nats_init.generate()` produces.
- `int()` conversion: refactored to set defaults first, then `try/except ValueError` with a warning, matching the pattern in `_scan_ports()`.
- Deprovision fallback: swapped order to `rmtree` first, then `git worktree prune`, so git sees the directory is gone and cleans up the stale registration.

**029e15e** — Minors:
- Stale comments updated: `compass env create` → `compass env provision`.
- `ORES_ENV_TYPE`: read from existing `.env` via `existing.get("ORES_ENV_TYPE", "full")` and written back to the full `.env` so it is not silently dropped after `env configure`.

The nit on the unconditional success message is noted but not changed — `shutil.rmtree` raises `OSError` on failure so it already propagates as an exception; the message is only reachable on actual success.

PR #1294, opened 2026-06-25
[build] Use sccache wrapper to enable cross-worktree cache hits

mcraveiro:
## Summary

- Each git worktree has a different absolute path prefix, so all \`-I\` flags in compiler invocations differ between worktrees. Without \`SCCACHE_BASEDIR\`, sccache hashes the raw absolute paths and produces 100% cache misses across worktrees (confirmed by simulation across 2335 TUs).
- The wrapper script sets \`SCCACHE_BASEDIR\` to the worktree root at build time, normalising all paths to relative before hashing, so all worktrees on the same commit share a single cache entry per TU.

## Changes

- Add \`build/scripts/sccache_wrapper.sh\`: resolves the project root from its own location (\`build/scripts/\` is two levels below root) and sets \`SCCACHE_BASEDIR\` before exec-ing sccache — no hardcoded paths, works across any worktree or machine
- \`CMakeLists.txt\`: point \`CMAKE_C/CXX_COMPILER_LAUNCHER\` at the wrapper instead of raw sccache; \`FORCE\` to override stale cached values on re-configure; add explanation comment

## Test plan

- [ ] Re-configure two worktrees on the same commit; build the same target in both; second build should show sccache hits
- [ ] Confirm \`sccache --show-stats\` shows cache hits after second worktree build

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


mcraveiro:

mcraveiro:
Fixed in a09fe7b7c:

1. **`set -e`** — added. A failed `cd` now aborts loudly rather than silently producing a wrong or empty `SCCACHE_BASEDIR`. (accepted — all 3 reviews)

2. **`${SCCACHE_BASEDIR:-...}` overridability** — wrapper now uses the env var if already set, falling back to the computed project root. (accepted — 1 review)

3. **`if(NOT EXISTS ...)` existence check** — added to CMakeLists.txt. Gives a clear `FATAL_ERROR` diagnostic if the wrapper is missing rather than a confusing per-TU shell error. (accepted — 1 review)

4. **`exec sccache` by PATH** — not changed. `cmake` `find_program` and the build environment share the same PATH, so they're always consistent in practice. A `configure_file` approach would reintroduce a generated script, defeating the goal of having a clean committed file. Added a comment in the script documenting this assumption. (declined — 3 reviews flagged)

5. **Task STARTED → DONE before merge** — not changed. Tasks close after merge per project workflow.

6. **`CCACHE_PROGRAM` rename** — out of scope, deferred. (all reviews agree)

PR #1303, opened 2026-06-25
[compass,refdata] Fix task/story state icons; sync messaging codegen for currency auxiliaries

mcraveiro:
## Summary

- **Compass icon fix**: `list` and `search` showed ☑️ for every task regardless of state. Added `read_state()` and `icon_for_doc()` to `ui.py` so tasks now render 🔲 (BACKLOG), ▶️ (STARTED), 🔴 (BLOCKED), ☑️ (DONE), or ❌ (ABANDONED). Same fix applied to stories. Updated `doc_list.py` and the search path in `compass.py` to use the new function.
- **Settings doc**: documented the compass invocation convention in `claude_code_settings.org` — always call `./compass.sh` from the repo root; absolute paths don't match the allow patterns.
- **Messaging codegen sync** (Task `00B62E70`): verified `protocol`, `nats-eventing`, and `nats-handler` profiles for `rounding_type`, `monetary_nature`, `currency_market_tier` against the repo — zero drift across all nine files. Files were on disk but untracked; committed them. Backlog capture filed for the 13 spurious ERRORs emitted when running messaging profiles over incompatible `table`/`junction` models in `--component` mode.

## Test plan

- [ ] `./compass.sh list --type task --tag commission_currency` — verify BACKLOG tasks show 🔲, STARTED show ▶️, DONE show ☑️
- [ ] `./compass.sh list --type story --tag sprint_21` — verify DONE stories show ✅, open stories show 📖
- [ ] `projects/ores.codegen/codegen.sh regenerate --component refdata-cpp --profile protocol` — three auxiliary protocol headers regenerate with zero diff
- [ ] Same for `nats-eventing` and `nats-handler` profiles
- [ ] No DB changes; no build required to validate docs and Python changes

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


mcraveiro:
Fixed in c69ceec38 — removed `STATE_RE` and `read_state()` from `compass.py` and redirected both call-sites (`cmd_where` in-flight scan and heading locate) to `ui.read_state()`. The third variant `_STATE_RE` at line 1393 (`_read_task_detail`) has a looser tolerance (`[^|]*` vs `\s*`) to handle the task-file table format — private to that function, left in place.

On the story `BLOCKED` note: stories use `#+todo: BACKLOG STARTED | DONE ABANDONED` — no `BLOCKED` state — so the fallback to `📖` is correct and intentional.

The `get_*_response.success = true` default and header guard naming are pre-existing codegen conventions; noted but not changed.

mcraveiro:
Fixed in 978bafa87.

**`success = true` default for list responses**: Changed to `success = false` in `cpp_protocol_hpp.org` (all three list response structs) and added explicit `resp.success = true;` in the list handler success paths in `cpp_nats_handler_hpp.org`. Fail-closed matches the mutating-op pattern and is safer against future code additions.

**Header guards missing `CORE_`**: Changed `{{component_upper}}` → `{{component_core_upper}}` in `cpp_nats_handler_hpp.org`. Handler guards now correctly read `ORES_REFDATA_CORE_MESSAGING_*` matching the file's include path.

Both templates re-tangled and codegen re-run for the three currency auxiliaries. Zero drift after the fix.

PR #1318, opened 2026-06-25
[build] Fix base32 alphabet: explicit char list, drop string-literal ctor

mcraveiro:
## Summary

- `std::array<char,N>` has no constructor accepting a string literal; the paren-init form is accepted by GCC/libstdc++ but rejected by Apple libc++ (Xcode 15.4) with *no matching constructor*.
- The bug was dormant because sccache had a cached TU from an earlier build; the `SCCACHE_BASEDIR` change in PR #1294 invalidated all macOS cache keys, recompiling everything from scratch and exposing the error.
- Fix: replace `alphabet("...")` with an explicit aggregate `{'A','B',...}` at size 32, dropping the unused null terminator.

Fixes macOS CI failure first seen in run [28188008457](https://github.com/OreStudio/OreStudio/actions/runs/28188008457).

## Test plan

- [ ] macOS CI compiles `ores.utility` without error
- [ ] Linux CI unaffected (was already passing)

### DIFF
--- a/doc/agile/versions/v0/sprint_21/base32-array-init/story.org
+++ b/doc/agile/versions/v0/sprint_21/base32-array-init/story.org
@@ -1,0 +1,46 @@
+:PROPERTIES:
+:ID: 31E34129-1687-4672-A7F1-E63ADFE7E04C
+:END:
+#+title: Story: Hotfix: base32 alphabet uses string-literal paren-init rejected by Apple libc++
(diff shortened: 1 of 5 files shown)


mcraveiro:
Fixed in 93f47a4dd — adopted the constexpr string_view suggestion; swapped `<array>` for `<string_view>` in the include. Declined to remove the redundant `static` (pre-existing, out of scope for this hotfix).

PR #1337, opened 2026-06-26
[synthetic] Step 3: DB write integration for EUR/USD FX spot feed

mcraveiro:
## Summary

- **DB write integration**: at startup, `application.cpp` looks up or creates the EUR/USD `market_series` via NATS; on each tick `fx_spot_feed.cpp` persists a `market_observation` (series_id, tenant_id, value to 6 d.p.) via `save_market_observations_request`
- **compass `env upgrade`**: new command that flips a light checkout to a full env — patches `.env` in-place (ORES_PROVISION_TYPE → full, ORES_PRESET → the supplied preset), initialises the vcpkg submodule, then runs `cmake --preset` to trigger vcpkg install
- **compass bearings**: identity line ("Your name is: bright faraday") added at the top
- **compass fleet**: replaced the multi-line block-per-worktree display with a compact aligned table (WORKTREE / IDENTITY / TYPE / BRANCH / TASK / PR / SYNC columns)

## Test plan

- [ ] Build passes: `ores.synthetic.service` links cleanly (verified locally — build [1060/1060])
- [ ] Run the service against a live stack; check NATS logs show `save_market_series_request` and `save_market_observations_request` on startup
- [ ] Verify EUR/USD observations appear in the `MarketSeriesMdiWindow` observation view
- [ ] `compass fleet` renders a clean table in both worktree and single-checkout setups
- [ ] `compass bearings` prints identity line at the top
- [ ] `compass env upgrade --preset linux-clang-debug-ninja -y` on a light env completes and produces a full build tree


mcraveiro:
Fixed in 90f4785 — addressing each point from round 1:

**Issue 1 (WARN on decode fail)**: Added `BOOST_LOG_SEV(lg(), warn)` in `application.cpp` immediately after the `rfl::json::read` call when `get_resp` is falsy, before the fallthrough to series-create.

**Issue 2 (request_sync blocking)**: Cannot add a timeout without an API change to `request_sync`. Documented the hard-block risk as a PoC-limitation in the `fx_spot_feed.hpp` class docstring. Tracked as a step 4 item.

**Issue 3 (non-atomic .env write)**: Fixed in `env_upgrade.py` — now writes to `.env.new`, `chmod 0o600`, then `Path.replace()` (atomic on POSIX), then writes the backup `.env.old` afterwards so `.env` is never absent on a crash mid-write.

**Minor (thread-confinement note on uuid_gen_)**: Added `// thread-confined — only used inside start()` to the member declaration.

**Minor (recorded_at == observation_datetime)**: Added inline comment noting the intentional equivalence for a synthetic feed with no record lag.

**Minor (unused f-string in bearings)**: Changed `print(f"...")` to plain `print("...")`.

**Minor (using namespace placement)**: Not changed. `using namespace ores::logging` must remain at namespace scope — moving it inside the anonymous namespace would remove severity identifiers (`warn`, `info`) from scope in `start()`, where `BOOST_LOG_SEV` calls them. Current placement matches `application.cpp` style.

**Observation (fleet drops story/task state)**: Intentional trade-off for the compact table. The old block format mixed three levels of context into one view; `compass where` and `compass journal where` still surface story/task/state in full. A `--detail` flag can be added if the loss proves painful day-to-day.

**Observation (fleet JSON additive change)**: Noted — `env_name` is an additive field; existing consumers that ignore unknown keys are unaffected.

PR #1340, opened 2026-06-26
[synthetic] Step 4: start/stop NATS handlers for FX spot feed

mcraveiro:
## Summary

- Add `market_feed_config_protocol.hpp` to `ores.marketdata.api` with start/stop request/response types on subjects `marketdata.v1.market_feed_configs.start` / `.stop`
- Add `feed_controller`: shared mutable state (mutex + atomic `running_`) holding the active `fx_spot_feed` and its tick thread; `start()` constructs GMM process + feed and launches the thread; `stop_signal()` sets the stop flag non-blocking; `shutdown()` stop-signals then joins at orderly exit
- Add `market_feed_config_handler`: header-only NATS handler that delegates to `feed_controller`; handles empty/malformed bodies by falling back to default-constructed request; idempotent (duplicate start or stop returns an error reply)
- Update `registrar::register_handlers` to accept `shared_ptr<feed_controller>` and wire both subjects on queue `ores.synthetic.service`
- Remove auto-start from `application.cpp`; feed is now started/stopped on demand via NATS

## Test plan

- [ ] Build passes (`cmake --build` 42/42 green)
- [ ] Send `marketdata.v1.market_feed_configs.start` via `nats pub` → service logs "Feed started: EUR/USD", ticks appear in DB
- [ ] Send `marketdata.v1.market_feed_configs.stop` → service logs "feed stop signalled", tick publishing stops
- [ ] Send start while already running → response `success=false, message="Feed already running: ..."`
- [ ] Send stop while not running → response `success=false, message="No feed is running"`
- [ ] Orderly shutdown (ctrl+C) after feed is running → `feed_->stop()` called, thread joined cleanly


mcraveiro:
Round 1 review addressed in c020591c1 (bulk fixes) and 1e42a9798 (shutdown guard).

**Fixed (9 items):**
- ✅ Destructor added to `feed_controller` — prevents `std::terminate` on exception path
- ✅ `is_running()` doc comment clarified — explicitly states flag tracks stop-not-signalled, not thread-exited
- ✅ `verifier_` and constructor parameter removed from `market_feed_config_handler`; unused includes dropped
- ✅ Dead `decode<stop_market_feed_config_request>` removed from `stop()` — req was never read
- ✅ `ctx` parameter removed from `registrar::register_handlers` signature — no handler uses DB context
- ✅ `mutable` removed from registrar lambdas
- ✅ Step-reference comment removed from `application.cpp`
- ✅ Typo "EUR/USD EUR/USD" fixed in protocol header doc comment
- ✅ `shutdown()` now guards `feed_->stop()` with `running_` check, matching `stop_signal()`

**Declined (2 items):**
- `thread_.join()` inside `mu_` on restart — PoC-acceptable; the `request_sync` blocking risk is already noted in `fx_spot_feed.hpp`; will be addressed when that timeout is fixed in a later step
- `[[maybe_unused]] cid` — consistent with the `market_series_handler` pattern used elsewhere in the service; weaving cid into all log lines is a broader change deferred to a future cross-cutting step

PR #1345, opened 2026-06-26
[ores.codegen, ores.qt] Sync Qt codegen for country; fix plural slot names; update recipes

mcraveiro:
## Summary

Regenerate all 12 Qt files for country via the new compass codegen entity command. Fix a bug in entity_pascal_short_plural derivation that generated wrong slot names (Countrys, Statuss) for entities with irregular or compound plurals. Update all stale codegen recipes that still referenced run_generator.sh or generator.py.

## Changes

- Fix entity_pascal_short_plural in core.py: derive from entity_plural last word (Countries, Statuses) instead of naive +s on pascal short form
- Regenerate all 12 Qt files for country (ClientCountryModel, CountryMdiWindow, CountryDetailDialog, CountryHistoryDialog, CountryController, .ui files) — disk files predated the template; template output accepted as canonical
- Replace all run_generator.sh / generator.py references in codegen, sql, and qt recipes with compass codegen commands
- Cross-reference entity-level and component-level entry points in the Qt UI regenerate recipe
- Add country unified-model migration task (BACKLOG) to the country story

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Commission: country](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/commission_country/story.html) | F0FC905B-6A70-486F-A23F-0064B6A3EE75 |
| Task | [Sync Qt codegen for country](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/commission_country/task_sync_qt_codegen_country.html) | 7663EB07-B3EF-43BA-8A44-5A29760C958A |


mcraveiro:

mcraveiro:
Round 1 addressed in b49e49634:

1. `#+updated:` stale date — Fixed (bumped to 2026-06-27).
2. Task Goal placeholder — Fixed (Goal and Acceptance sections filled in).
3. Silent fallback for missing `entity_plural` — Acknowledged. The entity lifecycle standard requires `entity_plural` for all entities; convention is the gate here rather than a code warning for now.
4. `Alpha-2 Code` label — Fixed via explicit `*** Detail fields` table in the entity model rather than editing the generated file directly. Zero-diff confirmed after regeneration.

mcraveiro:
Round 1 addressed in b49e49634:

1. `#+updated:` stale date — Fixed.
2. Task Goal placeholder — Fixed.
3. Alpha-2 Code label — Fixed via explicit `*** Detail fields` in the entity model; zero-diff confirmed.
4. Event subscription removed — Intentional regression. `EntityController` guards all subscription paths with `if (!eventName_.empty())` so passing `std::string_view{}` is a safe no-op, not a crash. Tracked in BACKLOG.
5. `maxLength` on `codeEdit` — Acknowledged. Server-side validates alpha-2 length. Restoration is in BACKLOG for when the full detail dialog (all ISO codes + official name) is hand-crafted.
6. Column reorder `UiPersistence` — Correct that `QHeaderView::saveState()` is index-based. Not a risk here: country Qt is commissioned fresh and the old hand-written implementation had a different column count, so any stale persisted state would fail to restore and fall back to defaults.
7. `all` vs `all-cpp` profiles — Intentional: `all = sql + all-cpp`; they are distinct profiles. The recipe now documents this distinction.

mcraveiro:
Round 1 addressed in b49e49634:

1. `#+updated:` stale date — Fixed.
2. Task Goal placeholder — Fixed.
3. Alpha-2 Code label — Fixed via explicit `*** Detail fields` table in the entity model (not by editing the generated file). Zero-diff after regeneration.
4. Dead branch in plural fix — Acknowledged. `str.split('_')` is never falsy so the else branch is unreachable; harmless. Left as-is for symmetry with the entity_singular guard above it.
5. Silent fallback — Acknowledged; convention (entity lifecycle standard) requires `entity_plural`. Low-priority warning addition.
6. Column reorder / `UiPersistence` — Index-based, confirmed. Not a risk here since country Qt is commissioned fresh; old code had different column count and any stale state would be discarded on restore.
7. Event subscription — Intentional regression; `EntityController` guards with `if (!eventName_.empty())`. Tracked in BACKLOG.
8. `maxLength` removed — Acknowledged; server validates; BACKLOG for full hand-crafted dialog.
9. API narrowing / include-path migration — Intentional; CI build is the gate for any missed callers.

PR #1342, opened 2026-06-26
[ores.refdata] Sync messaging codegen for all refdata entities

mcraveiro:
## Summary

Regenerate protocol, nats-eventing, and nats-handler headers for the full refdata entity set. Also ships the codegen model unification investigation (cross-cutting analysis, rides this PR for convenience).

## Changes

- Sync protocol, nats-eventing, and nats-handler codegen output for all refdata entities
- Normalise NATS subjects to underscores; rename list/delete/history fields to template standard
- Add missing *_changed_event.hpp files for entities that lacked them
- Fix book_status missing subcomponent flag (blocked messaging profiles)
- Add codegen model unification investigation doc (doc/analysis/)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Commission: party_status](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/commission_party_status/story.html) | 8B3D478B-DD68-44B7-A8BF-DB50F76C9506 |
| Task | [Sync C++ messaging codegen for party_status](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/commission_party_status/task_sync_cpp_messaging_codegen_party_status.html) | 7F64F55A-32BA-4774-8508-FF7F05FD984F |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1346, opened 2026-06-26
[synthetic] Step 5: ores.marketdata.client library + ClientManager::nats_client()

mcraveiro:
## Summary

- New `ores.marketdata.client` CMake sub-component with `fx_spot_subscription`: RAII wrapper around `nats::client::subscribe()` that converts an ORE key (`FX/RATE/EUR/USD`) to a NATS fan-out subject (`marketdata.v1.tick.fx.rate.eur.usd`), deserialises each message with `rfl::json`, and delivers `fx_spot_tick` to a caller-supplied handler; malformed messages logged and dropped
- Two smoke tests verify the ORE key → NATS subject formula without a live NATS connection
- `ClientManager::nats_client()` accessor added so `fx_spot_subscription` (and the step 6 chart window) can attach to the Qt session's raw NATS client without going through the entity-change notification path
- Fix three stale field references from PR #1342 codegen sync: `party_id_schemes` → `schemes`, `party_types` → `types`, `party_statuses` → `statuses`

## Test plan

- [ ] Build passes
- [ ] `ores.marketdata.client.tests` — 2 assertions in 2 test cases pass
- [ ] `ores.marketdata.api.tests`, `ores.marketdata.core.tests` still green (no regressions)
- [ ] `ClientManager::nats_client()` compiles and links in `ores.qt.api.lib`


mcraveiro:
Round 1 review addressed in 701d3b72b.

**Fixed (4 items):**
- ✅ Extracted `ore_key_to_subject` to `detail/subject_helpers.hpp`; both `fx_spot_subscription.cpp` and the tests now include the same header — tests exercise the production function directly
- ✅ Expanded tests from 2 to 4 cases (uppercase, already-lowercase, single-component, empty string)
- ✅ `ClientManager::nats_client()` now checks `is_logged_in()` before returning; `@throws` in the header updated to match
- ✅ Added `@note` to `fx_spot_subscription` class docstring: handler runs on NATS delivery thread, callers must marshal to Qt main thread

**Declined (1 item):**
- `inline static` in anonymous namespace — not changed; Review 3 confirmed this is the established convention across the codebase (client.cpp, registrar.cpp, etc.), so consistency takes priority

**Reverted (3 items from the original commit):**
- `EntityDetailDialog.cpp`, `LookupFetcher.cpp` field renames reverted; the codegen sync (PR #1342) that changed those field names is not present on current main after rebase — the original field names (`party_id_schemes`, `party_types`, `party_statuses`) are correct again

PR #1356, opened 2026-06-27
[codegen] org loader: comment blocks + zero-default; backlog captures

mcraveiro:
## Summary

Re-syncs the C++ core codegen output for currency and its currency_market_tier auxiliary (domain, repository, service, generator), fixes the template gaps the sync exposed, and restores three repository/table methods to standard template output. ores.refdata.api/core, ores.cli and refdata api tests build.

## Changes

- Fix domain-type generator template: guard the primary-key generator_expr so entities without a generator block (currency) no longer emit malformed code
- Drop redundant custom repository methods from the currency model now emitted as standard (fixes duplicate decl/def)
- Promote read_all(context), read_at_timepoint(context, as_of) and single-object convert_to_table back to standard template output, in the org sources so they survive re-tangling
- Port the primary-key generator guard into the generator org source (was mustache-only and reverted by tangle)
- Remove stray misplaced codegen output dirs (projects/ores.refdata.api etc, never compiled)
- Stopgaps: align currency_market_tier nats handler to the new service API, and disable the currency version-history HTTP route (501); both documented as gaps
- File a capture to define the canonical standard repository method set and regenerate all entities

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Commission: currency](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/commission_currency/story.html) | 7F9C0F29-4268-4B62-A1BB-2153ECF0A858 |
| Task | [Sync C++ core codegen for currency](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/commission_currency/task_sync_cpp_core_codegen_currency.html) | 8A84D304-431A-45CF-9E3C-6AC649F6A62F |


mcraveiro:

mcraveiro:
Addressed the review (commit `cdbd0048d`):

**1. Empty `currency_history_protocol.hpp`** — ✅ Deleted (`git rm`) rather than emptied. Its types live in `currency_protocol.hpp` now; removed the dead `#include`s from `protocol.hpp`, `risk_routes.cpp`, `registrar.cpp` and the shell currencies command.

**2. NATS subject hyphen → underscore** — ✅ Intended. Underscores are the normal subject convention; the new unified protocol template renders the snake_case `entity_plural`. The remaining hyphen entities (`party-statuses`, `business-units`, …) are legacy/pre-migration and will converge as they migrate. All in-repo consumers (handlers, Qt, shell) are regenerated to the new subjects.

**3. `display_order` lost `default 0`** — ✅ Fixed, and at the root cause: the loader parsed `:default: 0` to integer `0`, which Mustache treats as falsy so `{{#default}}` skipped it — the same zero-falsiness already guarded for `default_value`. The loader now keeps `default` as a raw string too; SQL regenerated with `default 0`.

**4. `list_currencies` debug log** — left as-is: that line is emitted by the shared service template (affects every entity, not currency-specific), so changing it belongs in a template tweak rather than this PR.

The flat-history regression you noted is tracked by capture `25EB0F7E` and the HistoryDialogBase Phase-C task.

mcraveiro:
**Scope change.** This PR was originally "Sync C++ core codegen for currency", but `main` independently landed its own unified-SQL migration for currency/currency_market_tier (`58cd323ce`) with different modelling choices (e.g. `coding_scheme: nullable`, `image_id: false`, `change_reason` kept in the validations table). Rather than carry a conflicting parallel migration, I've **conceded the migration to main** and reset this branch to salvage only the genuinely-unique, non-conflicting improvements:

- **org loader — comment blocks:** `parse_org` now drops `#+begin_comment`/`#+end_comment` contents, so model notes don't leak into generated doc comments.
- **org loader — zero column defaults:** the unified domain-entity column parser keeps `:default:` as a raw string (matching the table/junction parsers already on main), so `:default: 0` isn't silently dropped by Mustache's 0-is-falsy guard.
- **Two backlog captures** for observations found along the way: rename `_protocol.hpp` for the NATS world; define/regenerate the standard repository method set.

Everything migration-/consumer-related (model edits, handler/Qt/SQL regen, the validations bridge, the change_reason single-sourcing) is dropped as superseded by main. Findings about the bugs main's path should still verify (tenant isolation, etc.) were already captured.

mcraveiro:
Addressed the one outstanding nit from the follow-up review (commit `03d659484`): `standardise_codegen_repository_method_set.org` no longer references the unreachable commit `f0a20a4c1` (dropped in the scope-change reset) — replaced with a description of what surfaced the inconsistency, and fixed the stale pre-refactor template paths. Thanks — the underlying observation the capture tracks still stands.

PR #1361, opened 2026-06-28
[codegen] B5: physical-space address graph (HELD — do not merge until B8-B11)

mcraveiro:
## Summary

Implements the MASD physical-space binding model as an org-roam doc graph that codegen traverses: ores (root) -> technical space (ores.cpp/ores.sql/ores.doc/ores.plantuml) -> facet (ores.cpp.qt, ...) -> archetype (ores.cpp.qt.controller_header, ...). Node name = MASD address = --address value = ores.*.enabled drawer key, everywhere. One shared resolver (resolve_targets) used by both the codegen CLI and the compass wrapper (duplication removed); --address added with --profile kept as a deprecated compat shim. DRAFT/HELD: per the merge gate this lands only once B8-B11 are done and the commissions are off --profile; merging alone would leave main half-migrated.

## Changes

- Doc graph: root + 5 technical spaces + 21 facets + 52 archetype docs, all scaffolded via compass (new technical_space + archetype doc types)
- physical_space.py: load_graph + supported/target/generation sets (model-type filter, ores.*.enabled disables, most-specific-wins specificity); 15 pytest tests
- Single resolver resolve_targets() replaces facet_catalogue monolith reads; compass wrapper de-duplicated (-47 net lines); byte-equivalent generation verified across every legacy profile
- Graph-wide inventory generator (regenerate_physical_space_inventories.py) with --check gate regenerates all child tables from frontmatter
- Follow-ups filed: B8 (component archetypes), B9 (.mustache rename), B10 (retire monolith + content port), B11 (caller migration)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Resolve codegen model unification blockers](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/codegen-unification-blockers/story.html) | FA93094D-718E-42B0-A41A-434051CC6275 |
| Task | [B5: Implement supported/target set model for physical space binding](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/codegen-unification-blockers/task_b5-sql-only-guard.html) | 86510CDC-7543-4B18-A6DC-8EA3E1AF8A8F |


mcraveiro:
Thanks — addressed in c8bf447c5 and b32c04df1.

**P1 — facets_under("ores") dead code:** Fixed (c8bf447c5). Confirmed real: on the live graph ores.org is #+type: technical_space, so "ores" was a ts_facets key with an empty list and --address ores returned 0 facets. Moved the root check to the top of facets_under; --address ores now expands to all 21 facets.

**P2 — missing live-graph root test:** Fixed (c8bf447c5). Added test_target_root_resolves_all_facets_on_live_graph asserting compute_target_set('ores', live_graph) == all facets. The synthetic fixture omitted 'ores' from ts_facets, which is exactly why it couldn't catch this. 16 tests now.

**Minor — silent skips in resolve_targets:** Fixed (b32c04df1). Both skip paths (empty template/output, and output pattern that fails to resolve) now log.debug with the archetype address.

**Minor — --profile silent at runtime:** Fixed (b32c04df1). It now emits a runtime deprecation warning naming the --address equivalent, to accelerate the B11 caller migration.

**Minor — regex inconsistency:** Fixed (b32c04df1). Unified physical_space.py to the stricter \w[\w-]* form used by the inventory script.

**Minor — facets without model_types implicitly universal:** Acknowledged, not changed now. ores.sql.populate/sql.service own no archetypes yet, so they produce nothing. B8 (which populates the component/service archetypes) will add explicit #+model_types when it does. Noted as future work on those facets.

PR remains DRAFT/HELD per the merge gate (lands with B8–B11).

mcraveiro:
Same findings as the other review round — all addressed; see the consolidated reply above (P1 facets_under fix in c8bf447c5, minors in b32c04df1). 16 tests pass; --address ores now expands to 21 facets.

mcraveiro:
## B8 update — component & service-registry archetypes folded into the physical-space graph

This push adds 9 commits (`687ead9c2..e111b0058`) advancing **B8: Fold component-scaffolding archetypes into the physical-space graph**.

### Design (layer b)
Both non-entity scopes now select via the model's own declaration, re-unifying on the supported-set algebra:
- **Component** scope → a `#+component_kind` discriminator (flat/api/core/service).
- **Data** scope → dataset-as-org-model + drawer (design recorded; implementation is a follow-up data-scope task).

The first cut (`--address` subsets for components) was reversed during migration: the component templates have **output collisions** — `src/CMakeLists.txt` (4 kind variants) and `tests/stub_tests.cpp` (2) — that `--address` can't express but the kind discriminator resolves cleanly (only one variant is ever in the supported set).

### Engine seam (`physical_space.py` / `generate.py`)
- `#+default: disabled` → opt-in facets/archetypes.
- `is_enabled()` resolves `ores.*.enabled` most-specific-wins **to archetype depth**.
- `kind_matches()` — component-kind variant selection (no-op outside component generation).
- `resolve_targets` now reads the model's `:PROPERTIES:` drawer (closes the old B5 TODO) and `#+component_kind`.
- 26 unit tests pass.

### Migrations (doc-graph)
- **Service-registry** (5 archetypes): new TS `ores.shell` + facet `ores.shell.service`; populated `ores.sql.service`. Verified generation-neutral.
- **Component scaffolding** (25 archetypes): new TS `ores.cmake` + facets `ores.cmake.component`, `ores.cmake.service`, `ores.cpp.component`, `ores.cpp.service-app`. No `.mustache` touched.

### Verification
- Resolver reproduces every `facet_catalogue` profile exactly: **flat 10 / api 10 / core 10 / service 19**.
- `refdata/api`, `/core`, `/service` generate **byte-identical to both the legacy `--profile` path and the committed files**.

### Bonus bug fix
All 99 `component_overview.org` models were unloadable (bare `#+type: component`; 60 missing `#+name`/`#+full_name`) — component generation was **broken for every component** via both paths. Normalised all 99; they now load.

### Deferred (tracked in task notes + captures)
- `#+component_kind` tagging of ~40 non-standard subcomponents (`qt/*`, `http/server`, `iam/client`, …) — default to `flat` until tagged.
- Conditional folds: `cpp_protocol.cpp` (retire; no profile references it) and `sql_schema_artefact_create` (via default-off/`enabled`).
- `facet_catalogue` `* Component archetypes` retirement once compass/skills target the graph.
- Data-scope generation — its own task.
- Captures: `stale-generated-iam-sql`, `normalise-component-overview-models`.

PR #1362, opened 2026-06-28
[synthetic] Add market data generation config domain entities

mcraveiro:
## Summary

Adds the three domain entities for the synthetic generation configuration story (1D7FFB95): a market_data_generation_config container, its fx_spot_generation_config child, and the gmm_component grandchild. Each is hand-written across all layers to byte-for-pattern match the currency reference entity, so a future codegen pass regenerates them with minimal diff. Parent/child links use the codebase's established soft-FK pattern (UUID column validated in the insert trigger against the active parent row), as a real DB FK cannot target a bitemporal table's partial-unique key.

## Changes

- market_data_generation_config: full stack (domain/json_io/protocol, entity/mapper/repository/service/handler, temporal SQL, IAM perms)
- fx_spot_generation_config: full stack; soft-FK config_id -> container
- gmm_component: full stack; soft-FK fx_spot_config_id -> fx_spot config; component_index unique per config
- ores.synthetic.api converted from INTERFACE to a compiled library to match the refdata.api pattern
- Synthetic create/drop SQL aggregators wired into create.sql/drop.sql; 9 new synthetic IAM permission codes
- Verified: api/core/service build and link clean; db recreate provisions all three tables and permissions without error

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Synthetic generation configuration and config-driven feeds](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/synthetic_generation_configuration/story.html) | 1D7FFB95-E766-4A49-BEE6-23067B6E035B |
| Task | [Synthetic config entities across all layers (currency pattern)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/synthetic_generation_configuration/task_synthetic_config_entities.html) | FD9977D4-B5CB-4F72-AA8E-139F34A346D1 |

### DIFF
--- a/doc/agile/versions/v0/sprint_21/synthetic_generation_configuration/story.org
+++ b/doc/agile/versions/v0/sprint_21/synthetic_generation_configuration/story.org
@@ -8,7 +8,7 @@
 #+filetags: :market_data:synthetic:sprint_21:v0:
 #+created: 2026-06-28
 #+updated: 2026-06-28
-#+environment:
+#+environment: bright_faraday
 #+todo: BACKLOG STARTED | DONE ABANDONED
 
 This page documents a [[id:699A8D45-B67E-4D3E-9206-24FA17C51ADA][story]] in [[id:C34A07B2-6BE6-40E5-B1A5-640D5369CECD][Sprint 21]]. It captures the goal, current status, acceptance criteria, and the tasks that compose it.
@@ -29,7 +29,7 @@ data is correctly owned and traceable.
 
 | Field         | Value                                  |
 |---------------+----------------------------------------|
-| State         | BACKLOG                              |
+| State         | STARTED                              |
 | Parent sprint | [[id:C34A07B2-6BE6-40E5-B1A5-640D5369CECD][Sprint 21]] |
 | Now           | Not yet started.                       |
 | Waiting on    | Nothing.                               |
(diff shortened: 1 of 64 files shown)


mcraveiro:
Thanks — addressed the findings. Note CI also flagged a canary schema-validator failure (synthetic not a registered component prefix + missing RLS policies), now fixed in 8d0043038.

**Accepted & fixed:**
- **#1 ticks_per_hour / gmm_initial_price positive constraints** — fixed in a5547f668: SQL `check ("ticks_per_hour" > 0)` + `check ("gmm_initial_price" > 0)` and matching service guards.
- **#2 ore_key non-empty** — fixed in a5547f668: SQL `check ("ore_key" <> '')` + service guard.
- **#3 batch remove valid_to filter** — fixed in ceb571f04 across all three repositories.
- **#5 weight DB constraint** — fixed in a5547f668: `check ("weight" >= 0)`.

**Declined:**
- **#4 cascade soft-delete on parent removal** — acknowledged design limitation (soft-FK against a bitemporal partial-unique key). Out of scope for this entity PR; consumers/generators handle inactive parents, and a dedicated cascade story is the right place for it.

Verified: validate_schemas.sh 0 warnings; db recreate provisions all 3 tables + 6 RLS policies; core builds clean.

mcraveiro:
Thanks — addressed. (Also fixed a canary schema-validator failure: synthetic component prefix + RLS policies, in 8d0043038.)

**Accepted & fixed:**
- **Batch remove temporal filter** (all three repos) — fixed in ceb571f04.
- **ore_key non-empty** — fixed in a5547f668 (SQL CHECK + service guard); also added ticks_per_hour > 0 and gmm_initial_price > 0.
- **Truncated GPL header** in fx_spot/gmm handlers — fixed in be461c059.

**Declined:**
- **`list` handler swallows errors silently** — declined for now. This is inherited verbatim from the currency_handler template, and these entities are deliberately kept byte-for-pattern identical so a future codegen pass regenerates them. Changing the list response shape should be done globally across all handlers, not just here; raising it as a separate cross-cutting item rather than diverging this one entity.

**Informational notes** (read perm not enforced, weight-sum, optional description, party-scoped source uniqueness) — all intentional and match the established pattern / business intent; no change.

PR #1364, opened 2026-06-28
[codegen,ores.qt] Reconcile country to post-refactor codegen + base-class Qt flag/save/revert/pagination/geometry fixes

mcraveiro:
## Summary

Brings `country` fully into line with the refactored codegen and restores/repairs its Qt behaviour, doing the fixes at the **base-class + template-variability** level so every entity benefits and a future codegen re-sync can't wipe them. Driven throughout by literate org templates (prose per block; generated C++ stays comment-free).

## Codegen
- **clang-format as a built-in second step**: `generate.py` runs `clang-format -i` over generated C++ after rendering; `compass codegen entity --diff` formats before diffing. Template whitespace no longer reads as drift.
- **Restored per-method C++ doxygen** dropped during the cpp literate-org migration (service + repository header templates).
- Reconciled country's C++: removed custom repository methods the template now generates as standard (dedup); fixed `faker::internet::countryCode` → `faker::location::countryCode` API drift surfaced by the new synthetic generator.

## Qt base classes (defined once, can't be re-synced away)
- **`AbstractClientModel`**: `setImageCache`, `iconColumn()`, generic `flagDecoration(image_id)`.
- **`DetailDialogBase`**: flag editor (button + `FlagSelectorDialog` + inline key-field flag via `keyFlagField()`/`keyFlagIcon()`), `markDirty()`.
- **`EntityController`**: holds `ImageCache`.
- **`show_managed_window`**: restores MDI geometry *after* `addSubWindow` (fixes windows not remembering size/position on reopen — most visibly history).

## Qt template variability (gated, entities opt in)
- `has_flag_icon`: list flag column, detail "Icon" editor, inline alpha-2 flag, cache wiring controller→window→model and →dialog, persist changed flag on save.
- `has_change_reason_cache`: wires the change-reason cache (fixes Save doing nothing).
- `has_pagination`: request carries offset/limit and trusts the response total (fixes the list silently truncating at 100).
- Revert: copy the chosen version's values but reset `version` to 0 so the insert trigger appends (fixes Save rejected as a version conflict); `markDirty()` enables Save immediately.

## country model (`ores.refdata.country.org`)
Enabled `has_flag_icon`/`has_change_reason_cache`/`has_pagination`, added the missing detail fields (alpha3_code, numeric_code, official_name), and wired `RefdataPlugin` to pass the caches.

## Net user-visible fixes for country
Flags (list + detail + inline) restored, Save works, revert saves a new version, list pages past 100, detail dialog shows all fields, history/detail windows remember geometry.

## Verification
`ores.qt.api` and `ores.qt.refdata` build clean; template re-tangles with no `.mustache` diff; `template-drift` checks pass locally. Several follow-ups captured to inbox (GitHub-style history diff view; window titles using display name).


mcraveiro:

mcraveiro:

mcraveiro:
Thanks — thorough review. Dispositions:

**Bug — multi-tenancy in batch `remove`** — Fixed in `8268a5990`. You were right it's template-level: the generated `remove(ctx, vector<key>)` lacked the `tenant_id` filter. Added the `has_tenant_id`-gated predicate matching the single-key remove, so every regenerated entity is now isolated.

**Minor 1 — non-`static` max in `get_total_count`** — Fixed in `8268a5990` (template). Now `static const` like the rest.

**Minor 2 — missing debug logs (2-arg `read_at_timepoint`, unfiltered `read_all`)** — Fixed in `8268a5990` (country custom paste).

**Minor 3 — missing `@brief` on unfiltered `read_all`** — Fixed in `8268a5990` (added to the `country.org` paste block).

**Minor 4 — stale mustache `GENERATED FILE` org refs** — Fixed in `8f3575e35`: now point at `ores.cpp.repository.repository_header.org` / `ores.cpp.service.service_header.org`.

**Minor 5 — over-indented `_diff_entity` loop body** — Fixed in `d88d7baca`.

**Minor 6 — unary `+"-"` in the synthetic generator** — *Not changed.* It's a cosmetic generator-template artifact for an empty PK prefix; it compiles correctly and country's synthetic path is non-authoritative (the fictional generator is the test source). A proper fix (suppress the empty prefix in the generator template) is a generic generator-template cleanup better done on its own, not folded into this PR.

All template fixes re-tangled; `ores.refdata.core` builds clean.

PR #1367, opened 2026-06-29
[synthetic] Qt config UI for synthetic market data generation

mcraveiro:
## Summary

Adds the Market Simulator: a Qt configuration UI for authoring synthetic FX-spot market-data feeds and their GMM price models, plus the config-driven feed start/stop and a dry-run batch simulate endpoint that powers live preview charts. Includes a TradingView-style FX hero header and a fix for a service abort on zero-volatility components.

## Changes

- Market Simulator window: two-level feeds > FX-spot tree, child dialogs, MDI tabbed FX spot rate editor, dark-theme UX pass and wide layout
- Price behaviour: Simple/Advanced modes, geometric + arithmetic engines, GMM component table with volatility profiles, live return-distribution and sample-path preview charts
- Authenticated dry-run simulate endpoint (stateless batch sample paths) backing the live charts; config-driven Start/Stop feed
- FX overview hero header: big overlapping circular flags rendered crisply from SVG, bold pair title using refdata currency names
- Fix synthetic service abort (SIGABRT) on a zero-volatility GMM component: treat sigma<=0 as degenerate (draw the mean) in both process engines; add simulate-handler diagnostics
- Filed backlog captures: move generic quant primitives to ores.analytics, validate GMM params at the request boundary, and two compass services DX commands

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Synthetic generation configuration and config-driven feeds](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/synthetic_generation_configuration/story.html) | 1D7FFB95-E766-4A49-BEE6-23067B6E035B |
| Task | [Qt config UI for synthetic market data generation configs](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/synthetic_generation_configuration/task_qt_config_ui.html) | A433571D-5E17-4B3D-8C3C-72850BAE0554 |


mcraveiro:

mcraveiro:
Thanks for the thorough multi-reviewer pass. Addressed this round in two commits — `1a0d897` (service/protocol) and `78c0ce2` (Qt).

## Fixed
- **All-zero weights → `std::discrete_distribution` UB** (`onSaveClicked`): now rejected with a warning before sending. Same crash class as the σ=0 fix. `78c0ce2`
- **`ReturnDistributionChart` `QLineSeries` leak**: the line is now parented to its `QAreaSeries`, so `removeAllSeries()` destroys it too — no more leak per refresh. `78c0ce2`
- **No delete confirmation** (`onDeleteClicked`): added a `QMessageBox::question`; the feed variant warns that child FX rates are removed too. `78c0ce2`
- **Duplicated clamp limits**: hoisted `max_num_ticks`/`max_num_paths` into `simulate_fx_spot_paths_request`; service clamp and client spinners/clamp now share them. `1a0d897`
- **Silent fallback for unknown `process_type`**: added a `warn` log before defaulting to geometric. `1a0d897`
- **`make_gmm_process` dead code**: removed — confirmed zero callers (both feed controller and simulate handler use `make_process`). `1a0d897`
- **Default 50 sample paths**: lowered to 10 for a faster initial preview. `78c0ce2`

## Declined / corrected
- **"Non-deterministic tree ordering"**: not actually a bug — `feeds_`/`fxPairs_`/`components_` are `std::map`, not `unordered_map`, so iteration is deterministic (UUID order) and stable across reloads. Name-sorting would be a nice enhancement but isn't a correctness issue.
- **`version = 0` disables optimistic locking**: project-wide established pattern, not a regression — leaving as-is.
- **Dual signal connection / double currency fetch / eager `join_doubles` / `pi` vs `std::numbers` / `slug` Latin-1**: all benign (idempotent or cost-trivial / ASCII-only). Not changing this round.

## Deferred to backlog
- **Non-transactional multi-step save**: real limitation, but the temporal-history model means no corruption (old versions survive); the proper fix is a server-side atomic batch-save endpoint — too large for this round.
- **Regression test for the σ=0 fix**: agreed it's worth a test given it was a production crash.
- **Request-boundary GMM validation** (negative σ etc.): already captured in `validate_gmm_params_at_request_boundary.org`.

I'll file captures for the atomic batch-save endpoint and the zero-volatility regression test.

PR #1373, opened 2026-06-29
[synthetic] Config-driven multi-feed autonomous generation

mcraveiro:
## Summary

Lift the single-EUR/USD-feed PoC to autonomous, config-driven generation: on startup the synthetic service reads enabled configs from the DB and starts every enabled FX rate as a concurrent producer, resolving a market series per ORE key and publishing on its own synthetic producer channel. Also splits the spot-rate-blending summary into atomic knowledge notes. Verified end-to-end at the producer layer (DB + NATS); the live chart/grid awaits the marketdata authority and consumer stories.

## Changes

- feed_controller: map of concurrent feeds keyed by source name; per-feed start/stop or stop-all; shutdown joins all; resolves/creates a series per ORE key; publishes on synthetic.v1.tick.<source>
- application: auto-start enabled configs on startup (read from DB); drop hard-coded EUR/USD series pre-resolution
- fx_spot_feed: explicit publish subject; responsive stop (sleep in 100ms slices)
- protocol/handler/UI: carry source_name so manual start/stop works as an override
- knowledge: split spot rate blending into atomic cross-linked notes; enrich existing governance note; add blurb to Market Data Architecture; add codegen task

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Config-driven synthetic feed generation (per-config series, lifecycle)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/config_driven_generation/story.html) | E6679CB1-2246-44FD-B557-7F1791574B1B |
| Task | [Config-driven multi-feed generation (start a config)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_21/config_driven_generation/task_config_driven_multi_feed_generation.html) | A2CEB990-7DCA-487E-A6DA-867B725AD9F3 |

### DIFF
--- a/doc/agile/versions/v0/sprint_21/config_driven_generation/story.org
+++ b/doc/agile/versions/v0/sprint_21/config_driven_generation/story.org
@@ -55,7 +55,7 @@ This story is Scope 2 — the proper config-driven generation backend.
 
 | Task | State | Start | End | Description |
 |------+-------+-------+-----+-------------|
-| [[id:A2CEB990-7DCA-487E-A6DA-867B725AD9F3][Config-driven multi-feed generation (start a config)]] | STARTED | 2026-06-29 | | Make the synthetic service run many feeds at once and start/stop a whole config's FX rates together, publishing on producer-namespaced channels. |
+| [[id:A2CEB990-7DCA-487E-A6DA-867B725AD9F3][Config-driven multi-feed generation (start a config)]] | DONE | 2026-06-29 | 2026-06-29 | Make the synthetic service run many feeds at once and start/stop a whole config's FX rates together, publishing on producer-namespaced channels. |
 | [[id:645B48D0-E2DF-4ACA-AD28-526DC48AFDAD][Ornstein-Uhlenbeck (mean-reverting) process engine]] | BACKLOG | | | Add a mean-reverting Ornstein-Uhlenbeck price-process engine alongside geometric and arithmetic, selectable on the synthetic screen. |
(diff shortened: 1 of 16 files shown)


mcraveiro:

mcraveiro:
Thanks — three thorough reviews. Addressed in `a5ac182`.

## Fixed
- **Stop-flag race** (`fx_spot_feed::start()`): removed the `stop_flag_.store(false)` reset. The instance is always fresh, and the reset could clobber a `stop()` arriving after thread spawn → loop + `join()` hang forever. Good catch — this was a genuine bug.
- **Blocking RPC under the mutex** (`feed_controller::start()`): `resolve_series()` now runs *before* `mu_` is taken, so NATS round-trips no longer block concurrent `stop()`/`shutdown()`; it's find-or-create, so a concurrent duplicate start is harmless.
- **Unsanitised subject** (`producer_subject`): non-token chars (whitespace, `*`, `>`, etc.) are replaced with `_`; `.` is kept since source names are legitimately dotted and `.` is the intended hierarchy separator.
- **Malformed start/stop** (`market_feed_config_handler`): both now reply with an error and return, instead of silently starting an EUR/USD default / stopping all feeds.
- **Stale PoC comments**: refreshed in both the protocol header and the handler.
- **`running_count()` const** (`mu_` now `mutable`); **GMM lookup-key asymmetry** commented in `application.cpp`.

## Declined / deferred (with reasons)
- **NATS publish before persist**: kept publish-first deliberately — the live broadcast shouldn't wait on (or be lost to) a slow/down marketdata service; the historical-gap is a documented PoC trade-off.
- **N serial `list_series` on startup**: now outside the lock; collapsing to a single cached list is a worthwhile optimisation but PoC-scope — left as a follow-up.
- **Redundant explicit `shutdown()`**: harmless (idempotent) and gives an explicit ordering guarantee; left as-is.
- **`stop` wire break (`ore_key`→`source_name`)** and **per-feed fail-open on startup**: intended, as the reviews note.

PR #1376, opened 2026-06-30
Regenerate synthetic config types via codegen (backend)

mcraveiro:
## Summary

- Register `ores.synthetic` as a codegen component and fix the modern CLI blockers that prevented entity generation
- Fix Qt template gaps: UUID key wrapping (`key_to_string_prefix`/`suffix`), double field handling (`is_double` knob + `QString::number`/`.toDouble()` branches in detail and history dialog templates), and `setCreateMode` guard for non-UUID PK entities
- Add `checkDouble()` to `HistoryDialogBase` to support double field diffs in history dialogs
- Go all-codegen for `gmm_component` Qt layer; reconcile hand-written controller and window wiring
- Model and fully regenerate `fx_spot_generation_config` and `market_data_generation_config` (domain, json_io, table, table_io, generators, eventing, messaging protocol, handler, repository, service, Qt client model/controller/detail dialog/history dialog/mdi window, SQL create + notify triggers)
- Capture two backlog items: eventing pipeline gap in the synthetic service and a proposed codegen approach for per-entity eventing registrars

## Test plan

- [ ] Build `ores.qt.synthetic` cleanly (no autogen or GLOB cache issues after full reconfigure)
- [ ] GMM Component CRUD: create, edit, history diff shows double fields (mean/stdev/weight) correctly
- [ ] FX Spot Generation Config CRUD: double field (`gmm_initial_price`) serialises and displays correctly in detail and history dialogs
- [ ] Market Data Generation Config CRUD: basic create/edit/view cycle
- [ ] SQL: `ores_synthetic_fx_spot_generation_configs_tbl` and `ores_synthetic_market_data_generation_configs_tbl` tables and NOTIFY triggers present after migration
- [ ] `MarketSimulatorWindow` loads both config lists without field-name errors


mcraveiro:
Fixed in 1c97aaaf3.

**Accepted and fixed:**
- GMM components unique index: added `component_index` to `(tenant_id, party_id, fx_spot_config_id, component_index)` — the natural uniqueness anchor for a GMM component is the parent config + ordinal position.
- FX spot configs unique index: widened to `(tenant_id, party_id, config_id, base_currency_code, quote_currency_code)` to support multiple FX pairs per parent config.
- CHECK constraints restored on fx_spot_generation_configs: non-empty currency codes, `base <> quote`, non-empty source_name/ore_key, `gmm_initial_price > 0`, `ticks_per_hour > 0`, `process_type in ('geometric', 'arithmetic')`.
- `stdev >= 0` and `weight >= 0` checks restored on gmm_components.
- Double fields in domain structs: added `= 0.0` to `mean`, `stdev`, `weight` in `gmm_component` and `gmm_initial_price` in `fx_spot_generation_config`; restored `process_type = "geometric"`. (Note: `boost::uuids::uuid` default-constructs to nil in Boost so those fields were safe.)

**Declined (with capture):**
- Soft-FK trigger validations: the codegen SQL template does not generate trigger-side parent-existence checks. This is a template gap — it will be captured as a separate backlog item rather than hand-patching the generated file, which would be overwritten on the next codegen run.

**Minor note acknowledged:**
- `is_double` only matches bare `double`/`float`, not `std::optional<double>`. No current field exercises this path; will be addressed if optional doubles are introduced.

mcraveiro:
Soft-FK trigger validations now also fixed — in commit 8d8f8d294.

Rather than hand-patching the generated SQL (which would be overwritten next codegen run), I taught the SQL template to generate them from the model:

- **New org model subsection `** Soft FK validations`**: each child heading is a soft-FK entry with `:table:` and `:error_message:` properties. `org_loader.py` converts these into `domain_entity.sql.soft_fk_validations`, which the existing SQL template block already consumed.
- **New org model subsection `** Checks`**: an org table with an `expression` column populates `domain_entity.sql.extra_checks`, which the existing SQL template `{{#extra_checks}}` block already consumed.
- **Column `:default_value:` property**: used to restore `= 0.0` on all `double` fields and `= "geometric"` on `process_type`.
- **Natural key corrections**: added `component_index` to `gmm_component` natural keys (unique index now `(tenant, party, fx_spot_config, component_index)`); added `base_currency_code` + `quote_currency_code` to `fx_spot_generation_config` natural keys (unique index now `(tenant, party, config, base, quote)`).

Both entities regenerated cleanly. The `** Soft FK validations` and `** Checks` subsections are now available to all future org-model entities.

PR #1378, opened 2026-06-30
[compass] Compass quality of life — Sprint 21 (search)

mcraveiro:
## Summary

Third round of compass improvements, focused entirely on making `compass search` a reliable research tool for LLM agents.

### Bucketed search results by dimension (task 5DFFB2BB)

- Four fixed buckets (Recipes & how-to, Knowledge, Now, Past sprints) plus a dynamic tail for any doctype not covered — driven by `ui.BUCKET_AFFINITY` so new doctypes are picked up automatically without touching `compass.py`
- Each bucket independently selects its top-N from a large fetch pool (20× limit), preventing historic agile docs from drowning current results
- Past-sprint stories/tasks excluded by default (`--history` flag restores them)
- Heading-level anchors (subsections with `:ID:` but no `#+type:`) excluded by default — parent body FTS already covers their content

### Explicit weighted scoring model (task 8C8A7CBE)

New standalone module `projects/ores.compass/src/search_scorer.py` (no live-DB dependency, 21 unit tests):

- `QueryPlan` — tokenises query, strips stopwords, splits core words (question verbs removed) from full words
- `DocSignals` — all named signals: core-words title rank, full-words title rank, body BM25 proxy, inbound-link count, recipe-question bonus
- `Weights` — every coefficient and normalisation constant in one named dataclass
- `score_document()` — RRF (k=60) normalisation + linear combination → `ScoreResult` with `.pct` and `.breakdown`
- `global_floor()` — `max(absolute_floor, max_score × dropout_ratio)` applied globally before bucketing so a strong Recipes hit (80%) suppresses weak Knowledge hits (15%)
- `precision_at_k()` / `ndcg_at_k()` for corpus evaluation

### Search UX changes

- Score shown as `[74%]` per hit instead of opaque `[core:5]` labels
- Configuration block at top of every search: active sprint, floor, dropout ratio, override flags
- Tags line (`🏷`) on every hit; `📍` parent-path line for heading anchors
- `--floor PCT` and `--dropout RATIO` flags for live tuning without editing source
- `--all-buckets` disables the global dropout ratio
- Search header shows current sprint name and active exclusions

## Test plan

- [ ] `./compass.sh search "new sprint"` — should return recipe + runbook at ≥70%, nothing below 25%
- [ ] `./compass.sh search "how do i create a new sprint"` — same two docs lead, answer extract shown
- [ ] `./compass.sh search "new sprint" --history` — Past sprints bucket appears
- [ ] `./compass.sh search "new sprint" --all-buckets` — heading anchors + all buckets visible
- [ ] `./compass.sh search "new sprint" --floor 10 --dropout 0` — relaxed thresholds
- [ ] `python3 -m pytest projects/ores.compass/tests/test_search_scorer.py -v` — 21 tests pass
- [ ] New doctype added to `ui.BUCKET_AFFINITY` — appears in dynamic tail automatically


mcraveiro:
Fixed in f9c339754. All findings accepted and addressed:

**Bug** — overflow into dynamic tail: changed `for r in hits[:bucket_limit]` to `for r in hits` so all predicate-matched items are claimed before the dynamic tail loop runs.

**Moderate — body_bm25 default**: changed from 0.0 to -8.0 (= -bm25_scale → 0.0 body contribution). Updated field comment and renamed test to test_default_signals_score_zero with assertion pct==0.

**Moderate — Weights docstring**: updated to 'by convention, not enforced' and added `__post_init__` assertion that weights sum to 1.0.

**Moderate — STOPWORDS comment**: removed 'shared with compass.py'; both files now annotate that the two sets are intentionally distinct.

**Moderate — dead query parameter**: removed `query: QueryPlan | None = None` from `score_document`. Updated `pool_position` docstring to 'reserved for future use — not yet used in scoring'.

**Moderate — is_question divergence**: `question`/`how_do` in cmd_search now derived from `QueryPlan.from_query()` at query entry; `_qplan` reused in pretty-format block (eliminates double construction). Fixed `is_question` to use first-word approach (`q_words[0] in _QUESTION_STARTERS`) — matches the safe logic in the companion `_is_question()`.

**Minors**: test renamed; `import math` moved to module top; `full_words` comment updated (it's the hook for future divergence, not a true alias); `_sprint_prefix` alias removed; `--floor` help text drops `%(default)s`.

PR #1379, opened 2026-06-30
Fix duplicate natural key fields and codegen generator template gaps

mcraveiro:
## Summary

Fixes three bugs introduced in PR #1376 when promoting fields to natural keys:

- **Duplicate struct members**: `base_currency_code`/`quote_currency_code` in `fx_spot_generation_config` and `component_index` in `gmm_component` were present in both the Natural keys and Columns sections of their org models, generating duplicate C++ struct members and causing `rfl` static assertion failures at compile time. Removed the Columns duplicates.
- **Generator template `idx` undeclared**: The generator template appended `std::to_string(idx)` for all non-UUID natural keys but only declared `idx` for text-PK entities. Added a `has_text_natural_keys` flag in `core.py` and conditionally declare `idx` for UUID-PK entities that need it.
- **Generator template int + string concatenation**: Integer natural keys (`component_index`) used the string-append pattern, which is a type error. Added `is_int` detection to natural key processing in `core.py` and a dedicated template branch that emits just the generator expression.
- **Domain header missing `default_value` for natural keys**: The natural key section of the domain class header template didn't support `{{#default_value}}`; added the same branch used for columns so `component_index = 0` is emitted.

## Test plan

- [ ] Build `ores.synthetic` cleanly — no duplicate member errors, no `rfl` static assertions
- [ ] `gmm_component` generator produces structs with `component_index = 0` default
- [ ] `fx_spot_generation_config` generator produces unique `base_currency_code`/`quote_currency_code` values per instance

### DIFF
--- a/projects/ores.codegen/library/templates/cpp_domain_type_class.hpp.mustache
+++ b/projects/ores.codegen/library/templates/cpp_domain_type_class.hpp.mustache
@@ -66,7 +66,7 @@ struct {{entity_singular}} final {
      * {{{detail}}}
 {{/detail}}
      */
-    {{{cpp_type}}} {{column}};
+    {{{cpp_type}}} {{column}}{{#default_value}} = {{{default_value}}}{{/default_value}};
 
 {{/natural_keys}}
 {{/has_identity_group}}
--- a/projects/ores.codegen/library/templates/cpp_domain_type_entity.hpp.mustache
+++ b/projects/ores.codegen/library/templates/cpp_domain_type_entity.hpp.mustache
@@ -31,10 +31,10 @@ struct {{entity_singular}}_entity {
     int version = 0;
 {{#natural_keys}}
 {{#is_uuid}}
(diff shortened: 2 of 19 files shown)


mcraveiro:
Fixed in d6c4d63f6 and ff3e6c2e0. The entity header template now uses {{{cpp_type}}} for non-UUID natural keys (not unconditional std::string), so gmm_component_entity::component_index is int — no mapper type mismatch. The needs_counter flag (d6c4d63f6) guards the counter declaration so it is only emitted when actually used.

mcraveiro:
Both issues fixed: needs_counter flag added in d6c4d63f6 — counter declaration is now conditional on primary_key.is_text OR has_text_natural_keys, so gmm_component_generator.cpp no longer emits the unused variable. Entity header default_value gap fixed in ff3e6c2e0 — gmm_component_entity::component_index now has = 0.

PR #1408, opened 2026-07-03
[synthetic] Synthetic data librarian support: FX foundation

mcraveiro:
## Summary

Party creation now yields a usable, ticking synthetic FX spot configuration end to end, riding the existing DQ bundle-publish infrastructure rather than inventing a new mechanism.

## Changes

- Document the DQ bundle publication pipeline (new knowledge doc), including a worked example and a codegen-value assessment identifying the one real codegen gap: hand-written publish-from-dq functions.
- Hand-write the synthetic FX spot config's DQ artefact-type wiring (shadow table, publish function, NATS handler, registrar), mirroring the existing report_definitions instance.
- Seed EUR/USD and GBP/USD synthetic FX configs as members of the existing ore_analytics bundle (not a new bundle), so every provisioning path that already publishes reports picks them up.
- Fix a real bug found while testing: the generic stamp() helper clobbered account_party's target party_id with the caller's own party, silently breaking tenant-wide party association during provisioning.
- Fix a second bug: publishing an FX config never seeded a GMM component, leaving nothing to sample from.
- Split two pieces of follow-on work to the backlog rather than doing them here: a publish_from_dq codegen facet, and unifying report-definitions provisioning (both call sites still bypass the standard publish-from-dq path via a bespoke per-record RPC dance).
- Verified end-to-end via a real Barclays GLEIF provisioning run: tenant_admin correctly associated with all 81 operational parties, both FX configs created and actively publishing ticks over the existing ores.marketdata NATS path.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Synthetic data librarian support: FX foundation](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/synthetic_data_librarian_support/story.html) | E5EC80BB-789B-4BD6-B6E5-DC7AA790692A |
| Task | [Model synthetic FX spot config as a DQ artefact type](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/synthetic_data_librarian_support/task_model_synthetic_fx_config_as_dq_artefact.html) | 44030333-A650-4EBF-9674-5819B8496114 |


mcraveiro:

mcraveiro:
Fixed in 53db445 — captured the replaced fx-config ids in a temp table before voiding them, then void the matching `ores_synthetic_gmm_components_tbl` rows too. Verified against the live dev DB: a `replace_all` publish now leaves zero orphaned GMM rows pointing at a closed `fx_spot_config_id`.

mcraveiro:
Addressing each point:

1. **No unit test for stamp_account_party** — Declined for now. `ores::database::context` requires a real `sqlgen::ConnectionPool`, and there's no lightweight/mockable construction path used anywhere in this codebase's test conventions (checked `projects/ores.iam/core/tests/` — only domain-object and repository/service-level tests exist, no handler-level tests, matching the reporting equivalent you note). Adding one cleanly would mean building new test infrastructure, which is bigger than this fix warrants. Agreed it's a real gap worth closing given the bug was silent and tenant-wide — happy to file it as a follow-up capture if you'd like, rather than block this PR on new test infra.
2. **subject_to_fn allow-list** — Fixed in 53db445: added an explicit `is_known_fn()` allow-list checked before the SQL function name is used, so this stays safe if a wildcard subscription is ever added.
3. **Version churn on every GMM publish** — Not changed. As you note, this matches the existing unconditional-insert pattern immediately above it for `fx_spot_generation_configs` in the same function — consistent with the surrounding code rather than a new problem introduced here.

mcraveiro:
Addressing each point:

1. **replace_all orphans GMM components** — Fixed in 53db445: capture the replaced fx-config ids in a temp table before voiding them, then void the matching `ores_synthetic_gmm_components_tbl` rows too. Verified against the live dev DB: zero orphaned GMM rows after a `replace_all` publish.
2. **Phase 3.5 / [4/5] mismatch** — Fixed in 53db445: renumbered the comments to Phase 4/Phase 5 to match the printed `[4/5]`/`[5/5]` counters.
3. **Container name from alphabetically-first artefact row** — Not changed. Cosmetic, and matches this story's MVP scope (single-pair-is-enough); worth a proper decision once more pairs land, not before.
4. **No automated tests** — Same as raised on another review comment: `ores::database::context` needs a real connection pool, no lightweight mock path exists in this codebase's conventions today, and this matches the untested reporting equivalent. Agreed it's worth a follow-up given the bug was security-relevant; not blocking this PR on new test infra.

mcraveiro:
Follow-up: reconsidered — this codebase does have DB-backed test infra suitable for this (`database_helper`, `h.set_party()`), so declining for lack of a mock path wasn't right. Added a real regression test in 715f58b: `stamp_account_party_preserves_target_party_id` in `projects/ores.iam/core/tests/messaging_account_party_handler_tests.cpp`, simulating the exact scenario that exposed the bug (caller's context scoped to one party while targeting another). All 157 assertions pass.

PR #1409, opened 2026-07-03
[synthetic] GMM improvements: chart axes, per-component PDFs, Ornstein-Uhlenbeck

mcraveiro:
## Summary

Tidy up loose ends from the GMM-based synthetic market data generation work: readable chart axes, per-component PDF display, and a new Ornstein-Uhlenbeck process engine.

## Changes

- Fix invisible axis labels on the GMM preview charts (dark-theme QWidget::palette() bug — app theme is QSS-only, never QApplication::setPalette)
- Render each GMM component's own PDF curve with a matching colour-swatch indicator, alongside the combined mixture PDF in a distinct colour
- Implement Ornstein-Uhlenbeck as a new synthetic price-process engine end-to-end (ou_process, process_factory dispatch, SQL constraint, UI)
- Fix change-reason-code string literals to use existing ores.dq.api constants (reported during the session)
- Set source/point_id (tenor) on synthetic market observations, previously left empty (reported during the session)
- File backlog captures for out-of-scope process types (jump/Poisson, cross-rate correlation) and an unrelated SQL-codegen template drift

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [GMM improvements: tidy up synthetic data generation loose ends](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/gmm-improvements/story.html) | F4604E72-EF59-4894-A6D4-A68971C793EB |
| Task | [Add X/Y axis labels to GMM preview charts](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/gmm-improvements/task_gmm-chart-axes.html) | 7CB8B816-BE19-4E02-B0C0-7A800EAD6344 |

**Environment:** #+todo: DISCOVERED BACKLOG STARTED BLOCKED | DONE ABANDONED

### DIFF
--- a/doc/agile/product_backlog/inbox/codegen_sql_validation_fn_drift_synthetic.org
+++ b/doc/agile/product_backlog/inbox/codegen_sql_validation_fn_drift_synthetic.org
@@ -1,0 +1,52 @@
+:PROPERTIES:
+:ID: D583B57B-B896-48A6-AD66-1F01FAE3FA3D
+:END:
+#+title: Reconcile SQL codegen template drift on synthetic validation fn
+#+description: Regenerating the fx_spot_generation_config SQL layer drops the tracked system-tenant fallback logic from its validation function, indicating the template has drifted from the checked-in output.
+#+type: capture
+#+level: cross
+#+filetags: :codegen:sql:synthetic:drift:
+#+created: 2026-07-03
+#+updated: 2026-07-03
+
+This page is a [[id:671F18E4-E09C-4B3B-BD24-D33DF8AE38A6][capture]] in the *inbox* bucket of the product backlog — a pre-sprint idea, not yet pulled into a sprint as a story.
+
+* What
+
+Regenerating =projects/ores.synthetic/modeling/ores.synthetic.fx_spot_generation_config.org=
+via =compass codegen generate --model ... --profile sql= (done to add
+=ou= to the =process_type= check constraint for the GMM-improvements
+story) produced a diff that, besides the intended check-constraint
+change, also silently dropped custom validation logic from the
+tracked =ores_synthetic_validate_fx_spot_generation_config_fn()=
+function in
+=projects/ores.sql/create/synthetic/synthetic_fx_spot_generation_configs_create.sql=:
+the "system tenant canonical set" pass-through/fallback branch (the
+block starting "Allow pass-through if neither this tenant nor the
+system tenant has seeded active fx_spot_generation_configs yet...")
+was replaced by a simpler null-check-only body. Either the
+=sql_schema_domain_entity_create.mustache= template (or the model's
+=* Validation function= section) has drifted behind manual edits
+previously made to the generated file, or the template itself
+regressed. The regeneration was reverted; only the single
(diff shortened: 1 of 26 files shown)


mcraveiro:

mcraveiro:
Fixed in 1d07dae74 — onEngineChanged now forces Advanced mode and disables the Simple toggle while a single-regime engine ('ou') is selected, so Simple mode can no longer produce a kappa derived from unrelated drift/vol/jump sliders. Also replaced the scattered isOuEngine() checks with a data-driven EngineInfo table (code/label/supportsMixing) that drives the generic mixing-related gating, per a separate suggestion from the user to avoid hardcoding UI knowledge of specific process types — this generalizes cleanly for future single-regime engines (e.g. rate processes).

mcraveiro:
Fixed in 1d07dae74 — added a one-line comment on QLineSeries::setName() in ReturnDistributionChart.cpp noting the chart's legend is hidden and the colour-swatch column in FxSpotRateEditor's component table is the intended colour-to-component indicator. Left the misleading refreshCharts() comment as-is since a follow-up commit (1d07dae74) already rewrote that region; a unit test for ou_process is noted as a known gap in the task doc but not added, matching gmm_process/arithmetic_gmm_process's existing untested state — no established Catch2 convention exists for this directory yet.

mcraveiro:
Addressed observation 1 (silent, irreversible loss of mixture components) as part of the broader fix in 1d07dae74: onEngineChanged now forces Advanced mode when switching to a non-mixing engine, so the collapse to one row is immediately visible/editable rather than silent — the user sees exactly what happened and can fix values before Save. Declining observation 2 (duplicated theme colour constants) for this PR as you suggested — filed nowhere yet, will capture separately as a follow-up backlog item to hoist into a shared chart-theming helper. Observation 3 (Weight/tooltip still say 'Weight' for OU) — left as-is; the inline warning banner already documents the remapping, consistent with the same tradeoff the 'arithmetic' engine already makes for its reused fields.

mcraveiro:
Fixed in 247eccda3 — moved the forced-Advanced-mode logic out of onEngineChanged() (which only fires on the combo's currentIndexChanged signal) into updateEngineUi(), which already runs both on every engine change AND once at construction (end of buildBehaviourTab()), after engineCombo_->setCurrentIndex() has set the loaded record's engine. Opening an existing 'ou' record now correctly forces Advanced mode and disables the Simple toggle at construction time, closing the silent-kappa-corruption path you traced across the last three review rounds. Thanks for staying on this — the repeated re-checks caught a real gap I'd missed.

mcraveiro:
Same fix as my reply on the later round-3 comment: 247eccda3 moves the forcing logic into updateEngineUi(), which runs at construction time too, not just on live combo changes. Fixed.

mcraveiro:
Same fix, see 247eccda3 — thanks for the precise construction-order trace, that's exactly what was wrong.

mcraveiro:
Confirmed, thanks for tracing the full construction order. Agreed the double refreshCharts() call is harmless — not worth a follow-up.

mcraveiro:
Confirmed, thanks for the verification.

PR #1417, opened 2026-07-03
[ci] Trigger canary run for sccache zero-diff experiment

mcraveiro:
## Summary
- Tiny trigger PR: touches canary-linux.yml (comment only) to get a fresh canary run with the SCCACHE_LOG=debug instrumentation landed in #1414, so the zero-source-diff experiment (task analyze_sccache_zero_diff_experiment) has a "run 1" to rerun for "run 2".

## Test plan
- [ ] Canary run 1 (this PR) completes
- [ ] `gh run rerun` on the same run for zero-diff run 2
- [ ] Compare sccache hit rates

PR #1414, opened 2026-07-03
[agile] Correct canary capture into sccache-cache-invalidation story

mcraveiro:
## Summary
- Corrects a mis-framed capture: PR #1410's canary job wasn't stuck on cancellation, it just took a genuine ~3h to run. The real finding is a 1.36% sccache hit rate (33/2433 compile requests) on a PR touching ~20 files.
- Documents the evidence gathered (sccache stats, PCH ruled out, vcpkg submodule pointer confirmed stable since 2026-06-14, vcpkg.json manifest changes identified in the 24h before the slow build) and scopes a task to prove/disprove the causal link with a live sccache debug-log comparison.

## Test plan
- [x] Docs-only change (story + task scaffolding), no code touched.

PR #1418, opened 2026-07-03
[ci] Root-cause canary's sccache thrashing: cross-platform 10GB quota contention

mcraveiro:
## Summary
- Corrects the story's earlier premise (again): the max-size bump from PR #1414 didn't fix anything — a follow-up trigger PR (#1417) got a 0.00% sccache hit rate with the restore step reporting `Cache not found` entirely.
- Root cause, found via the GitHub cache API directly: GitHub's `actions/cache` is capped at a fixed, non-purchasable 10 GB per repository, shared across every OS/compiler/buildtype variant (linux-gcc/clang, windows-clang, macos-clang × debug/release) plus vcpkg/Qt caches. The Linux-gcc-debug cache that canary depends on had been evicted to make room for other platforms' caches touched more recently by `continuous-*.yml`.
- Immediate mitigation: stop persisting sccache across runs for windows/macos continuous builds (they run on schedule/tag push, not gating every PR) so Linux — which gates every PR via canary — gets the shared quota to itself. Deleted the existing windows/macos sccache entries (~3.6GB) for instant effect.
- Reverted the earlier (counterproductive) max-size bump and removed the SCCACHE_LOG=debug instrumentation, which added no diagnostic value.
- Filed a backlog capture recommending the durable fix: move sccache to an external cache backend (S3/GCS/Azure/self-hosted) that isn't subject to GitHub's shared 10GB quota — out of scope for this pass.

## Test plan
- [ ] This PR's own canary run should now find and use the (currently-empty, freshly-created) Linux-gcc-debug cache without competing against other platforms for eviction.


mcraveiro:

mcraveiro:
Fixed in 7c4252251 — filled in the What/Why/References/See also sections using the content already in the description and the story's Decisions section, matching the convention of other inbox captures.

PR #1420, opened 2026-07-03
[agile] Appraise counterparty: regen unsafe, validation logic missing from models

mcraveiro:
## Summary
- Appraised `counterparty` for commissioning (docs-only, no code changes). Confirmed it's already on the unified codegen model (ahead of `currency`, which still has a legacy `_table.org` companion).
- Attempted a scoped regen per the standard recipe — result was severely destructive on the parent table and both children (`counterparty_identifier`, `counterparty_contact_information`): hand-written SQL validation (FK checks, business rules, a cardinality limit, even a whole column) isn't captured in any of the three `.org` models, so regen silently deletes it. Reverted immediately, nothing destructive committed.
- Scaffolded a follow-up task to author the missing model sections before any further regen is attempted.
- Per direction, dropped the per-entity Wt/HTTP capture item from this story — that's tracked by the existing dedicated top-level commissioning stories instead.

## Test plan
- [x] Docs-only change; no code/SQL/generated files touched.


mcraveiro:

mcraveiro:
Fixed in 59787a311 — verified both findings directly (business_center_code's call-and-reassign pattern in the live SQL, and ores_refdata_validate_country_fn's actual null/empty-only body + zero call sites) before accepting. Reclassified business_center_code to the Validations table facet, and corrected the country_code claim to an explicit open gap rather than 'nothing new to build'. Also backfilled #+pr:/the PRs table on the appraisal task.

PR #1437, opened 2026-07-04
[ore,marketdata,shell,qt] Fix duplicate-key import failure with a warn/error policy

mcraveiro:
## Summary

Investigated the market.txt duplicate-observation-key bug filed while testing the ores.shell marketdata import command. Root cause was NOT the suspected series_key_registry decomposition collision: Legacy/Example_56's market.txt genuinely repeats ~100 (date, key) pairs with a different value later in the file. Adds duplicate handling to the importer itself (ores.ore.core), with a warn/error policy and a report, so callers (shell, Qt) know what was de-duplicated instead of it happening silently or failing outright.

## Changes

- parse_market_data/parse_fixings (ores.ore.core) gain duplicate_policy (warn/error) and parse_report (warnings/errors), de-duplicating repeated (date, key) pairs last-line-wins
- import_market_data_request gains duplicates_are_errors (default false); import_market_data_response gains warnings/errors. In error mode, affected content is not persisted
- ores.shell marketdata import and Qt ImportTradeDialog both now print/log warnings and errors instead of silently dropping them
- Verified live: Legacy/Example_56 now imports cleanly (176 series, 7777 observations, 103 warnings — exact match against the file's raw duplicate count); TA002 IR Swap/MarketRisk HistSimVar re-verified with no regression; synthetic duplicate fixings.txt confirmed the fixings-side fix keeps the later value
- 6 new unit tests added; all existing market_data_parser tests pass unchanged

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Full market.txt import hits duplicate observation key](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/full_market_txt_import_duplicate_key/story.html) | C1BE59FA-F9B6-4482-A5C0-868D37D47504 |
| Task | [Fix series key decomposition collision causing duplicate observation key](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/full_market_txt_import_duplicate_key/task_fix-series-key-collision.html) | 3AB326D6-F7CC-41A9-B69E-BF8804C1E182 |


mcraveiro:

mcraveiro:
Fixed in d1907cb14 — (1) shell and Qt now print series/observation/fixing counts alongside errors on partial failure, so a caller doesn't misread 'nothing happened'. (2) Added service_import_service_tests.cpp with 2 tests covering warn-mode dedup+counts and the exact mixed-section error-mode behavior you flagged (only the affected section skips persistence; the clean section's count still surfaces even though overall success=false).

mcraveiro:
Fixed in d1907cb14 — regenerated import_legacy_example_56.ores via tangle_shell_scripts, drift check should be green now. Also addressed the partial-success point in the same commit (see reply on the other review comment for detail).

mcraveiro:
Fixed in d1907cb14 — partial-import counts are now surfaced by both callers on failure, and added 2 service-level tests (database_helper-backed) exercising duplicates_are_errors=true with a mixed clean/errored section, confirming the clean section persists and its count is reported. Not changing the double-composite-construction point separately -- folded that simplification into the same commit (composites computed once, reused for both passes).

mcraveiro:
Declining the two minor nits: the sys_days conversion is intentional (clearer date-arithmetic intent even though year_month_day has comparison operators), and the +/to_string concatenation matches this file's existing style (see the exception-message concatenation above it) rather than being a new deviation.

PR #1441, opened 2026-07-05
[refdata,qt,build] Relocate currency generator logic; fix CI cache waste

mcraveiro:
## Summary

Move currency's random-pick generation out of the Qt UI into the generator layer (step 1 of the Sync Qt codegen for currency regenerate plan), and remove a CI build-directory cache that was silently forcing full rebuilds while degrading sccache's own hit rate.

## Changes

- Add ores::refdata::generators::generate_random_fictional_currency(); CurrencyDetailDialog::onGenerateClicked() now calls it directly instead of embedding std::uniform_int_distribution selection
- Remove the ineffective 'Restore/Save cmake build cache' steps from continuous-linux.yml (gcc-debug) and canary-linux.yml (gcc-release) — checkout always resets source mtimes newer than the restored objects, so it forced full rebuilds anyway while correlating with gcc-debug's sccache hit rate collapsing to ~4%
- Bump sccache max-size from 2000M to 2500M in both workflows

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Relocate random-pick generator logic out of currency Qt UI](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission_currency/task_relocate-currency-qt-generator-logic.html) | 78BF95FE-1513-4556-BCBC-8A4A1AF48907 |


mcraveiro:

mcraveiro:
Fixed in d171a7f7b — added generate_random_fictional_currency_returns_known_currency test case, following the file's existing per-function convention. Left the 'Last touched' date as-is (it's correct for when the doc was actually last edited, just after the PR's initial creation timestamp).

mcraveiro:
Fixed in d171a7f7b — added a test case for generate_random_fictional_currency() covering the same 'known ISO code' assertion pattern used elsewhere in the file.

mcraveiro:
Fixed in d171a7f7b — added a unit test for generate_random_fictional_currency(). Left the 'materializes the full 50-entry vector to pick one' performance nit as a non-blocking follow-up per your own note; it's unchanged behavior from the original inline code, not a regression.

PR #1442, opened 2026-07-05
[refdata] Move and commission business_day_convention_type

mcraveiro:
## Summary

Move business_day_convention_type from ores.trading to ores.refdata and commission it fully, since every other *_convention entity it belongs alongside already lives there.

## Changes

- Model business_day_convention_type fresh in ores.refdata via ores.codegen.entity (code/name/description/display_order, matching the currency_market_tier aux-type shape)
- Generate SQL, domain, repository, service, messaging, and Qt layers; wire into refdata registrar, LISTEN/NOTIFY-to-NATS relay, and SQL create/drop manifests
- Fix two latent bugs in the pre-existing Qt layer: invalid Icon::Calendar enum value and a delete-request field mismatch
- Remove the ores.trading original in full (SQL scripts, model, generated C++, hand-written NATS/CLI wiring) and repoint trading_swap_legs' validation at the new ores.refdata function

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Commission: business_day_convention_type](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission-business-day-convention-type/story.html) | 0345DCE3-4B85-4132-9A25-E58285632F76 |
| Task | [Move business_day_convention_type to ores.refdata and commission fully](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission-business-day-convention-type/task_move-and-commission-business-day-convention-type.html) | F7C39C66-DFA0-4A10-B473-F3827B8A7187 |


mcraveiro:

mcraveiro:
Fixed in 2f155a7ff — restored the existence-check + errcode 23503 validation, matching ores_refdata_validate_currency_market_tier_fn's pattern, removed the duplicate change_reason_code call, added the missing drop function statement, and de-duped the synthetic-counter increment in the generator. Verified locally: DB recreate, direct SQL calls for null/invalid/valid business_day_convention_code values, and a full rebuild of ores.refdata.api.lib.

Root cause for the validate-fn regression: a genuine codegen bug, not something specific to this entity — sql_schema_domain_entity_create.mustache (the archetype that actually fires for domain_entity SQL creates) reads domain_entity.validation_fn.scope_both/system/tenant, but those flags are only ever computed on the projected 'table' context used by the legacy table-model path, never on the raw domain_entity dict. Confirmed by regenerating currency_market_tier (a known-good sibling) with the live codegen tool — it reproduces the identical bug. Hand-fixed the SQL here rather than regenerating (regeneration would have broken it the same way), and filed a separate capture to track the actual codegen fix: doc/agile/product_backlog/inbox/codegen_domain_entity_validation_fn_scope_missing.org

PR #1445, opened 2026-07-06
[codegen,qt] Land CSV/XML import-export as a qt-profile capability

mcraveiro:
## Summary

Generalise currency's hand-rolled CSV export and XML import/export into an opt-in mdi_window/client_model template capability (step 2 of the currency Qt regenerate plan), so the eventual codegen regenerate doesn't silently drop it.

## Changes

- Add generic ImportEntityDialog (ores.qt/api), replacing the currency-specific ImportCurrencyDialog; Qt moc can't process class templates with Q_OBJECT, so it's a single concrete dialog driven by type-erased rows and label_of/import_one callbacks
- Add has_csv_xml_io opt-in flag to the qt-profile mdi_window/client_model mustache templates: toolbar actions (Import XML, Export CSV, Export XML), generated exportToCSV/exportToXML/importFromXML wired to the entity's csv/xml exporter/importer by naming convention, and a get<Plural>() bulk accessor on the client model
- Derive import_preview_columns (org_loader.py) from the existing columns list, filtering out audit/system fields (version, modified_by, recorded_at, ...) meaningless before import, rather than a second hand-authored column table per entity
- Flip has_csv_xml_io on for currency; verified by diffing the regenerated has_csv_xml_io sections against the hand-migrated code — match modulo formatting

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Land CSV/XML import-export as a qt-profile template capability](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission_currency/task_land-csv-xml-export-qt-template.html) | 2EAF2F99-B71F-410D-985A-7A616E40CB88 |


mcraveiro:

mcraveiro:
Fixed in 263d90206 — replaced the exclude-list derivation with an opt-in :import_preview: true column property on the Qt 'Columns (Qt model)' table (passed through by the existing generic column parser, no python code change needed there). Currency's model now marks exactly the 5 fields the hand-migrated dialog shows (iso_code, name, symbol, fraction_symbol, fractions_per_unit); re-ran codegen entity generate currency --address ores.cpp.qt --diff and confirmed the regenerated preview table now matches — this time the verification actually catches a mismatch if one exists, since the opt-in list only grows when an entity author explicitly adds a column. Left the 'no unit test for org_loader.py' note as a non-blocking follow-up (there's no existing test file for org_loader.py at all — a first one would be a bigger lift than this fix).

mcraveiro:
Follow-up in 263d90206: a later review pass caught that import_preview_columns' exclude-list logic over-included fields (11 vs the 5 the hand-migrated dialog actually shows). Replaced it with an opt-in per-column flag — see the other review thread for details.

mcraveiro:
Follow-up in 263d90206: same as above — the exclude-list derivation was replaced with an opt-in :import_preview: column flag after a later pass found it over-including fields for currency.

PR #1446, opened 2026-07-06
[codegen,qt] Land setting-gated action-visibility as a qt-profile capability

mcraveiro:
## Summary

Generalise currency's duplicated generate-button visibility mechanism (subscribe/notify/query a boolean system setting) into a reusable controller and opt-in mdi_window/detail_dialog template capability (step 3 of the currency Qt regenerate plan).

## Changes

- Add SettingGatedActionController (ores.qt/api): one controller instance gates several QActions across different settings with a single NATS subscription and a single list-settings fetch, instead of duplicating the ~50-line mechanism per action
- Migrate CurrencyMdiWindow and CurrencyDetailDialog's generate-button visibility to the shared controller, deleting the duplicated updateGenerateActionVisibility/onSystemSettingNotification/onConnectionEstablished methods from both
- Add has_setting_gated_actions opt-in flag (a 'Setting-gated actions' Qt table: action/setting rows) to the mdi_window/detail_dialog mustache templates, wiring the controller construction and per-action registration generically
- Flip has_setting_gated_actions on for currency (generateAction / system.synthetic_data_generation); verified by diffing the regenerated gated-action sections against the hand-migrated code

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Land setting-gated action-visibility knob as a qt-profile template capability](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission_currency/task_land-setting-gated-visibility-qt-template.html) | D9CB6853-797A-4144-BED0-AEBDDBBC173C |


mcraveiro:

mcraveiro:
Fixed in b7176c782 — added an optional guard predicate to SettingGatedActionController::registerAction(), ANDed with the setting's value before setVisible(). CurrencyDetailDialog passes [this]{ return !isReadOnly_; }, and the generic detail_dialog template does the same with its own readOnly_ member, so future has_setting_gated_actions consumers get the same protection. Also removed the QTimer::singleShot deferral around the initial refresh() call in both hand-written files to match the template's synchronous call (confirmed it was leftover scaffolding, not load-bearing) — resolves the divergence you flagged too. Left the double-construction-in-setClientManager nit as-is per your own assessment (not live today, no current call site reuses a dialog instance).

mcraveiro:
Fixed in b7176c782 — added an optional guard predicate to registerAction() so CurrencyDetailDialog's read-only state is now rechecked on every refresh, including notification-triggered ones (matches the old per-dialog behavior). On point 2 (isConnected vs isLoggedIn): kept the unification — list_settings_request goes through process_authenticated_request either way, so isLoggedIn() is the more correct gate and no caller depended on the looser isConnected() check. Point 3 (QFutureWatcher include) — agreed, no change, matches existing codebase convention.

mcraveiro:
Thanks — the read-only edge case you both flagged (and I'd already noted in the task doc) is fixed in b7176c782 via an optional guard predicate on registerAction(). Left the unit-test suggestion as a follow-up rather than blocking this PR, since the interesting paths need a live ClientManager/NATS connection as you noted.

mcraveiro:
Follow-up in 6b05cb53f: ran clang-format on SettingGatedActionController.hpp/.cpp (continuation-line indentation nit) and updated the task doc's implementation notes, which still described the read-only gap as an open tradeoff after b7176c782 already fixed it. Thanks for confirming the fix closes the regression.

PR #1449, opened 2026-07-06
[ores.refdata,ores.qt,ores.codegen] Sync C++ core codegen output for book

mcraveiro:
## Summary

Regenerates book's api/core/sql from the graph-based codegen and moves its Qt layer from ores.qt/trading to ores.qt/refdata to match, then fixes the resulting cross-plugin wiring and several bugs surfaced by manual UI testing. Along the way, lands four reusable codegen capabilities (cross-plugin export macros, change-reason-cache wiring, a fix for uuid-typed key fields, and the explorer-API/parent-entity wiring needed by Portfolio/Org explorer windows) plus a shared HistoryDialogBase fix that closing a history window doesn't accidentally close the whole app. Also restores five soft-FK/value validations on Book's insert trigger that the initial regeneration silently dropped (found by review).

## Changes

- Regenerate book's C++ core (api domain/messaging, core repository/service/handler) and SQL from the new graph-based codegen; db_timestamp, delete_* naming, pagination
- Move Book's Qt layer (BookController, BookDetailDialog, BookHistoryDialog, BookMdiWindow, ClientBookModel) from ores.qt/trading to ores.qt/refdata, matching the new codegen output location
- Fix cross-plugin symbol-lookup crash: BookController needed ORES_QT_REFDATA_EXPORT since TradingPlugin constructs it but it lives in libores.qt.refdata.so (built with -fvisibility=hidden)
- Fix book_changed_event field mismatch (e.ids -> e.book_ids) that broke the build
- Wire has_change_reason_cache for Book so saves can prompt for a change reason (was missing entirely, blocking all saves)
- Make Book's name renamable after creation (was wrongly modelled as an immutable is_key field) and surface the real UUID id, read-only, in the detail dialog
- New codegen capability: has_export_macro, an opt-in cross-plugin export annotation for Qt classes
- New codegen capability: has_explorer_api / parent_entity_singular, wiring openAdd/openEdit/openHistory entry points that PortfolioExplorerMdiWindow/OrgExplorerMdiWindow call into
- Fix codegen bug: is_key + uuid detail fields were treated as plain strings (missing primary-key lookup, unconditional std::string assignment) -- would not compile for any future uuid-typed key field
- Fix shared ores.qt/api bug: HistoryDialogBase's Close button called window()->close(), closing the whole app when embedded in an MDI subwindow instead of just the history dialog; added closeRequested()/requestClose() mirroring DetailDialogBase
- Restore five insert-trigger validations on Book (parent_portfolio_id/party_id/owner_unit_id soft FKs, ledger_ccy/book_status value checks) that the initial regeneration silently dropped because the model never had a `* Foreign keys` / `* Insert trigger` section; found independently by three automated PR reviews

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Commission: book](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission_book/story.html) | 04BE9B6F-B43C-426B-96D1-AED89F88FA93 |
| Task | [Sync C++ core codegen output for book](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission_book/task_sync_cpp_core_codegen_book.html) | 9AB461BD-AD46-4980-A801-BE6104270C52 |


mcraveiro:

mcraveiro:
Fixed in 81d0d4742 — restored the missing `* Foreign keys` (parent_portfolio_id -> portfolios, party_id -> parties, owner_unit_id -> business_units) and `* Insert trigger / ** Validations` (ledger_ccy -> ores_refdata_validate_currency_fn, book_status -> ores_refdata_validate_book_status_fn) sections in ores.refdata.book.org, mirroring ores.refdata.counterparty.org, then regenerated the SQL. Diffed the regenerated trigger against the pre-regen hand-written version — semantically identical (only whitespace formatting differs, plus the separately-legitimate workspace_id validation already present via the current template). Verified by a full `compass db recreate` applying cleanly with no errors. Also updated the PR description to call out the has_explorer_api/parent_entity_singular capability that was landed but not previously listed.

mcraveiro:
Fixed in 81d0d4742 — same fix as the parallel review pass on this PR (see the reply on the other review comment for full detail): restored the * Foreign keys / * Insert trigger sections in ores.refdata.book.org mirroring ores.refdata.counterparty.org, regenerated, diffed against the pre-regen SQL (semantically identical), and verified with a full db recreate.

mcraveiro:
Fixed in 81d0d4742 — same root cause and fix as the other two review passes on this PR. Also updated the PR description to explicitly call out has_explorer_api/parent_entity_singular as landed. On the other nits: the double-blank-line artifacts in book_mapper.cpp/book_repository.cpp predate this PR entirely (present in the very first regen commit, 40e36f0f2, before any of this session's changes) and match the identical pre-existing pattern in counterparty_mapper.cpp/counterparty_repository.cpp — a pystache template whitespace quirk affecting every entity, not something introduced here; leaving it out of scope rather than risking a much larger unrelated diff. Test coverage gap for Book noted as accurate but pre-existing (matches every other refdata entity currently) — not fixing in this PR.

PR #1451, opened 2026-07-06
[refdata] Migrate purpose_type to ores.codegen.entity; fix codegen template bugs

mcraveiro:
## Summary

Migrates purpose_type from the legacy ores.codegen.table model to ores.codegen.entity so it gets a changed-event type, client-side NATS subscription, and generated Qt/repository/service stack — this had been silently missing eventing entirely. Along the way, fixes two pre-existing codegen template bugs discovered while regenerating (duplicate change_reason_code validation emission; lost validation_fn tenant-scope flags that silently dropped the FK-existence check body). Also hand-wires purpose_type into application.cpp's Postgres LISTEN/NOTIFY-to-NATS pipeline as a stopgap; a follow-up story generates this per-entity instead.

## Changes

- Migrate ores.refdata.purpose_type_table.org (ores.codegen.table) to ores.refdata.purpose_type.org (ores.codegen.entity), regenerate the full stack
- Fix domain_entity SQL template: skip auto change_reason_code emission when the model already declares it explicitly, preventing a duplicate NEW.change_reason_code assignment
- Fix domain_entity_to_table_context/normalise_sql_table_context: sync computed validation_fn scope flags back onto domain_entity so the SQL template's bootstrap-passthrough and FK-existence validation body is no longer silently dropped
- Wire purpose_type into application.cpp's entity change-event pipeline (register_mapping + event_bus.subscribe) as a stopgap
- File follow-up stories: relocate book/book_status/portfolio to ores.trading; generate the per-entity NATS event-mapping registrar via codegen instead of hand-wiring application.cpp
- Record a memory: compass client start --instance-name defaults to the environment label; vary --colour only for parallel clients

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Commission: purpose_type](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission_purpose_type/story.html) | DCB5D77C-3BDA-40C1-A74A-F11F9A8C7FEC |
| Task | [Verify purpose_type Qt UI end-to-end post-NATS](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission_purpose_type/task_verify_qt.html) | 324912C7-DBE0-44CB-98E9-7DBF74A31D25 |


mcraveiro:

mcraveiro:
Both findings fixed in 95759a487:

1. **Registrar dead code** — replaced the inline hand-wired `purpose_type` block in `registrar.cpp` with `append(register_purpose_type_handlers(nats, ctx, verifier))`, matching `party_type`/every other migrated entity. Dropped `purpose_type` from the stale "no sub-registrar" comment.

2. **Pagination silently ignored** — root cause was the model missing the `has_pagination` Qt flag that `client_model_impl`'s template gates the `request.offset`/`request.limit` assignment on (defaults to `false`). Added `:has_pagination: true` to `purpose_type.org` and regenerated rather than hand-patching the generated file — this also fixed a second latent bug in the same block where `total_available_count` was computed from the page's local result size instead of the server-reported total.

Not adding regression tests for the two codegen template fixes in this PR — logging it as a follow-up rather than expanding this PR's scope further; happy to file that as a capture if useful.

PR #1454, opened 2026-07-07
[agile] Break out shell and CLI tasks for Commission: book

mcraveiro:
## Summary

The "Sync C++ core codegen output for book" task closed as the only task under the "Commission: book" story, but the story's acceptance covers far more (shell/CLI commands, manual chapter, Qt eventing/delete verification, Wt/HTTP backlog captures). Breaks out the shell and CLI task pair first, following the party_status precedent. Side-loads an unrelated CI fix: the Windows continuous build schedule was firing more often than a single run takes to complete.

## Changes

- Add task: Implement shell list/add/remove commands for book
- Add task: Implement CLI list/add/remove commands for book
- Widen the Windows continuous build cron from every 4h to every 12h. The last 20 scheduled runs took 3.5-7.8h (median ~6h) -- routinely longer than the 4h window, so the next trigger was firing before the previous run finished. Because the concurrency group uses `cancel-in-progress: false`, overlapping runs queued behind each other instead of being superseded, compounding the backlog.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Commission: book](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/commission_book/story.html) | 04BE9B6F-B43C-426B-96D1-AED89F88FA93 |

### DIFF
--- a/.github/workflows/continuous-windows.yml
+++ b/.github/workflows/continuous-windows.yml
@@ -23,7 +23,11 @@ env:
 
 on:  # yamllint disable-line rule:truthy
   schedule:
-    - cron: '0 */4 * * *'
+    # Recent runs typically take 5-8h (matrix: msvc/clang x debug/release);
+    # a 4h window meant the next scheduled trigger routinely fired before
+    # the previous run finished, queueing overlapping runs behind the
+    # per-matrix concurrency group instead of replacing them.
+    - cron: '0 */12 * * *'
   push:
     tags:
       - '*'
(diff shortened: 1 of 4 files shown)


mcraveiro:
Fixed in d0f74afc — dropped the extra shell/cli #+filetags layer to match the party_status precedent exactly.

mcraveiro:
Fixed in d0f74afc for both nits: dropped the extra shell/cli filetags layer, and narrowed story.org's Now/Next fields to no longer list shell/CLI as still-to-break-out (updated Last touched too). On the bundled CI fix: that was deliberate, per explicit user instruction to side-load it into this PR rather than open a separate one — not an oversight, but noted for the record.

PR #1455, opened 2026-07-07
[refdata,ore,qt,cli] Migrate fx_convention consumers to currency_pair and retire it

mcraveiro:
## Summary

Retires ores.refdata.fx_convention in favour of currency_pair/currency_pair_convention across ores.ore, the Qt refdata plugin, and the CLI.

## Changes

- conventions_mapper::map_fx/reverse_fx now map to a mapped_fx struct pairing currency_pair + currency_pair_convention instead of fx_convention
- SpotDays round-trips via a mapper-internal transport field since the file-based ore roundtrip path has no DB access to derive it from currency
- ores.cli's import_conventions writes currency_pair/currency_pair_convention instead of fx_convention
- RefdataPlugin's ORE Conventions menu now wires CurrencyPairController/CurrencyPairConventionController (previously generated but unwired) in place of FxConventionController
- Deleted all fx_convention SQL/C++/Qt/messaging files and its registrar wiring
- Fixed a prerequisite gap: currency_pair's SQL was generated but never wired into the master schema/RLS scripts
- Fixed a prerequisite gap: currency.spot_days was modelled but never regenerated (SQL/C++ only, to avoid clashing with the concurrent Commission: currency story)
- Drive-by fix: pre-existing build break in CurrencyDetailDialog.cpp (stale generator function name)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Migrate fx_convention consumers to currency_pair and retire it](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/currency-pair-refdata/task_migrate-off-fx-convention.html) | ED053F15-513A-4E3A-908A-7AE11A2B3316 |


mcraveiro:

mcraveiro:
Fixed in c3e7cdf2c — corrected the corrupted synthetic iso_code prefix (was `+"-"`, now uses an explicit generator block emitting "X", matching the "-" + counter convention used elsewhere). The pip_factor precision-drift note and missing map_fx/reverse_fx unit test are both correctly tracked as out of scope for this PR (task_test-ore-fx-convention-import.org, BACKLOG).

mcraveiro:
Fixed in c3e7cdf2c — same synthetic-iso_code fix as above, plus repointed the three dangling [[id:109FBFD6-...][fx_convention]] doc links (business_day_convention_type's model, generated domain header, and SQL create script) at currency_pair_convention's id. The map_fx/reverse_fx round-trip test gap is intentionally tracked separately (task_test-ore-fx-convention-import.org, BACKLOG), not an oversight.

mcraveiro:
Fixed in c3e7cdf2c — tick_size now defaults to 1.0 (a pip count) instead of duplicating pip_factor (an absolute rate-move value); also fixed the two dangling fx_convention doc links you flagged and the stray "USD1"/"USD=1" doc-comment artefact. Thanks for the thorough pass.

mcraveiro:
Fixed in 07a76974 — root-caused via the template this time: the generator_impl template's is_text primary-key branch had no guard for an empty generator_expr (unlike the natural_keys codepath a few lines below), so it silently emitted a stray unary-plus. Added the same #generator_expr/^generator_expr guard with a faker-noun fallback. This fixes currency_group and currency_pair_classification (the two new entities) without needing per-entity generator blocks, and the same guard will protect any future is_text primary key that omits one. Left country_generator.cpp alone — regenerating it hit an unrelated pre-existing drift between its org model and domain struct (coding_scheme_code), out of scope for this PR.

PR #1472, opened 2026-07-08
[refdata,qt,codegen] Fix currency pair flag sizing, badges, and dirty-flag bug

mcraveiro:
## Summary

Consistent flag-icon sizing and rendering across Currency/Country/Currency Pair/Currency Pair Convention list and detail views, fixes a spurious dirty-flag bug, adds cross-navigation toolbar buttons, corrects the classification badge colours away from RAG misuse, and documents the full badge catalogue.

## Changes

- Explicit shared icon size for every flag-bearing list view and combo box, replacing Qt's inconsistent implicit default
- Currency Pair Conventions list flags fixed (icon column detection missed single-field pair icons)
- Currency Pair and Currency Pair Convention detail dialogs show inline pair-code flags via new DetailDialogBase::initKeyFlagField()
- Fixed spurious dirty-flag bug from async combo repopulation
- Added Currency Pairs main-toolbar button and Conventions cross-navigation button
- Recoloured currency pair classification badges away from RAG (liquidity tier is a category, not a status)
- New doc/knowledge/ui/badge_catalogue.org badge reference, linked from ux_language.org

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Currency pair flag icons in list are too small](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/currency-pair-refdata/task_currency-pair-list-flag-icon-size.html) | DAE530F1-46FB-4901-80F3-8959EE683840 |


mcraveiro:

mcraveiro:
Fixed in 1a31babab — closed the 4 backlog tasks (D1F20726, 44F95CF2, BB347C3E, AA3A7511) this PR implements and updated the story table to DONE, each with a Result section pointing back to this PR. On the other two points: the BoolYesNoLabel unset->No (list) vs PartiallyChecked (detail) difference is intentional — a table cell wants a simple binary read, the detail dialog's checkbox needs a third visual state to distinguish 'never set' from 'explicitly false' during editing; documented in task_currency-pair-conventions-flag-badges.org's Result now. The AdminPlugin.cpp Scenario Runner lifecycle fix is genuinely in scope: I hit the reshow-blank-window bug while building the test scenario used to verify this PR's own fixes, and fixing it the right way (matching every other window's destroy-on-close lifecycle) was cheaper than working around it.

mcraveiro:
Fixed in 1a31babab — corrected the stale comment (commodity uses the dedicated classification_commodity badge, not nature_simulated) and dropped the hardcoded badge-definition count from badge_catalogue.org's summary rather than fixing the number, since it would just drift out of sync again as badges are added. On notifyOpenDialogs duplication: agreed it's identical, but per the third review pass it matches the established generated-boilerplate shape from cpp_qt_controller.cpp.mustache, so pulling it into a shared helper is a template-level change better done separately, not folded into this PR.

mcraveiro:
Fixed in 1a31babab — same stale-comment and badge-count fixes as the other review pass flagged. On the classification/business_day_convention combo defaulting to a non-empty value on create: intentional and correct given a static_combo can't represent 'unset' — dropping the classification_val.isEmpty() check in validateInput() was the right corresponding change, not an oversight.

PR #1475, opened 2026-07-08
[qt] Extract ores.qt.headless: fix Windows dllimport build failure

mcraveiro:
## Summary

Windows CI was red with MSVC C4273 (dllimport on a definition, fatal under /WX) in HierarchyModelBuilder.cpp/OrgDocRenderer.cpp. Root cause was architectural, not a missing flag: Widgets-free logic (OrgDocRenderer, TestScenarioResultsWriter, EnvironmentMetadata, HierarchyModelBuilder, FontUtils, RepoFileFinder, parse_trade_instrument/IInstrumentFormPopulator) lived inside ores.qt.api (which requires Widgets/display), forcing the test target to duplicate-compile specific .cpp files to dodge that requirement. Extracted it into a new ores.qt.headless component instead of patching the symptom.

## Changes

- New ores.qt.headless component: own export.hpp/ORES_QT_HEADLESS_API macro, distinct from ores.qt.api's ORES_QT_API to avoid a silent macro clash where both get included in the same TU
- ores.qt.api.lib links ores.qt.headless.lib PUBLIC — unchanged include paths for every existing consumer across ores.qt.admin/trading/refdata/compute/application
- ores.qt.api's tests/ directory retired — every test file in it was exactly the logic that moved; ores.qt.headless.tests is the new home, now linking the real shared library instead of duplicate-compiling source files, so this class of bug can't recur
- Verified: ores.qt.headless.lib/tests, ores.qt.api.lib, every ores.qt.* plugin, and ores.qt.exe all build and link clean on Linux; full test suite passes (198 assertions); the QA Validation Runner panel verified live end-to-end via --open-scenario after the extraction

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Fix Windows dllimport build failure via ores.qt.headless extraction](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/fix_windows_dllimport_build_failure/story.html) | 62034F29-D2EB-4BA4-9291-9B35D33A724A |
| Task | [Extract ores.qt.headless to fix the Windows build](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/fix_windows_dllimport_build_failure/task_extract_ores_qt_headless.html) | 9370492E-E752-4BBD-BEA4-15C474117C53 |


mcraveiro:

mcraveiro:
Fixed in ecbab6fe9 — renamed main.cpp's set_test_module_name and the .puml diagram node from ores.qt.api.tests to ores.qt.headless.tests. Also removed the stale FontUtils entry from ores.qt.api's own .puml/component_overview.org (moved to ores.qt.headless in the previous commit but those two docs weren't updated at the time). Noted move_trade_parse_out_of_qt.org — will follow up on closing/updating that backlog capture separately.

mcraveiro:
Fixed in ecbab6fe9 — added ORES_QT_HEADLESS_API to parse_trade_instrument()'s declaration. Good catch: exactly the bug class this PR fixes, just invisible on Linux (default GCC visibility is public) and would have been an unresolved external symbol on MSVC.

mcraveiro:
Thanks — good process note. Will trigger continuous-windows.yml via workflow_dispatch before merging to actually confirm the fix on MSVC, since the scheduled/PR checks won't exercise it.

PR #1487, opened 2026-07-09
[codegen,refdata,sql,qt] Regen country against latest codegen templates

mcraveiro:
## Summary

Country's generated files were stale relative to shared codegen template capability landed via currency's commissioning (sprint 22). Regenerates country, reconciles all drift, and generalizes one more shared-template gap found along the way.

## Changes

- Regenerated country via compass codegen entity generate; fixed a real synthetic-data-generator bug (garbage alpha2_code), picked up a previously-unreflected coding_scheme_code field
- SQL: clock_timestamp() instead of transaction-frozen current_timestamp; delete rule DO INSTEAD wrapped in parens; validate function's bootstrap pass-through now checks active rows only (fixes country's own historically-documented Item 2 bug)
- Picked up read_at_version() (temporal composite entity versioning, landed generically since country's last sync)
- Found and fixed a shared-template gap: EntityItemDelegate (icon/badge sizing) was only constructed when has_badge_columns was true, so country's own flag column bypassed the aspect-correct sizing fix landed for currency_pair. Widened the gate to a new needs_item_delegate flag; zero regen diff for currency/currency_pair confirms no regression
- Added and ran a live test scenario (6/6 steps PASSED) covering the flag-sizing fix, SQL validation fix, and generator fix

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Regen entities with latest codegen templates](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/regen_entities_latest_codegen_templates/story.html) | AF0EFC5C-899A-4AE9-BD09-7B2D4E304543 |
| Task | [Regen country against latest codegen templates](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/regen_entities_latest_codegen_templates/task_refresh_country_codegen.html) | 82BA40CA-3A4A-47B5-88ED-ADA0178E5695 |

### DIFF
--- a/doc/agile/product_backlog/inbox/gate_unordered_set_include_generator_template.org
+++ b/doc/agile/product_backlog/inbox/gate_unordered_set_include_generator_template.org
@@ -1,0 +1,45 @@
+:PROPERTIES:
+:ID: 8219B2CA-1C15-4AAF-87FC-DF27220EF462
+:END:
+#+title: Gate unordered_set include in the generator template on actual need
+#+description: cpp_domain_type_generator.cpp.mustache unconditionally includes <unordered_set>, but only entities with a hand-added bulk-unique-generator addendum (e.g. currency's generate_unique_synthetic_currencies) actually use it — every other generated *_generator.cpp carries an unused include (confirmed on country, plus 2 other already-regenerated entities). Compute a needs_unordered_set flag and gate the include.
+#+type: capture
+#+level: cross
+#+filetags: :codegen:templates:cleanup:
+#+created: 2026-07-09
+#+updated: 2026-07-09
+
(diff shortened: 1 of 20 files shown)


mcraveiro:

mcraveiro:
Addressed the review round: both passes found no blocking issues.

The one nit (unused `#include <unordered_set>` in `country_generator.cpp`) — declined fixing by hand-editing the generated file. It's emitted unconditionally by the shared `cpp_domain_type_generator.cpp.mustache` template for every entity, not something this PR introduced; 2 other already-regenerated entities carry the same unused include. Hand-patching one generated file would just drift back on the next regen. Filed as a follow-up capture instead: https://github.com/OreStudio/OreStudio/blob/feature/regen-entities-latest-codegen-templates/doc/agile/product_backlog/inbox/gate_unordered_set_include_generator_template.org

PR #1488, opened 2026-07-09
[codegen,qt] Add view_group support for Qt detail-dialog tabs

mcraveiro:
## Summary
The pre-codegen currency dialog grouped fields into separate tabs (General, Formatting, Rounding, Provenance — recovered from git history); codegen's Qt detail-dialog template only ever generates a single tab, a real regression. Adds a generic `view_group` concept to the `entity_org` schema and Qt codegen templates, verified fully backward-compatible.

## Changes
- New story with 3 substantive tasks: codegen schema/template design (done), currency pilot, manual chapter update.
- `codegen_input_org_schema.org`: optional `view_group` cell on the `Detail fields` table.
- `core.py`: `compute_view_groups()` groups the fully-enriched `detail_fields` list by `view_group`, producing `domain_entity.qt.view_groups` (groups in first-appearance order, computed widget names, per-group row index).
- `qt_detail_dialog_ui.mustache` (tangled from `ores.cpp.qt.detail_dialog_ui.org`) iterates `view_groups` to emit one tab per group instead of a single hardcoded `generalTab`.
- Backward compatibility verified mechanically (`compass codegen entity generate <entity> --diff --address ores.cpp.qt` shows zero diff for `purpose_type` and `currency`), not just asserted.
- 8 new unit tests (`test_view_groups.py`); full codegen suite (38 tests) passes. Also fixed a missing `pytest` entry in `requirements.txt`.

## Test plan
- [x] `compute_view_groups()` unit tests pass
- [x] Zero-diff regeneration confirmed for `purpose_type` and `currency`
- [x] Multi-tab path hand-verified against a modified `counterparty` model copy (well-formed XML, correct names/ordering)
- [ ] Currency pilot task (next) will exercise this live in the running app


mcraveiro:

mcraveiro:

mcraveiro:
Fixed in 12089e805 and 7024fcd86 — (1) compute_view_groups() now keys the groups dict by the same normalized form used to derive widget names (keeping the first-seen raw string as the display title), so case/whitespace-only view_group variants fold into one group instead of colliding; added a regression test (test_view_groups_differing_only_by_case_or_whitespace_fold_together). (2) Removed the dead has_multiple_view_groups field. (3) Moved the story's sprint.org row from Hotfixes to Epic: Codegen. (4) Filled in the scaffold task's Goal/Acceptance. (5) Added :codegen:/:qt: tags to the implement task. (6) Filled in the scenario's pending Commit field (ed36424c3). pytest projects/ores.codegen/tests/test_view_groups.py: 9/9 pass; compass codegen entity generate currency --diff --address ores.cpp.qt shows no view_group-related diff (currency's own values were already consistently cased, so its output is unchanged).

PR #1489, opened 2026-07-09
[refdata,qt] FX convention test, dup/inverted pair reject, eventing

mcraveiro:
## Summary

Bundle of currency-pair-refdata mop-up tasks: a real ORE conventions.xml round-trip test through currency_pair/currency_pair_convention (not just raw XML), duplicate and inverted pair_code rejection on create, and the root cause of the repeated 'no eventing' finding -- currency_pair/currency_pair_convention were never wired into the NATS event registrar. Also gives currency_pair_convention a proper flagged combo for pair_code, matching currency_pair's own base/quote UX.

## Changes

- New xml_fx_convention_mapper_roundtrip_tests.cpp: round-trips conventions_mapper::map_fx/reverse against the real ORE example file, not just raw XSD load/save; documents two real findings (Id not preserved, a genuine duplicate-pair collision in the source file)
- Reject exact-duplicate and inverted-duplicate (USD/EUR when EUR/USD exists) pair_code on Currency Pair create, client-side, with clear explanatory errors
- Fix: currency_pair and currency_pair_convention were never wired into event_registrar::register_event_mappings(), so saves never published NATS change events -- generated the missing registrars and wired them in
- currency_pair_convention's pair_code is now a flagged combo (dual-flag pair icon), selectable on create, visually-normal-but-locked on edit; new FlagSource::CurrencyPair + fetch_currency_pair_codes() generalize the existing flag/combo machinery
- Duplicate-convention-for-a-pair rejected on create, same pattern as currency_pair's own check
- New knowledge doc: Optimistic Concurrency Versioning -- the version=0 overload conflated across 4 use cases, root cause of why duplicate rejection can't live in the shared SQL trigger; filed as its own capture for the full DB-level fix
- 2 new test scenarios (duplicate/inverted reject, eventing live-refresh), a capture for grouping currency combos by classification, and devops-run-client skill fix (connect step must be a tracked * Steps entry, not just * Context prose)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Test ORE import/export of FX conventions against currency_pair](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/currency-pair-refdata/task_test-ore-fx-convention-import.html) | 2ABA2FD7-6381-481E-B236-FF401D4EFEB2 |


mcraveiro:
Fixed in a4c0ef5ce — extended the is_key branch to wire combo widget types to onFieldChanged (mirrors the ^is_key combo handling), gated onCodeChanged's generation on a new has_line_edit_key flag since it's now dead code for combo-widget keys, and hoisted the 100000 magic number to the existing lookup_fetch_limit constant (turned out to already be sufficient/shared, no new constant needed).

mcraveiro:
Fixed in a4c0ef5ce — both pre-checks now fail closed (block save with a warning) if the request itself fails, instead of silently falling through to save. Also hoisted the duplicated 100000 to the existing lookup_fetch_limit constant (1000 turned out to already be a safe shared limit, no new constant needed). Trailing blank-line item: turns out pre-existing, unrelated to this PR -- verified an already-merged, untouched entity (BookDetailDialog.cpp) has the identical double-blank-line pattern at the same template seam when it has no extra implementation content there, so it's a general codegen artifact, not something this PR introduced. Left as-is rather than risk breaking the populated case (verified my first attempt at a template fix did exactly that).

PR #1492, opened 2026-07-10
[marketdata] Detect and correct reversed FX spot quotes during import

mcraveiro:
## Summary

Wires fx_quote_convention_checker into ores.marketdata's import_service so a reversed FX/RATE key (e.g. Legacy/Example_56's GBP/USD, filed under FX/RATE/USD/GBP) is auto-corrected against ores.refdata's currency_pair reference data at import time, instead of silently mis-keying market data.

## Changes

- Add fx_quote_convention_checker (ores.ore.core): pure, dependency-free, 8 unit tests — detects a reversed FX pair against a known-pairs set and swaps the qualifier only, never the value.
- Wire it into import_service via ores.marketdata.service's first outbound service-to-service call: a client-credentials nats_client plus per-request On-Behalf-Of delegation of the importing caller's token, since currency_pair reference data is tenant/party-scoped.
- Remove the one-off SQL special-case that hardcoded GBP/USD's ore_key to the reversed key — no longer needed now that import corrects it; synthetic.gbpusd's ore_key now naturally resolves to FX/RATE/GBP/USD.
- Document the client-credentials vs. on-behalf-of pattern properly: new service_to_service_auth_patterns.org, cross-linked from Anatomy of a Service and market_data_architecture.org, including the anti-pattern (direct cross-service repository access) explicitly called out.
- Verified end-to-end against a live stack: correction fires on Legacy/Example_56's market.txt, resulting series lands on canonical GBP/USD, and a started synthetic feed's first tick matches the real imported vintage spot.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Synthetic data collections: Basic and Realistic](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/synthetic-data-collections/story.html) | D1234E9D-D79A-4C59-991C-988D1D8F515F |
| Task | [Detect and correct reversed FX spot quotes during market data import](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/synthetic-data-collections/task_fx_quote_inversion_detection.html) | F3CB17FB-6472-4DEB-BC5B-EC764C1368F8 |


mcraveiro:
Fixed in 0e661d505 — added a std::any_of short-circuit over the parsed batch before the fetch/checker construction, so non-FX imports (equity-only, fixings-only) no longer pay for the refdata round-trip. Also fixed the include-sort nit in the same commit.

mcraveiro:
Re: test coverage note — agreed, not fixing in this PR. The swap-on-match path is covered by fx_quote_convention_checker's own 8 unit tests plus this PR's manual live-stack verification; adding a fake in-process NATS server just to fabricate a currency_pairs.list response for this one PR isn't worth the new test infrastructure. Leaving as a known gap.

PR #1498, opened 2026-07-10
[synthetic] Seed Basic and Realistic FX spot config bundles

mcraveiro:
## Summary

Adds two new opt-in dataset bundles (synthetic_basic, synthetic_realistic) seeding all 8 major FX driver pairs from the curated 2016-02-05 vintage, differing only in process/GMM calibration. Makes the publish-from-dq function data-driven instead of hardcoding vintage/GMM values, fixes a real orphaned-component bug found during manual testing, and fixes a client-side bug where market observations were fetched unfiltered and silently truncated.

## Changes

- Add synthetic_basic and synthetic_realistic dataset bundles: all 8 major FX driver pairs, Basic uses a single exaggerated-volatility geometric component, Realistic uses a 2-component GBM Gaussian mixture calibrated to plausible 2016 realised FX volatility
- Make ores_synthetic_publish_fx_spot_configs_from_dq_fn data-driven: per-row price_source/vintage_source/vintage_date and a new per-pair GMM component artefact table, instead of hardcoded PL/pgSQL constants; backfill ore_analytics so its behaviour is unchanged
- Fix orphaned GMM component bug: switching a party between bundles left a prior dataset's extra components (e.g. Realistic's tail component) still valid when the new dataset has fewer components; now voided unconditionally, not just under replace_all
- Fix ClientMarketObservationModel to filter observations server-side by series_id instead of fetching an unfiltered page and filtering client-side, which silently truncated a series' rows once continuously-ticking synthetic feeds grew the table
- Capture a separate real bug found during testing: the FX Spot chart window doesn't refresh when feed bindings change while it's already open

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Synthetic data collections: Basic and Realistic](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/synthetic-data-collections/story.html) | D1234E9D-D79A-4C59-991C-988D1D8F515F |
| Task | [Seed Basic and Realistic dataset bundles](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/synthetic-data-collections/task_basic-realistic-bundles.html) | 73E7622C-CBD2-48C8-96B9-DEA8AA7C45C6 |


mcraveiro:

mcraveiro:
Thanks for the review. Findings:

1. **ClientMarketFixingModel has the same truncation bug** — confirmed, real. Captured as a separate backlog item (`fixings_model_unfiltered_page_truncation`) rather than folded into this PR, since it needs a protocol change (new `series_id` field on `get_market_fixings_request`) not just a client tweak — different scope than this PR's fix.
2. **`total_available_count` inconsistency** — acknowledged, pre-existing and not user-visible today (unused by `MarketObservationMdiWindow`). Left as-is; noted in the capture above as a related note for whoever picks up the fixings fix.
3. **No unique constraint on the GMM components artefact table** — acknowledged as defense-in-depth; not adding now since nothing in the current populate scripts can produce a duplicate (all guarded by the parent-table `exists` check) and the downstream unique index would catch it loudly if this ever changed.

No code changes needed in this PR — all three findings are either out-of-scope fast-follows (now tracked) or genuinely non-blocking. All "what checked out fine" items match the manual end-to-end verification already recorded in the task doc.

mcraveiro:
Missed two findings from earlier rounds in my last reply — fixed both in 98ebe0f40:

1. **Codegen literate source drift (round 1)** — added the missing `price_source`/`vintage_source`/`vintage_date` `** column` sections to `ores.dq.synthetic_fx_spot_config_lookup_entity.org`. Couldn't run the actual regeneration tool in this environment (the artefact-table template is invoked via a non-standard "dataset path", not the usual `codegen entity generate`), so verified by hand against `sql_schema_artefact_create.mustache`'s column-rendering logic line-by-line — including catching that `price_source`'s default needed to be `'vintage'` (quoted) in the `:default:` property, not bare `vintage`, to reproduce the hand-written SQL's `default 'vintage'` exactly.

2. **Fallback-branch orphan voiding (round 2)** — the `else` branch (single-placeholder-component fallback, used when a dataset has no GMM artefact rows) now voids orphaned higher-index components too, mirroring the fix already applied to the primary branch. Verified live against a real DB: published a party onto `synthetic_realistic` (2 components for EUR/USD), then onto a throwaway test dataset that exercises the fallback branch, confirmed the orphaned `component_index=1` row was correctly voided (only `index=0`, the placeholder, remained).

PR #1506, opened 2026-07-11
[agile] Deep backlog refinement: audit 559 items, discard 71

mcraveiro:
## Summary

First real run of the `agile-refine-backlog` skill's new deep refinement mode. Every item in `inbox` (240), `next` (258), and `deferred` (61) — 559 total — was verified against the live codebase via a parallel-agent workflow, duplicates were clustered and adjudicated, and the results applied.

## Changes

- 68 captures discarded (moved to `product_backlog/discarded/` with a bolded resolution note citing evidence) as already resolved elsewhere in the codebase
- 3 stories misfiled in the backlog buckets (despite already being DONE/ABANDONED and correctly recorded in their real sprint's table) restored to their actual sprint folder, fixing stale State fields along the way
- 1 duplicate pair merged, folding unique detail into the survivor before discarding the absorbed capture
- 3 "resolved" verdicts caught as misclassified — 2 by this run's own trustworthiness check, 1 by the PR review that followed — and corrected rather than silently discarded or left wrong
- Full per-item evidence and process notes in the linked task's Result section

Net: 559 → 488 live backlog items (13% smaller). Next/deferred bucket rescoring against sprint 23's mission is flagged as a follow-on run.

## Verification

Docs-only change (agile backlog org files) — no C++ source touched, no build/test applicable.


mcraveiro:
Fixed in 9377049c4 — accepted. The discard evidence was accurate but incomplete: make_service_token_provider is wired in 5 services, but a full sweep of every application.cpp found 13 more (analytics, trading, http.server, iam.service, workspace, dq, variability, scheduler, controller, compute.service, compute.wrapper, refdata, assets, telemetry) already making outbound NATS calls without the wiring. This is the same 'partial ask shipped, not all of it' failure mode the run's own trustworthiness check caught twice elsewhere -- restored to next/ with the corrected 13-service scope rather than re-discarded. Task's Result section updated with the correction and revised counts (68→ correction noted, net 71→ still 70 discards + 1 restored, see task doc for exact reconciliation).

mcraveiro:
Fixed in 6f80010f6 — accepted. The description said '6 of 19 wired; 13 more' but the enumerated list has 5 wired + 14 missing = 19. Corrected in the capture's description, next.org, and the task's Result section.

PR #1510, opened 2026-07-12
[analytics.quant] Add CRM driver/derived spanning-tree topology builder

mcraveiro:
## Summary

New dependency-light component ores.analytics.quant (only Boost::graph + immer, no database/service/NATS/refdata coupling), consumable standalone as a library. topology_builder validates a set of currency pairs into an immutable crm_topology via incremental union-find cycle detection during construction: any input admitting more than one path between two currencies is rejected at build time with readable, aggregated errors -- never silently resolved (a deliberate departure from QuantLib::ExchangeRateManager's own documented undefined behaviour in that case).

## Changes

- New component ores.analytics.quant: domain value types (currency_id, ccy_pair, ccy_pair_input, topology_error, topology_build_error, crm_topology) + service::topology_builder
- topology_builder uses boost::disjoint_sets for incremental cycle detection and a BFS from the pivot for parent/path assignment; missing-major and disconnected-currency checks; errors collected across the whole input and thrown together
- Added immer to vcpkg.json (for the deferred rate-engine task in this story)
- Component overview + class/build-sequence/runtime-sequence PlantUML diagrams under modeling/ and the task's design/ folder
- Verification: local build clean (linux-clang-debug-make); ctest ores.analytics.quant.tests 1/1 passed (6 test cases, 20 assertions)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Model the driver/derived spanning-tree topology](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/crm_implementation/task_model_driver_derived_topology.html) | E7E31384-3616-4DFB-AB20-6CB099114FB3 |


mcraveiro:

mcraveiro:
Fixed in 847571b53: ccy_pair now carries is_driver, threaded through topology_builder end to end (a test asserts it survives path_to_pivot). Excluded required majors from the general disconnected_currency sweep so an unreachable major is reported exactly once, not twice — strengthened the missing-major test to assert the error count. Removed the dead pivot-not-found branch (pivot_code is always resolved unconditionally before that lookup, so it could never fire) and replaced it with a direct .at() plus a comment explaining why. Left the self-loop UX nit as-is: cycle_conflict is still an accurate, if generic, diagnosis for base==quote, and a dedicated error kind felt like scope creep for this task.

mcraveiro:
Same fix as the other review pass, in 847571b53: majors excluded from the general disconnected_currency sweep (exactly-one-error test added), and the dead pivot-not-found branch removed since pivot_code is always resolved before that lookup runs.

mcraveiro:
Fixed in 847571b53: is_driver now flows through into ccy_pair/crm_topology (domain/crm_topology.hpp's doc comment is now accurate), with a new test covering the driver/derived direction on a mixed-driver topology. Dead pivot branch removed as suggested. Self-loop nit left as a generic cycle_conflict — declined for this task's scope, see the round-1 Review table on the task doc.

PR #1512, opened 2026-07-12
[analytics.quant] Add CRM runtime rate engine (triangulation/derivation)

mcraveiro:
## Summary

Runtime rate_engine on top of the crm_topology from PR #1510: consumes a stream of driver_quote ticks (single producer, possibly cross-thread) and serves batched derived-rate reads via an immer::atom<rate_snapshot> -- no locks on the hot path. Rates are a per-vertex cumulative log-rate from the pivot, so a derived rate is one subtraction and update() only recomputes the subtree below the changed edge. Staleness is a per-vertex timestamp propagated alongside the log-rate, so a stale major automatically taints every derived rate depending on it.

## Changes

- New: driver_quote, rate_status, staleness_policy, derived_rate, vertex_state, rate_snapshot domain types
- New: service::rate_engine (update/rate/rates), crm_topology gained parent()/edge_to_parent() accessors
- 8 new test cases (15 total, 43k+ assertions): triangulation, staleness propagation, subtree-only recompute, batched reads, rejection of non-edge pairs, and a concurrent-update-vs-concurrent-read stress test
- Ran clang-format over the whole component (picked up formatting drift from PR #1510)
- Verification: local build clean (linux-clang-debug-make); ctest ores.analytics.quant.tests 1/1 passed (15 test cases, 43k+ assertions, stable across repeated runs)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Implement triangulation/derivation](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/crm_implementation/task_implement_triangulation_derivation.html) | 96522E06-2067-4058-86FC-27B6086DCC26 |

### DIFF
--- a/doc/agile/versions/v0/sprint_23/crm_implementation/story.org
+++ b/doc/agile/versions/v0/sprint_23/crm_implementation/story.org
@@ -28,10 +28,10 @@ engine the architecture relies on.
 |---------------+----------------------------------------|
 | State         | STARTED                              |
 | Parent sprint | [[id:341C071C-3042-4386-BC14-0D52CE8E6250][Sprint 23]] |
-| Now           | Modelling the driver/derived spanning-tree topology. |
+| Now           | Implementing the runtime rate_engine (triangulation/derivation). |
 | Waiting on    | Nothing.                               |
-| Next          | Implement triangulation/derivation once the topology is in place. |
-| Last touched  | 2026-07-11                               |
+| Next          | Risk recentering, then wiring into ores.marketdata's ingest path. |
+| Last touched  | 2026-07-12                               |
 
 * Acceptance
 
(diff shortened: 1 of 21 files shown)


mcraveiro:

mcraveiro:
Fixed in 0fd992d32: added an explicit base_id == quote_id rejection before the parent-equality checks (the self-pair-on-pivot case), and a std::isfinite() && > 0.0 check on the rate, both throwing std::invalid_argument consistent with the existing checks. Also removed the unused edge_to_parent() accessor and corrected ccy_pair's stale doc comment about is_driver. New tests cover both rejections plus the pivot-state-not-mutated assertion.

mcraveiro:
Fixed in 0fd992d32: added the base_id == quote_id guard exactly as suggested (before the parent() comparisons), plus a finite/positive check on the rate. Also fixed the weak test (split unknown-currency vs known-but-unconnected into separate cases, the latter using EUR/JPY which both hang off USD but aren't connected to each other), removed the unused edge_to_parent() accessor, and corrected the is_driver doc comment on ccy_pair to reflect that direction actually comes from tree structure.

mcraveiro:
Fixed in 0fd992d32: base_id == quote_id is now rejected up front (covers the pivot self-pair case), and quote.rate is validated finite/positive before use. Also removed the unused edge_to_parent() accessor and corrected the ccy_pair doc comment about is_driver per your minor note -- good catch, the engine really does derive direction purely from tree structure.

PR #1514, opened 2026-07-12
[ores.iam,ores.qt] Add default party per account with quick-login checkbox

mcraveiro:
## Summary

Lets an account carry a default party so login can skip the party-selection step via a "Log in to default party" checkbox on the login dialog, and updates provisioning scripts to set it.

## Changes

- Add default_party_id (soft FK, no hard constraint) to domain::account, account_entity, account_mapper, and the accounts SQL table
- Add self-service set_my_default_party_request/handler, used by a new 'accounts set-default-party' shell command and wired into both the Barclays and Acme provisioning recipes
- Add login_response.default_party_id (multi-party branch only) and a quick-login checkbox in LoginDialog that calls ClientManager::selectParty() directly when ticked and a default is available, falling back to PartyPickerDialog otherwise
- Add admin-side editing as a regular 'Default Party' combo box in AccountPartiesWidget's Parties tab (sorted alphabetically, below the assignment list), going through the existing Save/change-reason/update_account_request/version-history path like every other account field
- Fix a pre-existing tenant-isolation bug in account_party_handler's list/by_account NATS handlers (used the handler's system-scoped context instead of the caller's JWT-derived tenant context, silently returning 0 rows under RLS for non-system-tenant accounts)
- Verification: local build clean (linux-clang-debug-make, full project); ctest 68/69 passed — the 1 failure (ores.diff.tests) is pre-existing and unrelated, from commit caadbcd9b already on main before this branch; this branch never touches projects/ores.diff

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Add a default party per account with a quick-login checkbox](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/account-default-party-login/story.html) | 7A92D9F1-2E2B-4EBE-BFC8-DD5AD9BA690B |
| Task | [Update provisioning scripts to set the account's default party](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/account-default-party-login/task_provisioning-scripts-set-default-party.html) | 18F72AD6-A5A1-4D7A-902F-1213075E0E4C |


mcraveiro:

mcraveiro:
Fixed in b1ed4f28e — added a default member initializer (boost::uuids::nil_uuid()) to AccountPartiesWidget::defaultPartyId_. Also added service-layer test coverage for set_my_default_party (success/idempotent/nonexistent-account) and update_account's default_party_id set/clear; declined a handler-level tenant-isolation regression test for this round as impractical without an existing NATS-handler test harness to extend in this file — noted as a follow-up gap. Also made set_my_default_party idempotent (re-setting the same party now succeeds silently instead of erroring).

mcraveiro:
Fixed in b1ed4f28e — AccountController::onRevertAccount now threads the historical version's default_party_id into the update_account_request, the same way email is already handled there. Good catch, thanks.

mcraveiro:
Fixed in bdf8570c2 — good catch, this was a real gap in my round-1 fix. hasPendingDefaultPartyChange() now returns false until the combo has been seeded from the account's actual stored default (new isDefaultPartyReady()); Save stays disabled until parties data has actually loaded (gated in updateSaveResetButtonState, re-evaluated via the dataLoaded signal); added a defensive re-check in onSaveClicked() against any other trigger path. Verified: rebuilt ores.qt.admin.lib clean, ran ctest -R iam (100% pass).

mcraveiro:
Fixed in c1103f71b — good catch, this was a real regression from the round-2 fix. Added a separate rebaseDefaultParty() that only updates the pending-change baseline (defaultPartyId_) without touching defaultPartyComboInitialized_; the post-save success handler now calls that instead of setDefaultPartyId(), which stays reserved for the initial per-account seeding path in setAccount(). Verified: rebuilt ores.qt.admin.lib clean, ran ctest -R iam (100% pass).

mcraveiro:
Fixed in d95d4d1a0 — good catch, thanks for tracing through the reload path. Made the setCurrentIndex re-seed in populateDefaultPartyCombo() unconditional on every call (not just first populate), so it can't desync from defaultPartyId_ regardless of which reload path triggered it. defaultPartyComboInitialized_ is now used only for gating isDefaultPartyReady()/hasPendingDefaultPartyChange(). Verified: rebuilt ores.qt.admin.lib clean, ran ctest -R iam (100% pass).

PR #1524, opened 2026-07-12
[analytics.quant] Fix intermittent SIGSEGV/SIGABRT in rate_engine concurrency test

mcraveiro:
## Summary

Root cause was a test bug, not a production bug: the concurrency test's 4 reader threads called Catch2's REQUIRE() directly from worker threads. Catch2's assertion macros are documented upstream as not thread-safe -- concurrent REQUIRE/CHECK from multiple threads is itself a data race, independent of anything rate_engine/crm_topology/immer::atom does. This explains every symptom in the original CI failures: reproduces only with the xml reporter, not in isolation, and the SIGABRT is Catch2's own OutputRedirect re-entrancy assertion firing. This is currently breaking CI on main and any branch that builds ores.analytics.quant.tests -- flagging as urgent.

## Changes

- Confirmed root cause via a standalone ThreadSanitizer build (clang++ -fsanitize=thread; TSAN not wired into any CMake preset) that found zero races in the library code itself under the exact crashing conditions -- redirected the investigation to the non-instrumented vcpkg-prebuilt libCatch2.a
- Fix: reader threads now collect failures into a plain std::atomic<int>; only the main thread asserts, once, after every thread has joined. No production code (rate_engine.cpp, crm_topology.hpp, rate_snapshot.*) changed
- Verification: local build clean (linux-clang-debug-make); 15/15 clean runs of the full test binary with the xml reporter (the exact condition that reproduced the crash 5/5 before the fix), 10/10 clean ctest -R runs, all 23 test cases in the component green

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Fix rate_engine concurrent update/rates crash (SIGSEGV/SIGABRT)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/crm_implementation/task_fix-rate-engine-concurrency-crash.html) | 5EAA2830-832A-4CD0-A60D-67F1033996A4 |


mcraveiro:

PR #1526, opened 2026-07-12
[refdata,sql,codegen] Migrate Portfolio/Book to composite-entity versioning

mcraveiro:
## Summary

Registers Book as a composite child of Portfolio (not book_status/regulatory_book_type, which are enum-like FK attributes) so add/edit/delete on a book bumps its parent portfolio's version, with recursive bump propagation up a nested portfolio tree. Fixes a real codegen-template bug (touch-function generation order vs Postgres's eager CREATE RULE resolution) uncovered by Portfolio's self-reference.

## Changes

- book.org: parent_portfolio_id FK gains :bump_parent_version: true, :list_by_as_of: true
- portfolio.org: :generate_touch_function: true on Flags; new self-referencing parent_portfolio_id FK with :bump_parent_version: true for recursive multi-level bump
- Fixed shared SQL template (org source + tangled mustache): touch-function now generated before insert trigger/delete rule, fixing self-referencing composite entities; confirmed no regression to the existing party/counterparty pilot
- Verified directly via SQL (not just generated code) at 2 and 3 levels deep: insert/delete bump correctly, and portfolio-as-of-version correctly composes contained books
- Manual QA scenario PASSED; the one reported delete-bump FAIL was root-caused via SQL cross-check against the real portfolio/tenant to the known Portfolio-window live-refresh gap, not a backend defect
- Verification: ores.refdata.api/core/service.tests 3/3 passed (linux-clang-debug-make)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Book codegen drift remediation](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/book_codegen_drift_remediation/story.html) | F0DAFC8D-3C94-4F97-B546-D7AF7469A29E |
| Task | [Migrate Portfolio/Book to composite-entity versioning](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/book_codegen_drift_remediation/task_migrate_book_to_composite_entity_versioning.html) | 01304B6F-C431-4BD3-885A-CE713114B5A4 |


mcraveiro:

mcraveiro:
Fixed in bd66726ce — added a blank line between the {{/list_by}} and {{#list_by_as_of}} sections in the 4 shared .org templates (repository header/impl, service header/impl), re-tangled, and regenerated book/book_status/regulatory_book_type. Also collapsed a pre-existing duplicate blank line before remove() at the same boundary that the fix surfaced.

mcraveiro:
Fixed in bd66726ce — addressed all 3 findings:
1. Cycle prevention: added a generic :prevent_cycle: true FK flag to the shared SQL template. A recursive-CTE ancestor walk in the insert trigger now rejects a parent_portfolio_id that would create a cycle (direct or transitive), raising a clean exception instead of hitting Postgres's stack-depth error via the touch-version recursion. Applied to portfolio.org's self-referencing FK.
2. Regenerated refdata_counterparties_create.sql (--address ores.sql.schema) so its touch-function ordering matches the current template — diff confirmed content unchanged, only relocated, as previously analysed.
3. Missing blank lines: same root cause/fix as the other review's blank-line finding — fixed in the shared templates and regenerated.

ores.refdata.api/core/service.tests 3/3 passed after the changes.

mcraveiro:
Fixed in bd66726ce and 0d45e2e78:
1. Silent decode failure in book_handler.hpp — added the missing else branch (error_reply/bad_request), matching the rest of the file and party_identifier_handler.hpp.
2. clang-format — it's on PATH here and the pipeline does invoke it; re-checked with clang-format --dry-run --Werror after the round-1 template fix (blank-line gap noted by the other review pass) — all four files are already clean, so this was resolved as a side effect of that fix, not a separate gap.
3. Architecture doc — updated "Multi-level composition" and the summary bullet to describe the proven Portfolio/Book self-referencing case and the new :prevent_cycle: flag.
4. Unreachable list_by_as_of methods — matches the existing pattern already used by counterparty_identifier/counterparty_contact_information (repository+service only, no protocol template for it yet); consistent with prior art, not changed.
5. Cycle protection — this review pass ran against a stale snapshot: :prevent_cycle: true (recursive-CTE ancestor check in the insert trigger) was already added in bd66726ce, addressing the equivalent finding from the other review pass. Confirmed still present.

ores.refdata.api/core/service.tests 3/3 passed after all changes.

mcraveiro:
Fixed in 8c23537e0:
1. Double blank line — normalized the 4 generated files (cat -s post-generation); root cause in the mustache renderer isn't fully understood (the phantom blank appears regardless of how the closing tags are laid out), so filed a backlog capture (codegen_double_blank_line_list_by_as_of.org) rather than continuing to chase it for a Low-severity cosmetic issue. Confirmed clang-format --dry-run --Werror clean.
2. Silent decode failure in book_handler.hpp — correction: this method IS generated (extra_list_requests section of cpp_nats_handler.hpp.mustache), not hand-written as I said in my first reply. My earlier hand-patch would have been silently reverted by the next regen. Fixed properly at the template source this time, matching the sibling hierarchy handler's else branch, and regenerated.
3. Architecture doc — already fixed in 0d45e2e78 (pushed before this review pass started); this pass ran against a slightly earlier snapshot.

ores.refdata.api/core/service.tests 3/3 passed after all changes.

mcraveiro:
Fixed in c100d717a — regenerated party_identifier_handler.hpp and counterparty_identifier_handler.hpp (--address ores.cpp.nats-handler, scoped to just this facet to avoid pulling in their unrelated Qt-layer staleness). Diff confirmed both picked up exactly the same else-branch fix as book_handler.hpp, nothing else.

ores.refdata.api/core/service.tests 3/3 passed after the change.

PR #1530, opened 2026-07-12
[marketdata] Wire the CRM engine into the ingest path

mcraveiro:
## Summary

crm_ingest_bridge owns one ores.analytics.quant::rate_engine per (tenant, party), built from crm_topology_config + enabled crm_driver_pair rows. Config-change events trigger a full refresh() that atomically swaps in a whole new engine set, so the hot path (update()/rate()/rates()) never races with a refresh. feed_ingest_loop gained an optional hook: every FX/RATE tick is offered to the bridge as a candidate driver update, silently ignored if not CRM-configured. Two new pull-only NATS endpoints: marketdata.v1.crm.rate (single pair) and marketdata.v1.crm.rates (every configured pair for a party in one batch call) -- both, per the story's architecture decision, never broadcast the derived set as ticks.

## Changes

- New: crm_ingest_bridge (service/app), crm_handler + crm_protocol (the two NATS endpoints), read_latest_all_tenants added to all three CRM entities (codegen)
- feed_ingest_loop hook: offers every FX/RATE tick to the bridge as a candidate driver update, never fails the tick's normal persist+remap if the CRM lookup fails
- application.cpp: constructs the bridge, initial refresh(), subscribes to crm_topology_config_changed_event/crm_driver_pair_changed_event, registers the two new endpoints
- Filed a second codegen-gap capture for the two SQL fixes carried forward from the entities task's review (cross-party FK validation, single-enabled-config uniqueness); crm_ingest_bridge carries an application-level safeguard (first enabled config wins, loudly) until the codegen facility exists
- 5 new integration tests (app_crm_ingest_bridge_tests.cpp, DB-backed): end-to-end refresh+update+triangulation, nullopt-vs-unavailable semantics, non-edge-pair no-op, batched rates(), ambiguous-config safety
- Filed a follow-up UI task (CRM cross-rates matrix panel, manual reload only) referencing a mockup
- Verification: full local build (all targets) clean; full test suite 71/71 suites, 0 failures; db recreate clean; validate_schemas.sh --strict zero warnings

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Wire the CRM into ores.marketdata's ingest path (service wiring)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/crm_implementation/task_wire_crm_service_layer.html) | 88FD9001-F2BA-4061-994C-5EBE8BBD590A |


mcraveiro:

mcraveiro:
Acknowledged both observations — no change needed. (1) update()'s doc will be left as-is; the non-finite/non-positive-rate case is caught by the same std::invalid_argument handler as the other two cases, so behavior is correct, just not explicitly enumerated in the comment — not worth churning the doc for. (2) marketdata::crm:read is intentionally the first read-gated CRM permission; IAM role grants for consumers of these endpoints are tracked as a rollout/ops task, not a code follow-up.

PR #1538, opened 2026-07-13
[marketdata] Support multiple concurrently-enabled named CRM configs per party

mcraveiro:
## Summary

crm_ingest_bridge now builds and holds one rate_engine per enabled crm_topology_config per party (named), not just one first-enabled-wins engine -- matching real FX desk tiering (majors/exotics). Also fixes the macOS CI build in the same touched type.

## Changes

- crm_ingest_bridge: engines map keyed by (tenant,party) -> vector of named engines; every enabled config builds its own engine
- update() feeds a tick to every named engine that has it as an edge
- rate()/rates() gain a required/optional crm_name param; new all-CRMs rates() overload returns named_rate-tagged results
- get_crm_rate_request/get_crm_rates_request/crm_rate_item carry crm_name through the NATS protocol
- Fix macOS build: std::atomic<shared_ptr<T>> -> mutex-guarded shared_ptr (portable, same semantics)
- Verification: local build clean (linux-clang-debug-make); ctest 71/71 passed; db recreate -y -k clean; validate_schemas.sh --strict clean

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Support multiple concurrently-enabled named CRM configs per party](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/crm_implementation/task_crm_support_multiple_named_configs_per_party.html) | BA82072F-0B2B-48F4-9C22-8D6D39337099 |


mcraveiro:

mcraveiro:
Fixed in df7498317 — ran clang-format on the touched files to clear the drift. No functional concerns raised.

mcraveiro:
Fixed in df7498317, but differently than proposed for finding #1 (phantom map entry) — switched refresh()'s lookup to find() and only insert via operator[] on the successful push_back path, so a config that fails after the lookup leaves no entry behind. Finding #2 (duplicate-name test coverage): investigated further — crm_topology_config's (party_id, name) natural key already has a codegened partial unique index (crm_topology_configs_party_id_name_uniq_idx) scoped to active rows, so two enabled configs sharing a name for one party is structurally impossible via the normal write path; the app-level warn-and-skip was dead code. Replaced it with a hard throw documenting that invariant (isolated per-config via the existing catch, so a violation can't crash the whole service), and added a test asserting the DB itself rejects the duplicate write — testing the real enforcement point rather than unreachable app logic. Finding #3 (crm_name asymmetry) is intentional, per the doc comment; confirmed as-is.

mcraveiro:
Fixed #2 in df7498317 — added an explicit empty-crm_name check in crm_handler::rate() replying "crm_name is required" before the bridge lookup, distinguishing it from the unknown-CRM-name case. Findings #1, #3, #4, #5 confirmed correct as designed, no change needed.

PR #1555, opened 2026-07-13
[analytics.quant] Add Vasicek/CIR/Hull-White short-rate processes

mcraveiro:
## Summary

Add IYieldCurveProcess (IStochasticProcess + discount_factor()) and three mean-reverting short-rate engines implementing it: hull_white_process, vasicek_process (composed on top of hull_white_process), and cir_process. Pure closed-form math, no QuantLib, mirroring ou_process's existing shape.

## Changes

- New IYieldCurveProcess interface extending IStochasticProcess with discount_factor(ticks_ahead) -- the model-implied zero-coupon bond price from the current state, so every instrument derived from one process is guaranteed to be a slice of one internally consistent curve
- hull_white_process: general 1-factor Gaussian short-rate engine with a piecewise-constant theta_path; written in target-level form so kappa <= 0 and constant-theta (Vasicek) reduce to exactly ou_process's formula, not an approximation; discount_factor() via a backward recursion mutually consistent with next()'s transition law
- vasicek_process: literal composition of hull_white_process with a single-element theta_path, not reimplementation; tests assert bit-identical paths and discount_factor() output against the equivalent hull_white_process directly
- cir_process: independent square-root-diffusion engine; exact simulation via the Poisson-mixture-of-chi-squared construction (Glasserman 2003), with a dedicated sigma == 0 deterministic-ODE branch (the general formulas have a sigma-denominator singularity there, not just an optimisation)
- New validate_yield_curve_process_parameters, kept separate from the existing validate_process_parameters rather than overloaded onto its incompatible means/stdevs/weights array shape
- process_factory::make_yield_curve_process() selects vasicek/cir/hull_white by name and throws on an unrecognised type (unlike make_process()'s geometric fallback)
- Comprehensive Catch2 suites per class: deterministic-seed reproducibility, edge cases, statistical mean-reversion checks, closed-form cross-checks against independently-recomputed oracle formulas
- Pure math-library addition with no consumer yet -- the Curve Template config task (ir_curve_generation_config, still BACKLOG) will wire make_yield_curve_process() into an actual caller
- Verification: local build/ctest in progress concurrently at PR-raise time; will report the result as a follow-up comment

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [IR Rates synthetic data generation](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/ir-rates-synthetic-generation/story.html) | 3ECC4BA8-6ACA-4028-9E42-AE7F25FC3B98 |
| Task | [Short-rate stochastic process (Hull-White/CIR/Vasicek)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/ir-rates-synthetic-generation/task_short-rate-process.html) | 01143977-5DA2-48AE-B4BE-109585BC4962 |


mcraveiro:
Local verification follow-up:

- Build: clean (linux-clang-debug-make)
- `ores.analytics.quant.tests` (the suite covering all new Hull-White/CIR/Vasicek code in this PR): **100% passed**
- 51/73 other test suites failed in this local run, all due to the same pre-existing environment issue — Postgres connection-pool exhaustion (`remaining connection slots are reserved for roles with the SUPERUSER attribute`), unrelated to this PR. `ores.analytics.quant` has no database dependency (pure math library), so it's unaffected.

mcraveiro:
Fixed in b5afccb57 — thanks for catching this.

1. cir_process::discount_factor overflow: fixed by rewriting B(tau)/A(tau) in the numerically-stable form that divides through by e^{gamma*tau} up front, keeping every exponent non-positive (gamma > kappa always for sigma > 0). Added a regression test with ticks_ahead up to 10000 asserting a finite result in (0,1].

2. hull_white_process_tests.cpp misleading test name — not changed, per your own note this is non-blocking and the real Vasicek cross-check exists elsewhere (vasicek_process_tests.cpp).

Also fixed a literal [[id:...]] placeholder link in task_document-stochastic-processes.org's prose that was breaking the CI site build (a pre-existing bug in that doc, unrelated to your findings but caught along the way).

PR #1557, opened 2026-07-13
[iam,refdata] Migrate party to the generated nats-event-cache

mcraveiro:
## Summary

Party is the first real consumer of the nats-event-cache facet: generates its cache under ores.iam and retires the hand-written party_cache.hpp prototype. Extends the facet with a generic aux-index mechanism so compute_visible_party_ids keeps working via the parent/child index, rather than becoming a party-only special case.

## Changes

- Extended the nats-event-cache facet with a generic, entity-agnostic aux-index mechanism: cache_aux_type flag + three paste points (type declaration, build-from-entries, extra methods), mirroring the existing custom_subscriptions paste-point convention
- Added an additive partitioned_cache::aux_type alias (Layer A) to support it -- no existing call site touches it
- party's model now sets cached_by: iam, read_for_cache: true, cache_aux_type: children_map, with paste-point blocks reproducing the hand-written prototype's children_map/compute_visible_party_ids logic exactly
- Generated party_cache.hpp/party_cache_registrar.hpp under ores.iam; deleted the hand-written ores.iam.core/service/party_cache.hpp prototype
- Rewired registrar.cpp (calls warm_and_subscribe_party_cache) and repointed auth_handler.hpp/account_handler.hpp/bootstrap_handler.hpp at the new include path and .lookup(...) method
- Verification: codegen python test suite 45/45 passed; --diff regression checks show unrelated entities and party's other (pre-existing, drifted) facets unchanged from baseline; grep confirms no stray references to the old path/method remain; no full C++ build available in this environment (vcpkg dependencies not installed), mitigated by close comparison against the previously-compiling hand-written logic being replaced

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Generic entity-mirror cache primitive + codegen facet (party as first consumer)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/iam_party_cache_invalidation/story.html) | AA4A5326-FA02-42D7-ACE2-9B27B20B19C7 |
| Task | [Migrate party to the generated cache: first real nats-event-cache consumer](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/iam_party_cache_invalidation/task_migrate_party_to_generated_cache.html) | 763C5969-19EA-4990-A662-BE2623740325 |


mcraveiro:

mcraveiro:
Fixed in f3c4ff227: gated <immer/map.hpp>/<immer/map_transient.hpp>/<vector> behind cache_aux_type (and <variant> now only emitted when is_uuid AND no cache_aux_type), added validate_cache_aux_type with test coverage, and restored the dropped 'Warming party cache for N tenant(s)' log line via a dedicated logger in the generated registrar-wiring snippet. Also reworded the task doc's 'exactly' to be precise about membership vs traversal order, and fixed the formatting nit. Left the pre-existing lack of cycle protection in compute_visible_party_ids as-is, per your own note that it's inherited from the hand-written prototype, not a regression.

mcraveiro:
Fixed in f3c4ff227: reworded the task doc so 'exactly' only claims visible-set membership, noting the iterative traversal order differs from the original recursive DFS (harmlessly, as you confirmed). CI did go green on this branch before merge.

mcraveiro:
Fixed in f3c4ff227: added validate_cache_aux_type (requires cached_by, mirrors validate_cached_by) with test_cache_aux_type.py, fixed the formatting nit on the cache_aux_type flag line, and restored the dropped warm-up log line. Left the cycle-protection gap as-is since you noted it's a faithful port of the original hand-written prototype's same gap, not a regression from this PR -- agreed a follow-up ticket would be the right place for that if parent_party_id integrity ever needs a defensive check.

PR #1566, opened 2026-07-14
[iam,refdata,codegen] Close entity-mirror cache story: auth check + bootstrap phases

mcraveiro:
## Summary

Closes out the remaining tasks on the Generic entity-mirror cache primitive story: verified the original activation-repro bug is fixed end-to-end, searched for (and found none of) other hand-rolled cross-service caches, and closed the read_for_cache auth gap -- which surfaced a real startup-ordering question for IAM authenticating against itself, resolved with a new architecture doc naming the general Phase 1/Phase 2 service-bootstrap pattern.

## Changes

- Verified via a fresh environment + live service fleet that party_cache reloads on the real save_party_request path with no IAM restart; scaffolded a companion manual test scenario for UI confirmation (descoped an automated Catch2/NATS integration test -- no existing precedent for one in this repo)
- Searched the codebase for other hand-rolled, cross-service entity-mirror caches to migrate -- none found; party's was the only instance (documented the search trail)
- Closed the read_for_cache auth gap: generated handler (and party's hand-written copy) now requires a valid signed JWT; generated cache attaches a service-account token via an optional provider (constructor + set_token_provider setter)
- New architecture doc + sequence diagram: Service Bootstrap Phases -- Phase 1 (wire/subscribe every subject) vs Phase 2 (mint tokens, warm caches, including self-referential calls), worked through IAM's party_cache; cross-linked from Anatomy of a Service and Service-to-Service Auth
- ores.iam.core/messaging/registrar.cpp restructured into the two phases; register_handlers gains a service_password parameter; ores.iam.core now depends on ores.iam.client.lib (verified non-circular)
- Story closed: all tasks DONE
- Verification: codegen python test suite 48/48 passed; --diff regression checks show no incidental impact on unrelated entities; no full C++ build available in this environment (vcpkg dependencies not installed) -- this PR touches more surface than usual (new CMake edge, a public signature change, a startup-ordering restructuring) so a real build before merge matters more than usual here

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Generic entity-mirror cache primitive + codegen facet (party as first consumer)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/iam_party_cache_invalidation/story.html) | AA4A5326-FA02-42D7-ACE2-9B27B20B19C7 |
| Task | [Add auth check to the read_for_cache archetype](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/iam_party_cache_invalidation/task_add_auth_check_to_read_for_cache_archetype.html) | 582D9310-5859-418C-97B5-47AF62A36B74 |


mcraveiro:

mcraveiro:
Fixed in 43de66c74: added an explicit code comment on read_for_cache (both the archetype and party's hand-written copy) noting it's authentication-only, not tenant-scoped, so it isn't copy-pasted as a tenant-authorized template. Also documented the narrow Phase-1/Phase-2 startup race the other two reviewers flagged, and added rendering tests for the auth check + token-provider wiring.

mcraveiro:
Fixed in 43de66c74: added a 'Known limitation' section to service_bootstrap_phases.org documenting the exact race you traced (login lands between auth_handler subscribing and party_cache being armed) as an accepted, self-healing tradeoff, plus a Checklist item to watch for it in future services; added the authentication-only/not-tenant-scoped code comment on read_for_cache; added test_read_for_cache_auth_rendering.py so the auth-check/token-provider code is now asserted against rendered template output, not just model-flag validation.

mcraveiro:
Fixed in 43de66c74: documented the startup race explicitly (accepted, self-healing, with a Checklist item) rather than leaving it implicit; added a code comment clarifying read_for_cache's authentication-only/not-tenant-scoped design so it doesn't get copy-pasted assuming tenant isolation; added rendering tests asserting the auth check and token-provider wiring appear in the generated output. Left the application.cpp value-capture as-is per your own confirmation it's correct given register_fn runs exactly once.

PR #1570, opened 2026-07-14
[qt,marketdata,sql] Redesign CRM cross-rates matrix UI with real manual QA fixes

mcraveiro:
## Summary

Delivers the CRM cross-rates matrix Qt panel after 6 rounds of real manual QA against live data (not just build/tests) -- picker-based per-named-CRM windows, Hide Empty/Show Inverted toggles, auto-refresh, persisted geometry -- plus several real bugs it surfaced along the way: a missing NATS event subscription blocking CRM engine refresh, an RLS policy blocking crm_ingest_bridge's cross-tenant startup scan, an always-editable id field on every UUID-keyed entity's Add dialog (fixed at the codegen template level), and a silent NATS max-payload timeout in Market Simulator's feed start.

## Changes

- CrmCrossRatesMatrixMdiWindow/Controller redesign: picker-on-open (custom app-styled dialog, not QInputDialog), per-name window reuse, independently persisted geometry/toggle state, Hide Empty and Show Inverted footer toggles, auto-refresh on by default, new CrmRateCellWidget/CrmRateSparklineWidget
- Fixed crm_ingest_bridge missing crm_enabled_derived_pair_changed_event subscription (derived-pair-only changes never refreshed the rate engine)
- Fixed RLS blocking crm_ingest_bridge's cross-tenant startup scan after the CRM tables moved to ores_refdata_* (added a scoped permissive read policy + the RESTRICTIVE party-isolation policy the schema validator expects)
- Fixed always-editable id field on every UUID-primary-keyed entity's Add dialog, at the shared codegen template level (new is_auto_generated_key flag); filed a follow-up story for server-side UUID generation
- Fixed silent Market Simulator feed-start timeout from an oversized market_observations.list NATS reply; added paginated list_observations_page()
- Added a third named CRM tier (scandies: EUR + Nordic currencies), real Fed H.10 data for USD/NOK and USD/DKK via tools/fetch_fed_h10_rates.py, matching synthetic feed configs
- Verification: local build clean (linux-clang-debug-make); ctest 73/73 passed

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [CRM cross-rates matrix UI panel](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/crm_implementation/task_crm_cross_rates_matrix_ui.html) | C685A8BE-0DF9-4D23-8AAB-E9DF9069E482 |


mcraveiro:

mcraveiro:
Fixed in 84fb3694b — Book's editable id field is already tracked separately (book_id_field_not_readonly.org, filed before this PR) and out of scope here since no Book code is in this PR's diff. Everything else (client-side inversion/delta as known debt) is already captured in task_crm_move_numeric_logic_to_quant.org.

mcraveiro:
Fixed in 84fb3694b — both confirmed bugs fixed: (1) inverted-cell sparkline now records history under the same display_key/display_rate computed in the per-cell rendering loop (was a separate pass keyed by the server's own pair direction), (2) Book's editable id field is a pre-existing, separately-tracked gap (book_id_field_not_readonly.org, filed before this PR), not something to fold into this PR's diff.

mcraveiro:
Superseded by the fuller review posted right after (comment 4974581959) — see that reply for the fix details.

mcraveiro:
Fixed in 84fb3694b — both confirmed bugs fixed: (1) onCellSelected now indexes into separate displayedRowCurrencies_/displayedColumnCurrencies_ lists instead of one list only populated from rows, (2) CrmCrossRatesMatrixController now emits detachableWindowDestroyed when a matrix window closes. Book's editable id field is already tracked separately (book_id_field_not_readonly.org, filed before this PR) and out of scope here since no Book code is in this PR's diff.

PR #1572, opened 2026-07-15
[refdata,synthetic,dq,qt] Instrument code shorthand, asset_class badge, CRM subject fix

mcraveiro:
## Summary

Follow-up polish on the Curve Template product catalogue (short FIX-inspired codes, asset_class badge, full descriptions), menu wiring for the previously-unreachable IR curve config screens, and a fix for a long-standing crm_topology_bundles seed-data regression that silently hung party provisioning.

## Changes

- instrument_code: redesigned code as a short, FIX-inspired mnemonic (FRA, IRS, CDS, FXBAR...) instead of ORE's own oreTradeType spelling, with a new nullable ore_trade_type column preserving the literal ORE string for future trade-file interop.
- asset_class_code/instrument_code: full paragraph-length descriptions for all 8 asset classes and all 115 instrument codes; added an asset_class badge (8 dedicated colors) shown on instrument_code's list view.
- SyntheticPlugin: wired ir_curve_generation_config/ir_curve_template_entry Qt CRUD into Market Data > Synthetic > Advanced (generated by codegen in a prior PR but never reachable in the UI).
- Fixed a regression (introduced by d07edfbdd) where crm_topology_bundles' DQ artefact seed pointed at the wrong (marketdata) NATS subject, causing any party provisioning that publishes this bundle to hang indefinitely rather than fail loudly.
- Verification: local build clean (linux-clang-debug-make); ctest 73/73 passed; db recreate clean; Barclays Plc provisioning verified end-to-end including the previously-hanging crm_topology_bundles step (now completes in ~2s).

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [IR Rates synthetic data generation](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/ir-rates-synthetic-generation/story.html) | 3ECC4BA8-6ACA-4028-9E42-AE7F25FC3B98 |
| Task | [Tenor-collision/gap validation on the Curve Template](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/ir-rates-synthetic-generation/task_tenor-collision-validation.html) | E6B7DEF3-F35F-454A-8ECF-CCD9E8E07A83 |


mcraveiro:

mcraveiro:
Not changed. Both observations are pre-existing, generic patterns rather than gaps in this PR: the opt_str<T> bool branch is the standard codegen table-printer template (identical in currency_table.cpp, deposit_convention_table.cpp, etc.) generated for every optional field regardless of type, not something worth special-casing per-entity. The unwired instrument_code_history_field_mapper matches the same not-yet-wired state of most other *_history_field_mapper files (tenor, calendar, business_centre...) -- only currency/country are wired today, which is a separate, pre-existing rollout gap, not something this PR introduced incompletely.

mcraveiro:
Not changed. Same reasoning as the other review pass: opt_str<T>'s bool branch is the shared codegen table-printer template used identically across every entity with an optional field (currency_table.cpp, deposit_convention_table.cpp...), not something specific to instrument_code worth simplifying in isolation. The unwired history field mapper matches the existing not-yet-wired state of most other *_history_field_mapper files in the repo -- a separate, pre-existing rollout gap.

mcraveiro:
Fixed in 31fe486be — corrected all 13 grammar-bug descriptions, and added an idempotent backfill INSERT for crm_topology_bundles that corrects any already-provisioned database still carrying the old marketdata_* values (verified against a simulated pre-fix row: value corrected, version bumped correctly, re-running is a no-op once fixed). Apologies for missing these two in the earlier reply round -- I'd only addressed the two later, non-blocking observations.

mcraveiro:
Good catch -- they were genuinely missed, not a deliberate decision. Fixed both in 31fe486be: grammar bugs corrected across all 13 rows, and an idempotent backfill added for crm_topology_bundles so the subject fix reaches already-provisioned databases, not just fresh recreates. Verified the backfill against a simulated pre-fix row.

mcraveiro:
Fixed in c91d981ae -- applied the same fix to the two remaining COMMBVS/COMMBSKT rows for consistency with the already-fixed FX/Equity siblings.

PR #1581, opened 2026-07-15
[refdata,qt,codegen] Commission party, counterparty, and party_status

mcraveiro:
## Summary

Finishes commissioning `party`, `counterparty`, and `party_status` across the Qt UI, superseding the three abandoned per-entity Sprint 22 stories now that the composite child-entity/hierarchy widgets and temporal composite versioning blockers are both DONE.

- Regenerated `party`/`counterparty` against current codegen templates (zero-diff), migrated Qt from `ores.qt.party` to `ores.qt.refdata`.
- Fully commissioned `party_identifier`, `counterparty_identifier`, `party_contact_information`, `counterparty_contact_information` under codegen — deleted all bespoke hand-written NATS handlers and ad-hoc Qt forms, replaced with generated dialogs (own explicit Save button) bolted onto the composite parent, with zero loss of composite-versioning behavior.
- Built the composite child-entity table + hierarchy Qt facet (generic `ChildEntityTableWidget`, `HierarchyModelBuilder`/`HierarchyTreeWidget`) shared by party/counterparty.
- Adopted the new consolidated generic `HistoryDialog` for both entities and all 4 child types, deleting the legacy per-entity `XHistoryDialog` classes.
- Verified live end-to-end (list/read/update/delete/identifiers/contacts/hierarchy/history/cross-client eventing) against a running environment, fixing every regression found along the way rather than filing around it:
  - `entity_type_of(party)` missing (party's hand-patched domain profile never picked up generic `HistoryDialog` dispatch).
  - `crm_topology_bundles` DQ artefact type seeded with a `marketdata` prefix instead of `refdata`, permanently hanging tenant provisioning.
  - `party_category` NULL-constraint Create failure (missing server-side default).
  - Codegen `is_auto_generated_key` bug: matched any `is_key` Detail field on an entity with a UUID primary key, not specifically the field that *is* that key — wrongly locked party/counterparty's natural `short_code` permanently read-only, silently blocking Create.
  - `has_pagination: true` never set on party/counterparty's Qt facet — reported the current page's row count as the total, permanently hiding anything past page 1.
  - `changeReasonCache` never forwarded from the parent detail dialog to its child identifier/contact-information tables, silently blocking every LEI/contact-information save.
  - 8 stale `ores.refdata.core.tests` files (pre-context-refactor repository signatures) plus 2 synthetic-generator bugs (invalid enum-suffix values, hardcoded unseeded country codes) the fix then surfaced.
- Filed follow-up tasks for scope intentionally left out: a parent-party picker (Create currently can't set `parent_party_id`, so a second party/counterparty can't be created — DB-enforced single-root-party constraint has no UI path around it yet), wiring `generate_short_code()` into Create dialogs, removing the now-dead `EntityDetailDialog`/`EntityDetailOperations` abstraction, and linking composite parent history to the specific child change that caused it.

## Test plan

- [x] `ores.refdata.core.tests` — all 113 cases pass
- [x] Full backend + Qt build clean (`ores.refdata.core/api/service.lib`, `ores.qt.refdata.lib`, `ores.qt.exe`)
- [x] Live manual verification against BARCLAYS PLC in two Qt client instances (blue/red): list, read, update, delete, identifiers tab (add/edit/delete), contact information tab (add/edit/delete, country flag combo), hierarchy tab, history dialog, cross-client eventing
- [ ] Create (second party/counterparty) — intentionally deferred, tracked as a separate follow-up task


mcraveiro:
Thanks for the thorough pass. Addressed in 0705f18d5:

1. **Pagination bug in the 4 child list models** — Fixed in <sha 0705f18d5>: added `:has_pagination: true` to party_identifier/counterparty_identifier/party_contact_information/counterparty_contact_information's `.org` models and regenerated. Same root cause as the parent-list bug, now consistent across all six paginated lists.

2. **Generator fix durability risk** — Fixed in <sha 0705f18d5>. Root-caused to the domain-type generator template (`cpp_domain_type_generator.cpp.mustache`) unconditionally appending a "-<idx>" uniqueness suffix to any natural-key text column's `generator_expr`, which corrupts validated-enum columns. Added a `no_generator_suffix` per-field facet (plain drawer property — no core.py changes needed, the org loader already copies arbitrary properties through) and moved the rotating-valid-values expression into the `.org` generator blocks themselves, so it survives regeneration. Verified via `codegen entity generate <entity> --diff` showing a clean, minimal diff.

3. **Nit: duplicate scheme codes instead of `party_id_scheme_constants::all`** — Not changed. This matches the existing codebase convention exactly: other generators with FK-validated/enum-like fields (`party_generator.cpp`'s `party_type`/`business_center_code`/`status`, `currency_pair_generator.cpp`'s `base_currency`/`quote_currency`) already hardcode plain literals directly rather than reusing a constants header. Reusing the header would add a new dependency generators don't otherwise have, for a benefit (avoiding duplicate literals) the rest of the codebase doesn't currently value enough to do itself.

4. **PR description discrepancy re: crm_topology_bundles/party_category** — Not changed (description accurate). Both fixes are genuinely present in this branch's history (commit 24d5d38de for crm_topology_bundles; the party_category server-side default was hand-patched onto party.hpp earlier in the branch, since party's domain profile is parked/not auto-regenerated). Likely just missed in the diff search against origin/main given the branch's size.

5. **Minor: ClientCounterpartyModel dropped Transliterated Name column** — Not changed, captured separately as a backlog item ([party_grid_ignores_transliterated_name](https://github.com/OreStudio/OreStudio)) since it's a pre-existing gap, not introduced by this PR.

All 120 `ores.refdata.core.tests` cases pass; full backend + Qt build clean.

PR #1592, opened 2026-07-15
[controller] Disable ores.wt.service by default (crash-loop hotfix)

mcraveiro:
## Summary
- `ores.wt.service`'s Wt httpd crashes intermittently on startup with a fatal glibc error: pthread-priority-protect assertion in `tpp.c`, inside Wt's own threading internals — not ORE Studio code.
- The existing controller retry/backoff self-heals eventually, but noisily: up to 5 minutes between attempts, shows "failed" status in the meantime.
- Root-causing needs a debugger session against Wt/Boost's compiled internals to find the exact call site. Until then, disable the service by default in the controller service-definitions seed so fresh environments don't hit the crash loop.

## Changes
- Set `enabled = 0` for `ores.wt.service` in `controller_service_definitions_populate.sql`; every other service unaffected.
- Documented the crash and a one-line SQL re-enable command inline.
- Applied the same change directly to the local `bright_faraday` database; confirmed `ores.wt.service` no longer spawns and the remaining 23 services are healthy.

## Test plan
- [x] `services status` after `services stop && services start`: `ores.wt.service` absent, 23/23 other services running.
- [ ] Fresh `db recreate` on another environment picks up `enabled = 0` for `ores.wt.service`.

### DIFF
--- a/doc/agile/product_backlog/inbox/root_cause_wt_service_glibc_crash.org
+++ b/doc/agile/product_backlog/inbox/root_cause_wt_service_glibc_crash.org
@@ -1,0 +1,66 @@
+:PROPERTIES:
+:ID: E6902594-0075-4B66-ADDC-4D79622CCF41
+:END:
+#+title: Root-cause ores.wt.service's intermittent glibc pthread crash
+#+description: ores.wt.service was disabled by default (hotfix, PR #1592) after its Wt httpd crashed intermittently on startup with a fatal glibc pthread-priority-protect assertion. Needs a debugger session to find the exact call site inside Wt/Boost's compiled internals and either fix it or work around it properly, then re-enable the service.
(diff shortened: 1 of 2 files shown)


mcraveiro:
Thanks for the thorough review. Findings 1/2 confirmed correct, no action needed. #3 (idempotency) and the two test-plan checkboxes already acknowledge this honestly -- no code change warranted, just a known limitation of this insert-only seed pattern. #4 (tuple structure) declined for now: a single well-commented one-off case-when is proportionate; happy to promote it to a per-row enabled column if/when a second service needs the same treatment. #5 accepted: filed doc/agile/product_backlog/inbox/root_cause_wt_service_glibc_crash.org with full crash details so the disable-by-default doesn't quietly become permanent.

PR #1594, opened 2026-07-15
[codegen,qt] Add required parent picker to party/counterparty Create

mcraveiro:
## Summary

Closes the last blocker on the party/counterparty commissioning story: Create had no way to set `parent_party_id`/`parent_counterparty_id`, and the DB enforces exactly one root party per tenant — so creating any second party or counterparty always failed with a duplicate-key error on `parties_root_party_uniq_idx`, with no UI path around it.

- Added a required "Parent Party"/"Parent Counterparty" `dynamic_combo` to both Create/Edit dialogs, self-referencing (populated from the existing party/counterparty list via new `fetch_parties`/`fetch_counterparties` `LookupFetcher` functions).
- Generalizes the existing codegen `dynamic_combo` facet to a self-referencing UUID-valued combo — reused almost all existing machinery (`is_uuid`/`combo_code_field` for UUID-keyed combos, `required_dynamic_combo_fields` for the disabled-Save-until-selected behavior) rather than building new infrastructure.
- Real gap found and fixed at the template source (`ores.cpp.qt.detail_dialog_impl.org`, tangled into `cpp_qt_detail_dialog.cpp.mustache`): the non-locked `is_dynamic_combo` branches of `updateUiFromX()`/`updateXFromUi()` only handled plain `std::string`-valued combos, not `is_uuid`/`is_optional_uuid` (needed since `parent_party_id` is nullable — the root party genuinely has none). Extending this surfaced and fixed a related latent bug: the item-to-combo-value extractor lambda was gated on the *current field's* uuid-ness instead of whether `combo_code_field` on the *fetched* entity is a uuid column — these coincided for every prior usage but not for this self-referencing one.
- Self-parenting is impossible by construction on Create (the new entity's id doesn't exist yet when the combo populates). Descendant-cycle prevention on Edit is explicitly out of scope for this pass.
- Verified zero regression via `codegen entity generate <entity> --diff` roundtrip across all 9 other entities using `combo_domain_type`.

## Test plan

- [x] `ores.refdata.core.tests` — all 120 cases pass
- [x] Full backend + Qt build clean
- [x] Live manual verification against BARCLAYS PLC (scenario `709B8639-638A-48D5-876C-A460F29B1BCB`, PASSED): required combo present on both Party/Counterparty Create, Save stays disabled with no selection, second party/counterparty created successfully with a parent selected, edit round-trips the selected parent correctly


mcraveiro:
Thanks for catching this — genuine bug. Fixed in 88aa06e77:

1. **Self-parenting on Edit / recursive CTE hang** — Fixed. Added a generic `exclude_if` predicate parameter to the shared `populateDynamicCombo` helper (`DynamicComboSetup.hpp`), applied on the main thread right after the fetch completes — safe, since `setParty()`/`setCounterparty()` always runs synchronously before the async fetch's completion handler fires, so reading `party_.id`/`counterparty_.id` there is race-free (unlike inside the fetch functor itself, which runs on a background thread). Added a new `is_self_referencing_combo` codegen flag (`core.py`), computed purely by comparing `combo_domain_type` against the entity's own `domain_class` — no new model property needed. Verified via full regen that every other entity using `combo_domain_type` (9 of them) now gets an explicit no-op `[](const auto&) { return false; }` in that slot, confirming zero behavior change elsewhere.

2. **Sort by `version` not alphabetically** — Not changed. Already called out as a deliberate tradeoff in the PR description: `populateDynamicCombo`'s `sort_key_of` is `int`-typed and neither `party` nor `counterparty` has a more meaningful int ordinal. Widening that widely-shared generic helper's signature (e.g. to support a string/lexicographic sort) for this one combo felt like more risk than the benefit justified for this pass.

3. **Save may not stay disabled by default (QComboBox auto-selects index 0)** — Not changed. Confirmed this is real, but it's pre-existing, shared behavior already in production across 5+ other entities' required dynamic combos (`crm_driver_pair`, `crm_enabled_derived_pair`, `tenor_convention` x2, `business_centre`, `instrument_code`) — not something this PR introduces. Fixing it means changing `populateDynamicCombo`'s default-selection behavior for all of them, which needs its own re-verification pass rather than riding along here. Filed for visibility but out of scope for this PR.

4. **`counterparty.hpp` include ordering** — Fixed in 88aa06e77.

All 120 `ores.refdata.core.tests` cases pass; full build clean.

mcraveiro:
Thanks for the thorough review — no changes needed here, confirming the include-ordering nit you flagged was addressed as part of the other review's self-parenting fix (commit 88aa06e77).

PR #1618, opened 2026-07-17
[compass,llm] Deny piped/redirected compass commands via PreToolUse hook

mcraveiro:
## Summary
- Add a PreToolUse hook (`deny_piped_compass_hook.py`) that blocks any Bash command invoking `compass.sh` (any of its 4 granted forms) when the same command string also contains a pipe or redirect. Permission allow/deny rules can't express this — they evaluate compound commands per-subcommand, so a pipeline's `compass.sh ...` and `| tail` pieces are each independently authorised with no way to see "compass piped into tail" as a shape to deny.
- Give `compass build --direct <target>` (site/settings/skills/...) the same well-known, reported log-path treatment `compass build`'s cmake path already had (`📝 Build output: <path> (tail -f to follow)`), instead of silently swallowing emacs batch-mode output.
- Document the rule prominently in CLAUDE.md (read at the start of every session) so it's durable across sessions, not just this one.
- File the follow-up task under the "Compass improvements" story: move all compass output from ad-hoc `/tmp` files into a rotated, per-checkout `./tmp/compass_output/` directory (log rotation, command-executed header, build-lock pointing at the new location), plus a new services-start lock capping concurrent running environments at 4. Not implemented in this PR — this PR only lays the groundwork (the hook + the reported-log-path pattern) that task will build on.

## Test plan
- [x] `python3 -m py_compile` on both changed/added Python files.
- [x] `./compass.sh build --direct settings` regenerates `.claude/settings.json` cleanly with the new `hooks` key.
- [x] Hook script exercised directly against representative bare/piped/redirected commands, confirming allow/deny behavior and that unrelated pipes (`git log | head`) and standalone `tail -f <logfile>` are unaffected.
- [x] Docs-only additions (task/story wiring) re-indexed via `compass index --org-roam-db-sync` with no errors.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,6 +2,13 @@
 
 Run `./compass.sh bearings` to orient yourself before acting.
 
+**Never pipe or redirect a `compass` command.** Run it bare — no `|`,
+no `2>&1`, no `>`. Every `compass build`/`deploy`/`site` command that
+produces meaningful output prints its own well-known log file path
+(e.g. `📝 Build output: /tmp/<label>_<target>_build.log`). To watch
+progress, `tail -f`/`tail -n` *that reported file* as a separate,
+standalone command — never by piping the compass invocation itself.
+
 **Memory rule:** All project memories live in `doc/llm/memory/` and must
 be created with `compass add memory`. Never write to the Claude Code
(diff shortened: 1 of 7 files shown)


mcraveiro:

mcraveiro:

mcraveiro:

PR #1628, opened 2026-07-18
[codegen,agile] Knobbed entity_org scaffolding (--shape) and post-generation checklist

mcraveiro:
## Summary
Closes task EE237306 (Scaffold model .org files per entity type), the follow-on to PR #1620's documentation work.

- `compass add entity_org --shape <name>` (`simple-text-key`, `fk-scoped`, `hierarchical-composite`, `richest`, `timeseries`) scaffolds a knobbed entity model instead of the previous bare, unknobbed skeleton (empty `PROPERTIES` drawers, no Qt/foreign-key/insert-trigger sections at all).
- The template (`doc_entity_org.org.mustache`, tangled from `ores.doc.modeling.entity_org.org`) carries the non-structural variability as mustache section-conditionals gated by the shape's knob bundle (`ENTITY_ORG_SHAPE_PRESETS` in `doc_generate.py`) — one template covers every shape, no per-shape file copying or per-shape template duplication. Individual `--entity-<knob>` flags override any preset value.
- Anything a shape can't derive (actual columns, FK targets, display fields, CSV/XML export class names, ...) is left as a visible, one-line `TODO` in the scaffolded output — never silently omitted, per the acceptance criteria.
- New "Codegen entity — post-generation checklist" page enumerates the fixed manual integration steps every newly-generated entity still needs (sub-registrar wiring, event-registrar wiring, CMake wiring, Qt plugin wiring), each explaining why it isn't generated yet and linking the capture/task tracking automation where one exists — replaces a dangling "see the hub's frontmatter" reference that pointed nowhere.
- Filed a capture to audit whether every facet has both a generated TS-inventory link and a hand-authored TS-description link, specifically checking `ores.codegen` itself.

The five shape presets are reconnaissance picks sampled from one exemplar each, not a final catalogue — expect refinement as the sibling task (819B050A) actually uses the scaffold to author the DQ reference entities.

## Test plan
- [x] `compass build --direct tangle_codegen_templates` re-tangles cleanly with no drift beyond the intended change
- [x] All five `--shape` presets exercised end-to-end via the real CLI; verified correct section presence/absence (Qt on/off, Foreign keys, hierarchy fields) against the regenerated `.mustache`
- [x] `compass build --direct site` succeeds cleanly (no broken id-links)
- [ ] Reviewer pass over the shape-preset knob bundles for accuracy against the real exemplar entities


mcraveiro:
Fixed in 496bfb3c5 — verified each finding directly against the real exemplar .org files and corrected all four presets: richest now sets has_tenant_id (currency.org has it true in both frontmatter and Flags); simple-text-key drops the spurious has_export_macro and adds has_insert_trigger_validations/qt_has_change_reason_cache/qt_has_pagination (matches country.org); hierarchical-composite drops the spurious has_foreign_keys/qt_has_export_macro claim against party.org alone and instead matches counterparty.org (the richer of its two named exemplars, which does have Foreign keys + Insert trigger + has_change_reason_cache, but not has_export_macro); timeseries drops the spurious has_foreign_keys (market_observation.org's party_id/series_id are plain Natural keys, no Foreign keys section). fk-scoped needed no changes. Also updated the 'How do I add a new codegen entity model?' recipe to cover --shape/--entity-<knob> and link the post-generation checklist, per a separate request.

PR #1635, opened 2026-07-18
[dq] Author the five DQ reference-entity codegen models (819B050A)

mcraveiro:
## Summary
- Authors codegen `.org` models for catalog, change_reason_category, change_reason, data_domain, subject_area -- the five DQ entities named in story FFA8F092/task 819B050A that had no codegen model at all.
- All modeled as `lookup_entity` (natural-key-only bitemporal PK, no UUID surrogate), matching each entity's real SQL schema -- confirmed by checking `projects/ores.sql/create/dq/*.sql` before authoring, not assumed.
- Each model's SQL generation is verified against its real hand-written table (`compass codegen generate --address ores.sql.schema`, diffed, generated file reverted -- only the `.org` model is committed).
- `sql_schema_table_create.mustache` unconditionally emits a `change_reason_code` column for every `lookup_entity` (no suppression knob exists, unlike `domain_entity`'s `#+no_audit_columns`); the hand-written `change_reason_category` and `change_reason` tables both predate it (avoiding the circularity of a change-reason record referencing its own change reason) -- generating and swapping either table over will need to account for that column, not just cosmetic formatting differences.
- `subject_area`'s real table has a compound natural key (`name` + `domain_name`), which `lookup_entity` doesn't support yet (task 52FE40FA, BACKLOG); per an earlier user-approved descoping decision, `domain_name` is modeled as a plain `Column` with a soft-FK `Validations` entry instead -- documented in the model with the resulting weaker PK/uniqueness coverage.
- 819B050A is left `BLOCKED` on task 464B5EC6 (migrate lookup_entity into domain_entity) -- C++ generation and retiring the hand-written core/repository/service/handler/protocol code is out of scope until that migration lands (see 464B5EC6/DDBD522D for why).

## Test plan
- [x] `compass codegen generate --model <entity>.org --address ores.sql.schema --dry-run/real` for all five entities, each diffed against its real hand-written table
- [x] catalog, change_reason_category, data_domain: structurally correct match (only cosmetic differences plus an auto-generated validate_fn the hand-written originals predate; change_reason_category/change_reason also gain a change_reason_code column the hand-written originals predate, see above)
- [x] change_reason: soft-FK validation and category_code index both generate correctly via Validations/Indexes sections
- [x] subject_area: generates the expected (intentionally weaker, single-column-PK) DDL shape per the descoping decision


mcraveiro:
Fixed in 8c7ae14b1 — restored task_fix-lookup-entity-model-type-mapping.org from origin/main (an earlier rebase-conflict resolution had used --theirs backwards, keeping this branch's own stale stub instead of main's DONE content). Also fixed: added the same change_reason_code-column caveat to change_reason's bullet (matching change_reason_category's) and to the PR description, since the same sql_schema_table_create.mustache gap applies to both.

PR #1638, opened 2026-07-18
[qt] Add calendar assignment UI: combo picker, flag icons, junction widgets

mcraveiro:
## Summary

Adds the Qt screens for calendars: a reusable combo picker and junction-assignment widget, flag icon support, and cross-domain navigation from Currency Pair Conventions.

## Changes

- Reusable OreCalendarComboBox / CalendarAssignmentWidget pair for editing currency-to-calendar and currency_pair_convention-to-calendar junction assignments, mirroring AccountRolesWidget's staged pending-adds/removes pattern
- Calendar flag-icon support in ImageCache/FlagIconHelper, plus a fetch_calendar_codes lookup
- CalendarAssignmentWidget embedded in Currency and Currency Pair Convention detail dialogs' new Calendars tab
- Junction NATS/service support for currency_calendars and currency_pair_convention_calendars added via currency/currency_pair_convention's own protocol/service/handler/registrar through codegen paste points
- Toolbar button on CurrencyPairConventionMdiWindow opening the Calendars list, wired via a cross-domain controller relay mirroring the existing Book to BookStatus pattern

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Model calendars as proper ORE Studio reference data](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/calendar-reference-data/story.html) | E1196536-38E8-4365-B0E6-A269F7CA3923 |
| Task | [Qt screens for calendars: list/add/edit, combo picker, flag display](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_23/calendar-reference-data/task_calendar-qt-screens.html) | 24FCC62F-D1AB-4872-8337-E806675EBF7C |


mcraveiro:

mcraveiro:
Fixed in 3a3d7b392 — assignmentsChanged no longer routes through onFieldChanged(); it only enables Save. onSaveClicked() now skips the entity save when only calendar assignments are pending, so calendar-only edits produce a single change-reason prompt (the calendar commit's), not two. Also fixed the commitChanges() partial-failure/doc mismatch you flagged in issue 3: it now always reloads after a commit attempt, dropping already-applied items from the pending lists on partial failure, and the doc comment now matches the actual stop-at-first-failure behaviour. Issue 2 (tab reload discarding staged edits on tab-switch) is a real gap but a larger behavioural change than this round's fix — deferring it as a follow-up rather than risking a rushed fix; the AccountRolesWidget precedent you cited only reloads once from setAccount(), and CalendarAssignmentWidget should do the same, but that requires tracking whether it's already been loaded for the current leftKey, which I don't want to rush in this round.

mcraveiro:
Fixed in 3a3d7b392 — calendar-only changes no longer trigger a spurious no-op entity save/version bump; onSaveClicked() checks whether the entity's own fields changed (hasChanges_) separately from the calendar widget's pending changes, and skips the entity save when only the latter are set. On issue 2 (Calendars tab unusable in create mode): confirmed, this is a real functional gap relative to the AccountRolesWidget/AccountPartiesWidget precedent. Given the scope of this round I'm deferring the pendingAdds-capture-before-create restructuring as a follow-up rather than rushing it — will track it against the parent story. Issue 3 (read-only not reapplied when paging history) is fixed: setReadOnly()/setReadOnly(bool, int) now reapply to the calendar widget when its tab is active.

mcraveiro:
Fixed in 3a3d7b392 — consolidated so a calendar-only change no longer fires both handlers on one Save click: assignmentsChanged now only enables the Save button (doesn't set hasChanges_), and onSaveClicked() returns early when only calendar assignments are pending, leaving the calendar-commit handler as the sole prompt/save for that case. On the minor note: commitChanges() now always reloads after a commit attempt (not just on full success), so a partial failure drops already-applied items from the pending lists instead of leaving them staged for a resend. On the PR description wording: fair point — currency_protocol.hpp/handler/registrar for calendars were added in the prior PR (#1625) that landed the junction repository/NATS backfill; this PR only adds the Qt-side consumers of that existing protocol. Will keep that distinction clearer going forward.

mcraveiro:

mcraveiro:
Fixed in cd1bd9cd8 — commitChanges() now tracks which adds/removes were actually applied before stopping at the first failure, and its finished handler trims only those out of pendingAdds_/pendingRemoves_. load() now distinguishes a same-key forced refresh (post-commit) from a fresh/different-key load, and only clears pending state in the latter case, so the unattempted remainder now genuinely survives a partial-failure reload as the doc comment promises.

mcraveiro:

mcraveiro:
Fixed in fdcd671fd — consolidated onSaveClicked into a single change-reason prompt for both writes. calendars_tab_save_guard now handles the calendar-only case directly (its own prompt, commit, return) instead of deferring to a separate saveButton connection, which has been removed entirely from both CurrencyDetailDialog and CurrencyPairConventionDetailDialog. For the combined-change case, a new base-template paste point (on_save_success) runs inside the entity save's async finished handler, only once the entity save has succeeded, and reuses that same prompt's reason/commentary (threaded in via the watcher lambda's capture list) to commit the calendar changes — so there's exactly one change-reason dialog per Save click regardless of what changed.

mcraveiro:

PR #1641, opened 2026-07-18
[dq] Migrate the five DQ lookup_entity models to domain_entity (464B5EC6)

mcraveiro:
## Summary
- `lookup_entity` (`#+type: ores.codegen.lookup_entity`) turned out to be architectural duplication, not a distinct metatype: `domain_entity` already supports natural-key-only entities (no UUID surrogate) end to end, including `coding_scheme`/`image_id` (already existed, used by dozens of real refdata entities like `country`/`currency`).
- Piloted the migration on `catalog` first: verified both SQL and C++ generation against the real hand-written table and pre-existing (unwired) generated files, confirming `domain_entity`'s pipeline produces *better* output than `lookup_entity`'s ever did (proper `tenant_id` type instead of `std::string`, an added `entity_type_of()` dispatch function, `clock_timestamp()` instead of `current_timestamp` for same-transaction correctness, `security definer set search_path` hardening).
- Migrated all five DQ entities from 819B050A: `catalog`, `change_reason_category`, `change_reason`, `data_domain`, `subject_area`. Each verified the same way (real SQL generation, diffed against the real hand-written table, reverted -- only `.org` models committed).
- `subject_area`'s compound natural key (`name` + `domain_name`) still isn't a true composite primary key under `domain_entity` (52FE40FA remains open and necessary), but now carries a compound *unique index* enforcing the pair for current rows -- stronger than the `lookup_entity` version, which had no such enforcement at all.
- `has_artefact_insert_fn` turned out to be dead metadata, unused by any SQL/C++ template on either metatype -- not ported.
- 10 `lookup_entity` models remain outside `dq` (iam/trading/analytics); filed as a follow-up task (AEFAE247) rather than blocking 819B050A further. `ores.codegen.lookup_entity`'s type/loader/dispatch isn't retired yet -- still has 10 live consumers.
- 819B050A is unblocked and back to `STARTED`.

## Test plan
- [x] `compass codegen generate --model <entity>.org --address ores.sql.schema` (real, non-dry-run) for all five entities, each diffed against its real hand-written table
- [x] `catalog`'s C++ generation additionally verified against pre-existing tracked generated files (caught and fixed an `rm -f`-deletes-tracked-files trap during cleanup -- restored via `git checkout --`)
- [x] Full `ores.codegen` pytest suite (53 tests) green


mcraveiro:
Fixed in a67fa4736 — you're right, has_artefact_insert_fn is live (sql_schema_artefact_create.mustache, wired via get_schema_template_mappings()), not dead as claimed. Corrected the claim in task_migrate-lookup-entity-into-domain-entity.org's Acceptance/Notes/Result and added an explicit warning to AEFAE247's Notes so it doesn't silently drop lei_entity/lei_relationship's artefact insert function when migrating them. Confirmed harmless for this PR specifically since none of the 5 migrated entities used the flag.

PR #1644, opened 2026-07-19
[dq] Commission catalog fully onto the generated codegen stack

mcraveiro:
## Summary
- First of the five DQ entities from 819B050A commissioned end to end (picked as the pilot per user decision to do one entity fully before the rest).
- Regenerated for real: domain, generator, repository/entity/mapper, service, protocol, NATS handler, NATS sub-registrar.
- catalog's hand-written service/protocol/handler were combined with data_domain/subject_area/dataset_dependency under `data_organization_service`/`_protocol`/`_handler` -- removed catalog's methods/structs/handlers from all three (other entities untouched) and wired the generated `catalog_service`/`_protocol`/`_handler`/`_registrar` into `registrar.cpp` via the standard `register_catalog_handlers()` sub-registrar call, matching the `badge_definition`/`badge_severity`/`code_domain` convention already there.
- `catalog_repository`/`_entity`/`_mapper` already existed as tracked, generated-shaped files predating catalog's demotion to `lookup_entity`-only; regenerating moved the repository from stateful (constructor-injected context) to the now-standard stateless (context per call) shape.
- The new `catalog_protocol.hpp` renamed `delete_catalog_request.codes` -> `.names` and `get_catalog_history_request.code` -> `.name` (matching the real primary-key column); updated the 5 Qt callers that used the old names or included the old combined protocol header.
- Found and fixed a real pre-existing bug in `CatalogDetailDialog`: `promptChangeReason()`'s result was never assigned to `catalog_.change_reason_code` before saving -- would have failed every catalog save's change-reason validation once this entity went live on the generated stack.
- Updated two test files for the new stateless repository shape and fixed one test assertion that was wrong by the codebase's own convention (expected a freshly-generated, unwritten catalog's `version == 1`; every domain_entity generator, confirmed against `country_generator.cpp`, sets `version = 0`, incremented to 1 only by the insert trigger on first write).
- Eventing and NATS event-relay wiring remain known, separately-tracked gaps (post-generation checklist), not attempted here. Stopping after catalog per the agreed scope -- the remaining four entities follow in later work.

## Test plan
- [x] `ores.dq.api.lib`, `ores.dq.core.lib`, `ores.qt.data_management.lib`, `ores.qt.exe`, `ores.dq.service.lib`, `ores.http.server.exe` all build clean
- [x] Full `ores.dq.api.tests` (267 assertions/46 cases) passes
- [x] Full `ores.dq.core.tests` (318 assertions/104 cases) passes, including catalog's own 9 test cases against a live DB


mcraveiro:
Fixed in cd978687d — thanks for the thorough diff-against-siblings check. (1) Added the missing version/Version row to catalog.org's Table display section and regenerated. (2) Added the missing #include <chrono> to catalog.org's Domain includes block and regenerated. (3) Updated the stale doxygen comment in data_organization_service.hpp. (4) Filed 5DF07DAD as a tracked follow-up for the change_reason_code pattern found in ~9 other dialogs, rather than fixing it in this PR.

PR #1694, opened 2026-07-25
[ores.codegen,ores.refdata] Add readonly-paginated-list Qt facet knob + QuantLib calendar materialisation

mcraveiro:
## Summary

Progress on materialising QuantLib calendar holidays and calendar adjustments: rule-engine domain, calendar_rule/calendar_exception/calendar_dates entities, calendar_adjustment export, and a new ores.codegen Qt facet knob (has_readonly_paginated_list) for junction-backed read-only paginated lists (first designed for calendar_dates).

## Changes

- Add calendar_rule/calendar_exception/calendar_dates refdata entities with materialisation and on-demand regenerate NATS command
- Add calendar_adjustment export service and NATS command
- Add has_readonly_paginated_list Qt codegen knob: strips add/edit/delete/history from a generated controller/MDI window, verified with compass codegen entity generate --diff against book and currency (zero attributable drift)
- Verification: local build clean (linux-clang-debug-make); ctest 73/74 test suites passed -- the one failure (ores.refdata.core.tests) is pre-existing DB schema drift on this environment (missing columns from an earlier migration, flagged stale before this work started), not a regression from these changes

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Materialise QuantLib calendar holidays and support calendar adjustments](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/calendar-followups/task_quantlib-calendar-transparency-and-adjustments.html) | 7FF8A057-F95D-4654-B709-13D74525892B |


mcraveiro:

mcraveiro:
Fixed in 010f3b08e:

1. Year-boundary bug — widened the year loop in instantiate_holidays_batch by one on each side (the per-date filters already clip to [start,end], so this is safe); added a regression test isolating the boundary via effective_from.
2. Unbounded end_year — clamped horizon_year to current_year+100 before the materialisation loop runs, covering both the request-supplied value and the system-setting fallback.
3. has_readonly_paginated_list gaps — gated updateActionStates()'s declaration to match its already-gated definition, and added an explicit default matching has_pagination's pattern. On the untested-positive-path point: smoke-tested by temporarily flipping the knob on currency (which has full messaging wired), regenerating, compiling ores.qt.refdata.lib clean, then reverting — calendar_dates itself still has no messaging/NATS layer generated, so it can't be the real consumer yet; that's tracked as a known follow-on on the (closed) codegen task.

Not changed (declined for this round): the four minor items in #4 (CSR grouping precondition, datetime .ok() validation, untested ruleset boundary years, hardcoded 100000 read limits, silent out-of-range fallback). You flagged these as low-priority/non-blocking; keeping this round scoped to the near-blocking items. Recorded in the task's Review table for follow-up if not picked up before the story closes.

mcraveiro:
Fixed in c7698e582:

- Critical soft-FK bug: root-caused in the codegen template (sql_schema_domain_entity_create.mustache hardcoded the joined column as `id`) rather than hand-patched in generated SQL. Added a `:referenced_column:` override to the Foreign keys drawer (defaults to `id`, unchanged for every other entity), set to `code` on calendar_rule and calendar_exception. Verified: regenerated both, triggers now check `code = NEW.calendar_code`; book/currency (existing entities using the default) diff clean.
- High: day_offset spinbox couldn't represent -2 (Good Friday) — added a per-field `:spin_min:` override column to the Detail fields table (also fixed org_loader dropping blank table cells so per-field overrides actually take effect instead of being shadowed by an empty string). day_offset now uses -9999 as its unset sentinel, leaving its real range fully representable.
- High: unvalidated date text crashing on Save — added a non-throwing `datetime::is_valid_iso8601_date()` and wired every `is_date` detail field into the generated validateInput() generically (not just calendar_exception). Also added the missing `.ok()` check to from_iso8601_date() itself.

Verified: targeted build of ores.platform.lib + ores.qt.refdata.lib clean; compass codegen entity generate --diff clean on currency (Qt) and book (SQL) -- no drift on entities using the pre-existing defaults.

mcraveiro:
Thanks for the thorough verification, including tracing the org_loader blank-cell fix's blast radius across the other ~20 entities' Detail fields tables. Leaving the cosmetic {{field}}_date_val naming nit as a follow-up (non-blocking, both reviews agree) rather than a merge-blocking fix. Proceeding to merge.

PR #1696, opened 2026-07-26
[agile] Split image-batch story into decision and implementation tasks

mcraveiro:
## Summary
- Doc-only: rework the task breakdown for "Fix image-batch NATS payload overflow, add SVG compression" -- the single bundled task conflated a design decision with two independent implementation concerns.
- Split into a decision task (NATS-byte-aware-batching+compression vs. HTTP storage-endpoint transport) plus two implementation tasks (compression + raw-byte wire encoding, byte-size-aware batching); old task marked ABANDONED with a pointer to the split.
- Recorded the NATS-vs-HTTP-transport leaning and the base64-in-JSON root cause (rfl::json wire format, not a NATS limitation) in the story.
- Filed the broader "compress all NATS JSON messages" question as a separate backlog capture rather than folding it into this story's scope.

## Test plan
- [x] Doc-only change, no code touched -- no build/test applicable.


mcraveiro:
@claude please review this PR

mcraveiro:
Fixed in 53730e358: bumped both #+updated: properties to 2026-07-26, added :nats:messaging:performance: filetags to the new capture (matching audit-codegen-facet-doc-coverage.org's convention), and filled in the * Goal sections for the decision and byte-size-aware-batching tasks so each is self-contained rather than depending on its description line. Thanks for the thorough check on the rfl::json/base64 claim.

mcraveiro:
Not changed. Agreed the * Acceptance section is still framed singular and is technically contingent on the decision task's outcome, but as you noted this PR is explicitly a mechanical split -- rewriting acceptance criteria now would mean guessing which of the decision task's possible outcomes to write it for. Leaving it to be updated once that task actually lands and the path is confirmed, per the story's own = Decisions section convention.

PR #1701, opened 2026-07-27
[ores.codegen,ores.refdata] Downgrade address/model-type incompatibility to skip in --component mode

mcraveiro:
## Summary

codegen.sh regenerate --component runs emitted one ERROR per entity whose model type isn't supported by the requested address (e.g. a junction model against ores.cpp.qt), inflating the run's error count for expected, harmless skips. Fixed by distinguishing a genuine type/address incompatibility from an ordinary empty-intersection outcome, and downgrading the former to a silent DEBUG skip in --component (auto-discovery) mode while keeping it a hard ERROR for explicit single-model invocations.

## Changes

- Added address_supports_model_type() to physical_space.py: true if any facet under an address admits a given model type, ignoring enablement.
- generate.py's _generate_single() takes a component_mode flag; type/address incompatibility is DEBUG-only + rc=0 in component mode, ERROR + rc=1 for explicit single-model calls. The pre-existing 'compatible type, nothing enabled' WARNING path is unchanged.
- 10 new unit tests (test_generate_component_mode.py + 4 cases in test_physical_space.py).
- Required precondition: migrated all 56 ores.refdata entity org models to the unified Columns/:primary_key:/:natural_key: shape -- every one of them was failing to load for any address due to another environment's now-closed story changing the parser without migrating the models; verified via dry-run (zero ERROR/WARNING across every address) since real regeneration would pull in unrelated stale-template diffs (filed as a separate capture).
- Filed 3 backlog captures for deferred follow-ups: stale generated code vs. current templates, ~75 lookup_entity models outside refdata needing the same key-shape migration, and the codegen meta-model knowledge doc.
- Verification: local build clean (linux-clang-debug-make); ctest 74/74 passed; ores.codegen suite 77/77 passed

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Downgrade address/model-type incompatibility to skip in --component mode](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/codegen_dx_improvements/task_downgrade_component_mode_type_incompatibility_to_skip.html) | 22055006-D8BE-4E85-8E04-3D9D1B77A6ED |


mcraveiro:

mcraveiro:
Addressed the review feedback in bf4ce738a:

1. **Redundant `load_graph()` reload on the empty-units skip path** — Fixed by caching `load_graph()` via `functools.lru_cache` keyed on `templates_dir`. Safe within a process invocation (template library doesn't change mid-run, returned `Graph` is never mutated by callers). Went with caching over threading the graph through `resolve_targets()`'s return value since that function has two external callers (`compass_codegen_entity.py`) unpacking its 3-tuple — changing the signature would ripple further than this fix needs to.
2. **Dropped `party.org` documentation paragraph** — Restored as body text under `* Columns` instead of silently discarding it.
3. **Behaviour change: explicit single-model calls now rc=1 (was rc=0) for a genuine type/address incompatibility** — Not changed; this is exactly the acceptance criterion this task implements ("keep ERROR for explicit single-entity invocations"), and no script in the repo relies on the old rc=0 for this path.

Also folded in: two more entity models (`calendar_exception`, `calendar_rule`) landed on `main` via a `pr sync` rebase after the original 56-file migration, still on the old key shape — migrated both to match.

Local re-verification: build clean, ctest 74/74, `ores.codegen` suite 77/77, dry-run zero ERROR/WARNING across every `refdata` address.

PR #1706, opened 2026-07-27
[nats] Add transparent transport-level gzip compression for image-batch payloads

mcraveiro:
## Summary

Fixes the image-batch NATS payload overflow by adding transparent, transport-level gzip compression to ores.nats, plus fixes an unrelated codegen bug found while verifying.

## Changes

- Add compress_if_worthwhile/decompress_if_flagged in ores.nats/domain/compression, wired into make_msg()/extract_message() -- every NATS message above 4096 bytes gets transparent compression, no protocol/domain type changes needed
- 7 new unit tests covering threshold gating, the compression-doesnt-help safety net, header precedence, and round-trip correctness
- Close decision and compression tasks; split raw-bytes-not-base64 into a separate follow-up task (not load-bearing, ~25 percent further improvement)
- Fix unrelated HTML-entity corruption bug in three SQL codegen templates (found while recreating the local DB to verify) that broke db recreate for anyone on current main

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Fix image-batch NATS payload overflow, add SVG compression](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/fix-image-batch-payload-and-compression/story.html) | EFFEEFAC-0406-4FF3-AAFB-075EB0324DB1 |
| Task | [Add gzip compression to the image NATS pipeline, send raw bytes instead of base64](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/fix-image-batch-payload-and-compression/task_compress-image-nats-payload.html) | BB6947FC-EE92-44E9-984A-BDFFC25A9BCA |

### DIFF
--- a/doc/agile/versions/v0/sprint_24/fix-image-batch-payload-and-compression/story.org
+++ b/doc/agile/versions/v0/sprint_24/fix-image-batch-payload-and-compression/story.org
@@ -80,9 +80,9 @@ if NATS-with-compression genuinely can't be made to fit.
 |---------------+----------------------------------------|
 | State         | STARTED                              |
 | Parent sprint | [[id:D269862E-2035-4857-A0D0-A6E93B0E24C9][Sprint 24]] |
-| Now           | Compression task started.              |
+| Now           | Compression done; batching remains.    |
 | Waiting on    | Nothing.                               |
-| Next          | Implement compression; then batching.  |
+| Next          | Start the byte-size-aware batching task. |
 | Last touched  | 2026-07-26                               |
 
 * Acceptance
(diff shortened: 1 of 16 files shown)


mcraveiro:

mcraveiro:
Fixed in 80a6a8d39. Moved extract_message(msg) inside on_msg's try block, with single-destroy handling of the natsMsg* across both the success and exception paths (needed since extract_message needs msg alive, but the pointer must not be double-freed if the handler itself later throws). Also added the suggested test: decompress_if_flagged now has a case asserting it throws std::runtime_error on a gzip-flagged, non-gzip payload, pinning the exact contract on_msg's try block depends on. request_sync's direct call to extract_message is unaffected since it's synchronous, not a callback. Full ctest 74/74 passing.

PR #1709, opened 2026-07-27
[docker] Service-runtime container: NATS sidecar, controller crash fix

mcraveiro:
## Summary

Builds out the Docker service-runtime deployment for the WSL-offload story: a NATS sidecar pod pattern, per-service NATS env stamping so the containerized controller can start from an env-file alone, and a real crash fix in ores.controller.service found while validating it end-to-end.

## Changes

- Add docker/nats.Dockerfile (scratch-based, vendored static nats-server binary) and wire it into docker/run-pod.sh/stop-pod.sh as a pod sidecar sharing the services container's network namespace.
- Stamp ORES_CONTROLLER_SERVICE_NATS_* env vars in env_init.py -- the controller is the only process a container entrypoint launches directly (it fans the rest out itself via CLI args), and it never had these stamped since compass_services.py normally passes them as explicit flags instead.
- Fix a real use-after-free race in application::run(): supervisor.start_all() was spawned detached, so a NATS connect() exception could destroy process_supervisor (joining its thread pool) while start_all() was still running on it, corrupting state mid-access -- confirmed via gdb backtrace, reproduces natively too. Now joined and awaited regardless of success/failure, with the original exception rethrown after cleanup so exit codes are unaffected.
- Stage NATS client TLS certs into a podman-managed volume rather than bind-mounting them directly -- bind-mounting breaks the TLS handshake for reasons isolated by elimination (not a path/permission/seccomp/capability/openssl.cnf issue; baking the same files into an image or a plain podman volume both work).
- File a follow-up story (split services into one container per service) and a follow-up task (debug: start_all() coroutine never resumes inside the container) for what's discovered but not yet resolved.
- Verification: local build clean (linux-clang-debug-make), ores.controller.{core,service,api}.tests build clean; ctest DB-connectivity failures are pre-existing environment noise (remote Postgres pg_hba.conf rejecting this host on port 5432), reproduced identically on main before this branch's changes.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Set up Docker for the service runtime deployment](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/offload-services-to-wsl-host/task_setup_docker_service_runtime.html) | EE315554-2A5B-4576-B25E-FE282E9B90CF |


mcraveiro:

mcraveiro:
Fixed in b34db4b30 — wrapped the start_all_task join in its own try/catch, folding into the existing failure exception_ptr only if not already set, so stop_all() always runs and the earlier failure (if any) is preserved. Also fixed the docker/.env sourcing nit you flagged (docker/run-pod.sh:21,27) — both scripts now source docker/.env explicitly instead of relying on it matching root .env by convention. Declined the nats_store_dir rw-mount nit as its own thing since it's now folded into fix #5 below (same PR, different comment thread) rather than duplicated here — see the stage-runtime.sh preset and image-tag-pinning notes: left as-is for this task, tracked as later hardening, not blocking.

mcraveiro:
Fixed in b34db4b30 — wrapped the start_all_task join in try/catch (folds into the existing failure exception_ptr only if unset) so stop_all() always runs and no failure is silently dropped. Also fixed the private-key exposure: chmod on the staged volume is now u+rwX,go-rwx instead of a+rX — the services container already runs as the same host uid that does the staging, so owner-only is sufficient.

mcraveiro:
Fixed in b34db4b30 — start_all_task's join is now wrapped in try/catch, folding into the existing failure exception_ptr only if unset, so stop_all() runs regardless and no earlier failure is silently replaced. On the cert-staging breadth: declined narrowing to per-container subsets for now — the controller genuinely needs every other service's cert path too (it derives each supervised service's --nats-tls-cert/key from that same directory when launching it, process_supervisor.cpp), so this is really the filed split-into-one-container-per-service story's job, not a local tweak here. Did tighten the chmod on the staged volume to owner-only (u+rwX,go-rwx) instead of world-readable, which addresses the actual private-key-exposure risk independent of that breadth question. Noted your point about stop_all() now also being blocked behind the in-container start_all() hang -- folded that into the scope of the already-filed follow-up task (67C78BEF) rather than duplicating it here, since it's the same root cause that task exists to investigate.

PR #1712, opened 2026-07-27
[assets,qt] Make ImageCache batching byte-size aware

mcraveiro:
## Summary

Backstop for the image-batch NATS payload overflow story: cut fetch batches by cumulative estimated payload size, not image count alone.

## Changes

- Add image_info.size_bytes (free server-side, data already loaded to build the list)
- ImageCache caches sizes in available_images_ across full and incremental loads
- fetchImagesInBatches cuts batches on cumulative estimated wire size (900KB threshold, unknown-size images get their own batch)
- MAX_IMAGES_PER_REQUEST repurposed as a generous sanity bound (200) rather than the primary defence

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Fix image-batch NATS payload overflow, add SVG compression](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/fix-image-batch-payload-and-compression/story.html) | EFFEEFAC-0406-4FF3-AAFB-075EB0324DB1 |
| Task | [Make ImageCache batching byte-size aware](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/fix-image-batch-payload-and-compression/task_byte-size-aware-batching.html) | DF92CC86-8A91-44FE-AA34-5CC78D1B2F12 |


mcraveiro:

mcraveiro:
Fixed in ca9512f27. loadAll()'s chain now defers its final loadImagesByIds(pending_image_ids_) call via finishLoadAllChain()/resumeDeferredLoadAllChain() when available_images_ isn't populated yet, instead of proceeding with every id treated as unknown-size. It triggers loadImageList() only if one isn't already in flight (avoiding a redundant load when MainWindow's own call wins the race), and resumes once onImageListLoaded() fires regardless of success/failure so nothing gets stuck. Declined the two minor findings (O(n*m) merge scan, no unit tests for fetchImagesInBatches) as non-blocking, matching your own framing.

mcraveiro:
Fixed in ca9512f27 -- same fix as the other review passes on this same finding (loadAll()/loadImageList() race). See the reply on comment 5097216955 for the full rundown.

mcraveiro:
Fixed in ca9512f27 -- same fix as the other review passes on this same finding (loadAll()/loadImageList() race). See the reply on comment 5097216955 for the full rundown.

mcraveiro:
Fixed in ca9512f27. loadAll()'s chain now defers its final fetch via finishLoadAllChain()/resumeDeferredLoadAllChain() until available_images_ is populated, rather than depending on incidental async ordering against loadImageList(). Full details in the reply on comment 5097216955.

mcraveiro:
Fixed in f5f44bcd1. Good catch -- finishLoadAllChain() now checks the client manager's connection state itself before deferring; when disconnected it fetches immediately (matching every call site's pre-existing behaviour before finishLoadAllChain existed), since deferring couldn't lead anywhere while disconnected anyway. Also updated the doc comment. Full ctest 74/74 passing after the fix.

PR #1718, opened 2026-07-28
[agile] Scaffold NATS configurable wire format story

mcraveiro:
## Summary
- Doc-only: scaffolds a new story splitting off from the image-batch story's raw-bytes-not-base64 follow-up task.
- While scoping that task, found reflect-cpp already supports MessagePack (rfl::msgpack), which natively packs std::vector<uint8_t> as binary with zero custom framing code -- solving the base64 problem generically for every NATS message type, not just images.
- New story: a single, non-polymorphic wire_codec chosen once at process startup from .env, threaded through ores.service/ores.qt/ores.shell via dependency injection. No per-message negotiation or runtime auto-detection.
- Five tasks: write-up/design, wire_codec foundation, and one migration task per consuming layer (server, Qt client, shell client), plus end-to-end verification.
- Marks the superseded raw-bytes-not-base64-for-images task ABANDONED with a pointer to the new story.

## Test plan
- [x] Doc-only change, no code touched -- no build/test applicable.


mcraveiro:

mcraveiro:
Fixed in 1e60af045. Removed the leftover empty * Acceptance placeholder (right after * Status) from all 6 task files, keeping the populated one after * Goal.

mcraveiro:
Fixed in 1e60af045 -- same fix as the other review pass on this same finding. Confirmed all 6 files (including task_e2e-verify-msgpack-wire-format.org) now have exactly one * Acceptance section each.

mcraveiro:
Fixed in 8775c144c. Moved * Acceptance to its canonical position after * Status in all 6 files, matching the template and task_raw-bytes-not-base64-for-images.org's order (Goal, Status, Acceptance, Plan, ...). Also cleaned up a stray double-blank-line left by the move.

PR #1719, opened 2026-07-28
[codegen] Add has_parent_scoped_list Qt knob for parent-scoped lists

mcraveiro:
## Summary

Adds a has_parent_scoped_list knob (paired with parent_key_field/parent_key_param) to the ores.cpp.qt facet, so a has_readonly_paginated_list list window's get-request can be scoped to an owning parent key and opened for a specific parent, with that key sent on every paginated fetch.

## Changes

- The generated Client model class gains a conditional constructor parameter, stores it, and sends it on the paginated get-request every fetch alongside offset/limit; the generated MdiWindow class threads the same parameter into the client model it constructs; the generated Controller class gains openForParent(const QString&) mirroring the existing openAddWithParent pattern, while showListWindow() still satisfies EntityController's pure-virtual zero-arg override.
- Kept distinctly named from the pre-existing parent_entity_singular/has_parent_relationship mechanism, which solves a different problem (pre-filling a foreign key on create for a still-full-CRUD entity) -- documented the distinction in entity_meta_model_cpp_qt.org.
- Wiring calendar_dates itself onto this knob is the follow-on task (Generate calendar_dates Qt facet and wire Browse Holidays action); this PR is templates-only.
- Verification: local build clean (linux-clang-debug-make, all targets including ores.qt.exe); ctest ores.qt.headless.tests and ores.qt.tests, 2 of 2 passed; --diff against book/currency with the new knob unset shows zero drift attributable to this change; a temporary, reverted smoke test on currency.org with the knob set produced valid, correctly-gated generated C++.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Codegen: design and implement Qt parent-scoping for has_readonly_paginated_list](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/calendar-followups/task_codegen-qt-parent-scoped-list.html) | 1C673C53-9ED0-4C02-B856-EF031A1CC95F |


mcraveiro:

mcraveiro:
Fixed in 20eb7598a — guarded the destroyed-cleanup lambda to only null listWindow_/listMdiSubWindow_ when they still reference the specific subwindow being destroyed, so closing an older parent-scoped window can no longer orphan a still-open, newer one. Also added validate_parent_scoped_list() (raises ValueError, following the existing validate_explorer_interface/validate_cached_by pattern) so a missing parent_key_field/parent_key_param now fails fast at generation time instead of rendering broken C++. Full multi-window reload/staleness tracking across concurrently-open parent-scoped windows is a separate, larger change to EntityController's single-listWindow() contract -- documented as a known limitation in a code comment on reloadListWindow() rather than fixed here, since it would touch every generated entity, not just this knob.

mcraveiro:
Fixed in 20eb7598a — same fix as the sibling review comment: guarded the destroyed-cleanup lambda (only nulls the pointers when they still reference the destroyed subwindow) and added validate_parent_scoped_list() for the missing-companion-fields case. The multi-window reload/staleness-marking limitation you also called out is real and left as a documented, separately-scoped follow-up (touches EntityController's base contract, not just this knob's templates).

mcraveiro:
Fixed in 20eb7598a — guarded the destroyed-cleanup lambda so it only nulls listWindow_/listMdiSubWindow_ when they still reference the subwindow actually being destroyed, preventing a newer parent-scoped window from being orphaned when an older one closes. Kept the deeper multi-window reload/stale-marking limitation you identified as a documented, out-of-scope follow-up (it needs a wider EntityController change, not something this knob's templates alone should take on). Also picked up the parent_key_param_ initializer-list nit and added generation-time validation for the required companion fields.

PR #1721, opened 2026-07-28
[marketdata] Add oresmd parsing/projection/resolution support

mcraveiro:
## Summary

Implements the oresmd design inside ores.marketdata: five per-asset-class identifier/requirement structs (std::variant, no base class/virtual dispatch), parse_oresmd/to_uri (Boost.URL), resolve(), and projection rules into ORE's index name, curve key, and quote key. No consumer migration in this change.

## Changes

- Data model in ores.marketdata.api/domain: oresmd_enums, market_data_identifier/market_data_requirement (5 asset classes each), oresmd_uri
- oresmd_parser (parse/to_uri), oresmd_projections (index name/curve key/quote key), oresmd_resolver in ores.marketdata.core, using boost::urls and magic_enum
- Explicit per-asset-class query-key validation -- an invalid field (e.g. tenor on an FX URI) throws rather than being silently ignored
- 77-test-case suite covering every worked example in the design doc plus invalid-input cases; surfaced and fixed two real bugs (wrong instrument_type gate on index-name/curve-key projections; empty-string-as-present bug in resolver fallback)
- Verification: local build clean; ores.marketdata.core.tests 77/77 passed, ores.marketdata.api.tests and ores.marketdata.client.tests passed, ores.marketdata.service.exe builds clean

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Migrate to oresmd, delete market_series qualifier and ore_key](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/migrate-to-oresmd-delete-legacy-schemes/story.html) | CA11A754-C153-42A7-AF8B-FA8447CE50B1 |
| Task | [Add oresmd support to ores.marketdata](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/migrate-to-oresmd-delete-legacy-schemes/task_oresmd-parsing-projection-library.html) | 657F1AF9-0B49-40DA-9F2D-68006B53F814 |


mcraveiro:

mcraveiro:
Fixed in cbaa14c17: query_params::from now throws oresmd_exception on any query key outside the seven known ones, so a typo like tenr=3m fails loudly instead of being silently dropped. Also added parse_fx entity validation (must be a 6-letter alphabetic pair) so a malformed URI like oresmd://fx/eur fails at parse time rather than only downstream in projections.

mcraveiro:
Fixed in cbaa14c17: extracted the duplicated to_upper/to_lower into a shared oresmd/detail/oresmd_string_utils.hpp, and the duplicated is_overnight classification into oresmd/detail/oresmd_index_family_utils.hpp, used by both oresmd_parser and oresmd_projections. Also added validation in parse_ir: a term index (libor/euribor) type=fixing without a tenor now throws, matching the metric-without-type=quote check already in place. Reworded oresmd_resolver's docstring to describe the actual safe, throwing mismatched-defaults behaviour rather than implying undefined behaviour. 3 new test cases added, 80/80 passing.

PR #1724, opened 2026-07-28
[ores.controller] Fix start_all() never resuming (co_spawn use_awaitable hazard)

mcraveiro:
## Summary

Fixes a real, cross-environment service-startup bug found while validating the Docker service-runtime deployment end-to-end and independently confirmed by bright faraday hitting the identical symptom natively (no containers): process_supervisor::start_all() never gets its spawned coroutine's first resume scheduled because it was spawned via co_spawn(..., use_awaitable) without being awaited immediately, and the enclosing coroutine goes on to await the long-lived main service loop. Confirmed by direct experiment across dozens of test runs and hundreds of JWKS retries / hours of runtime that swapping to boost::asio::detached makes the exact same coroutine body enter immediately and reliably every time.

## Changes

- Fix application::run(): spawn start_all() detached instead of holding an unawaited use_awaitable task; signal completion via a steady_timer moved into the past (the standard Asio one-shot-event idiom) so the crash-safety property from the earlier use-after-free fix (always wait for start_all() to finish before touching process_supervisor) is preserved without the broken pattern.
- Fix a real console-logging bug found while debugging: the console sink never called auto_flush(true) (unlike the file sink, which already does), so console output sat buffered under non-interactive capture (podman logs, redirected stdout) -- looked exactly like a stall when it was just buffering.
- docker/run-pod.sh: set --userns=keep-id on the pod (at pod-create time -- podman rejects it on a container joining an existing pod). Rootless podman remaps container uids through a user namespace by default, so the services container's uid did not actually match the host uid that staged/chmod'd the NATS certs volume, causing a permission-denied reading them.
- Note left for next session, not yet independently verified: do_launch()'s co_await boost::asio::post(bind_executor(db_pool_.get_executor(), use_awaitable)) is structurally similar but immediately awaited (unlike the fixed case), which should make it safe -- promoted to trace-level logging so it's easy to check if service startup is ever observed stalling there specifically.
- Verification: local build clean (linux-clang-debug-make); ores.controller.{core,service,api}.tests build clean; ctest DB-connectivity failures are pre-existing environment noise (remote Postgres pg_hba.conf rejecting this host on port 5432), reproduced identically before this branch's changes. Native runs confirmed start_all() now enters reliably and completes end-to-end (DB writes with fresh timestamps, IAM launches) across many repeated test runs.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Debug: start_all() coroutine never resumes inside podman container](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/offload-services-to-wsl-host/task_debug-start-all-never-resumes-in-container.html) | 67C78BEF-C664-4B57-AE4D-CBCF3782FBF0 |


mcraveiro:

mcraveiro:
Fixed in 89903d453 -- do_launch() now posts back onto ioc_'s executor before returning (mirroring monitor_process()'s existing pattern), so start_all()'s coroutine chain and the completion signal in application::run() always resume on io_ctx, never on db_pool_'s thread.

mcraveiro:
Fixed in 89903d453 -- same fix as the completion-signal race noted above (do_launch() now posts back onto ioc_ before returning). Also addressed the catch(...) and operation_aborted-check minor notes.

mcraveiro:
Fixed in 89903d453 -- do_launch() posts back onto ioc_ before returning, addressing the race. Also added catch(...) alongside catch(const std::exception&), and the async_wait catch now rethrows anything other than operation_aborted instead of swallowing all system_error.

mcraveiro:
Fixed in 89903d453 -- do_launch() now hops back onto ioc_'s executor before returning (co_await post(bind_executor(ioc_.get_executor(), use_awaitable))), mirroring monitor_process(). This keeps start_all()'s whole coroutine chain, and application::run()'s completion signal, resuming on io_ctx rather than db_pool_'s thread. The NOTE comment at process_supervisor.cpp is updated to reflect this.

mcraveiro:
@claude please re-review this pull request

PR #1727, opened 2026-07-28
[synthetic,devops] Merge asset-class folders into one theme unit; migrate build storage to btrfs

mcraveiro:
Two strands of work on this branch.

## 1. Synthetic theme as one atomic dataset

Merges the per-asset-class theme datasets into a single atomic dataset, with
the accompanying SQL populate changes and DQ/registrar updates. Design note
and agile task records included.

## 2. Migrate shared ccache/build storage to btrfs

Story `03831B51`. `ccache`'s `hard_link=true`, combined with a `cache_dir`
shared across concurrent worktrees, is documented as unsafe: two worktrees
compiling the same TU can corrupt the shared inode. That was root-caused via
GDB after a heap-corruption crash in `ores.refdata.service`.

ext4 has no reflink support, so the safe alternative (`file_clone=true`)
silently degraded to plain copies. `/mnt/development` is now btrfs.

**btrfs over XFS.** Both support reflinks and both fix the bug; XFS is
otherwise the better fit for this workload. btrfs wins on one decisive point:
`file_clone=true` **force-disables ccache's own compression**, so the cache
store is stored raw. btrfs transparent zstd gives that back — measured **6.6G
of uncompressed cache entries occupying 1.3G on disk**.

### `compass` change

`env configure` defaulted `ORES_COMPILER_CACHE` to `sccache` whenever `.env`
had no existing value. `.env` is gitignored, so every worktree's
`ORES_COMPILER_CACHE=ccache` was lost with the reformatted partition and fresh
worktrees silently fell through to sccache — bypassing the migration entirely.
Default is now `ccache`. sccache remains selectable; CI is unaffected as it
sets `SCCACHE_DIR` directly and never runs `env configure`.

Also drops the stale `SCCACHE_DIR`/`SCCACHE_CACHE_SIZE` block from the
generated `.env` — version 10 recorded these as removed but the template kept
writing them — and fixes a comment pointing at a script that does not exist.
Bumps `.env` format to **version 15**.

### Validation

| Claim | Evidence |
|---|---|
| No shared-inode hazard | 3074/3074 and 3078/3078 objects at link count 1; none above |
| Reflink, not copy | `logical-resolve`: one extent, two files, independent inodes |
| Cache correctness | Full rebuild from cache: 2382/2382 hits, 0 errors |
| Cross-worktree sharing | 72.5% hits in a second worktree on a different branch |
| Compression | ccache store 6.6G logical → 1.3G on disk (5.1x) |
| Net effect | Two built worktrees + 9.6G cache: 38.9G logical → **13G on disk** |

Two fully built worktrees plus the shared cache now occupy less disk than one
worktree did before.

> [!NOTE]
> `ORES_ENV_VERSION` is bumped to 15, so existing checkouts must re-run
> `compass env configure --preset <preset> -y` before `db recreate`.


mcraveiro:
Addressed findings 1 and 2 in `e8ea50ff7`.

**1. `replace_all` unconditionally voiding both asset classes** — confirmed and fixed. Each void block is now gated by the same `if exists (select 1 from ..._artefact_tbl where dataset_id = p_dataset_id)` predicate as its corresponding population loop, so the delete path honours the same contract.

Verified against a live database (`db recreate`): the function installs cleanly and the gates are present in the installed body. Worth noting the change is **zero-regression for current data** — all three themes carry both asset classes, so both gates evaluate true and behaviour is byte-identical to before:

| dataset | fx artefacts | ir artefacts |
|---|---|---|
| `synthetic.themes.uniform_demo` | 11 | 20 |
| `synthetic.themes.ore_samples_2016` | 13 | 4 |
| `synthetic.themes.realistic_2026` | 13 | 20 |

That is also the limitation: I could not exercise the single-asset-class path behaviourally, because no such dataset exists to publish. The fix only alters a path that cannot currently be reached — which is precisely why it was worth fixing before someone creates that dataset.

**2. Stale dataset code in debug message** — fixed; `uniform_demo`'s FX populate now says `synthetic.themes.uniform_demo`, matching the lookup above it and its two sibling files.

**3. Header-comment drift** — left as a follow-up cleanup pass, as suggested.

**4. Two table names in `artefact_table`** — agreed it is descriptive-only; no change.

PR #1735, opened 2026-07-29
[qt] Extend SyntheticBindingDialog to cover IR curve configs

mcraveiro:
## Summary

SyntheticBindingDialog only ever listed fx_spot_generation_config records and bulk-created feed bindings for them, so IR curve feed ticks could never be persisted into market_observations even manually. Generalizes the dialog to cover IR curves too, bringing them to FX's existing quality bar.

## Changes

- loadConfigs() now fires both get_fx_spot_generation_configs_request and get_ir_curve_generation_configs_request, merging both into one local BindableConfig{kind, ore_key, source_name} list
- Adds a Type (FX/IR) column so mixed rows are unambiguous
- IR curve ore_key is computed client-side (currency_code + strip_currency_prefix(currency_code, index_name)) mirroring ir_curve_feed.cpp's own market_series qualifier lookup
- createBindings() is otherwise kind-agnostic already, no change needed there
- Manually verified: binding and starting an IR curve feed persists ticks into ores_marketdata_market_observations_tbl the same way FX already does; pre-ticking works for both kinds

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [IR Rates synthetic data: dataset seeding, index cleanup, dual-curve, quoting conventions](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/ir-rates-followups/story.html) | C29FE5A3-61F0-4D2F-BE4F-8A02223EABEF |
| Task | [Extend SyntheticBindingDialog to cover IR curve configs](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/ir-rates-followups/task_extend-synthetic-binding-dialog-to-ir-curves.html) | 48D6933D-05F4-4B34-B796-FE1B4A2A6444 |


mcraveiro:

mcraveiro:
Fixed in d8c4a5d83 — loadConfigs() now falls back to FX-only rows (with a warning log) if the IR service is unavailable, instead of failing the dialog closed. Added a comment explaining the IR ore_key's deliberate qualifier-only shape and why it's harmless today. The FX Spot grid regression itself is intentionally not fixed here — it's tracked as a separate follow-up task (task_filter-feed-bindings-by-product-kind.org), called out explicitly in this task's Result section rather than left implicit.

mcraveiro:
Fixed in d8c4a5d83 — addressed all three findings: (1) FX Spot regression acknowledged as a known, deliberate scope call with the follow-up task linked from this task's Result; (2) added a comment on the IR ore_key qualifier-only shape explaining why it's safe today and pointing at the follow-up task for the real fix; (3) loadConfigs() no longer fails closed on IR service unavailability — falls back to FX-only rows with a warning log.

mcraveiro:
Thanks for confirming the strip_currency_prefix()/ore_key derivation checks out. In d8c4a5d83 I also picked up the resilience point from the parallel review pass — loadConfigs() no longer fails closed if the IR service is unavailable — and added a comment documenting the deliberate IR ore_key shape.

PR #1740, opened 2026-07-29
[marketdata,qt,codegen] Filter FX Spot grid by product kind (feed_binding.asset_class)

mcraveiro:
## Summary

SyntheticBindingDialog can now bind IR curve configs (previous PR #1735), but feed_binding had no product-kind discriminator, so FxSpotGridWindow rendered every enabled binding as an FX currency pair -- IR bindings showed up as garbled rows. Adds feed_binding.asset_class, reusing the existing ores::marketdata::domain::asset_class taxonomy already used by market_series, and filters FxSpotGridWindow to FX-only.

## Changes

- feed_binding.asset_class added via the codegen entity model, reusing domain::asset_class (soft FK to refdata.asset_class_code) rather than inventing a separate value set
- SyntheticBindingDialog sets asset_class at bind time; FxSpotGridWindow filters get_feed_bindings_request results to asset_class == fx
- Audited every other consumer of get_feed_bindings_request: FeedBindingMdiWindow (generic admin list) unaffected by design, CrmCrossRatesMatrixMdiWindow doesn't touch feed_bindings at all
- Fixed a codegen bug (org_loader.py): default_value on an is_enum column was SQL-string-quoted for the C++ template too, producing invalid C++; now emits the correct scoped-enum expression
- Migrated feed_binding.org to the unified Columns key-modeling format (prerequisite: org_loader.py no longer parses the legacy split for domain_entity models) -- filed the underlying regression as its own task
- Fixed IR ore_key: was missing its RATES/YIELD prefix (bare qualifier only, e.g. AUD/AONIA instead of RATES/YIELD/AUD/AONIA), traced against ir_curve_feed.cpp's own find_series call and real ORE key conventions in external/ore/examples
- Added a selectable ORE key label to IrCurveEditor, mirroring FxSpotRateEditor's existing one, so binding no longer requires working the key out by hand
- Manually verified end-to-end via QA scenario: bind FX + IR feeds, both tick, FX Spot grid shows only the FX row, IR tick still persists to market_observations despite being filtered from the display

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [IR Rates synthetic data: dataset seeding, index cleanup, dual-curve, quoting conventions](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/ir-rates-followups/story.html) | C29FE5A3-61F0-4D2F-BE4F-8A02223EABEF |
| Task | [Feed bindings have no product-kind filter: FX Spot grid renders IR bindings as garbage currency-pair rows](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/ir-rates-followups/task_filter-feed-bindings-by-product-kind.html) | C2F23164-99A2-4640-8DA2-8BFB017ACB12 |


mcraveiro:

mcraveiro:
Fixed in d04ecc8c4 — restored the 'Bind synthetic' toolbar action/slot in FeedBindingMdiWindow.hpp/.cpp exactly as it was before the codegen regeneration accidentally dropped it, and added comments on both sides flagging it as hand-maintained code with no codegen extension point, so it survives (by being re-added, not forgotten) after any future 'compass codegen entity generate feed_binding'. Also removed the now-dead FeedBindingHistoryDialog files per your other note.

mcraveiro:
Fixed in d04ecc8c4 — confirmed the root cause was exactly commit 138c476a's regeneration (feed_binding org-model format migration) silently dropping the hand-maintained toolbar action. Restored it, with comments explaining why it needs to be re-added after future regenerations. Also removed the dead FeedBindingHistoryDialog files and added the org_loader.py is_enum default_value unit test you suggested (test_org_loader_keys.py, two new cases: enum default becomes a scoped enum expression, non-enum text default still SQL-quoted). Left the other three minor notes (tooltip branch, unused lambda captures, unused include) as-is since they are all in codegen-generated files/templates, out of scope for this PR.

mcraveiro:
Fixed in d04ecc8c4 — restored the toolbar action/slot rather than routing through FeedBindingController, since it constructs SyntheticBindingDialog directly with no server round-trip needed first (unlike onAddNewRequested/showAddWindow's flow) -- kept the original hand-maintained pattern rather than introducing a new one. Also removed the dead FeedBindingHistoryDialog files and left a comment on the generator's version=0 vs sibling's version=1 divergence would be worth aligning, but did not change it in this round since it is genuinely harmless (the insert trigger overwrites version unconditionally) and touching generated code isn't this PR's job -- happy to align it if you'd still like that as a follow-up.

mcraveiro:
Correction to my earlier reply: the first fix hand-edited FeedBindingMdiWindow/FeedBindingController directly, which is wrong for this project (generated files are never hand-edited -- there's a real extension mechanism, paste blocks, for exactly this). Reworked properly in 713215bf3: the toolbar action/signal/controller-relay are now paste blocks in ores.marketdata.feed_binding.org (mirroring ores.refdata.party's own cross-navigation seam), so they regenerate correctly on every future codegen run instead of needing to be re-added by hand. SyntheticBindingDialog is now constructed in MarketdataPlugin.cpp (the hand-written plugin composition root), and simplified to fetch its own already-bound source names instead of requiring a model snapshot from the caller.

PR #1747, opened 2026-07-29
[docker] Deploy service runtime to Newton: fixes, partial success, corruption flagged

mcraveiro:
## Summary

Deploys the containerized service runtime to the actual Newton WSL host (192.168.1.22, WSL2 Debian 11 bullseye) for the first time -- prior PRs (#1709, #1724) only validated the pod on this dev box's own podman. Reaches partial success and fixes several real deploy-blocking bugs found along the way; does **not** reach the task's full acceptance (all 18 services running, Qt client connecting) -- blocked on an apparent memory-corruption issue in `process_supervisor` surfacing under Newton's real load, deliberately not chased down here (see below).

## Changes

- `docker/service-runtime.Dockerfile`: bake in writable `/app/log` and `/app/run`. `process_supervisor` always launches every child service with `--log-enabled` and a relative `--log-directory ../log`, and writes PID files under `../run` -- both resolve under `/app`, and neither directory existed at all in the chainguard `glibc-dynamic` final image, so every child launch failed writing its log/PID on a real deploy regardless of uid. Created (world-writable, since the runtime uid is only known at `podman run --user` time) in the debian `strip` build stage, which has a shell, and `COPY`'d into the final stage, which doesn't.
- `.gitignore`: ignore `docker/.env.*` -- a Newton-specific env file with secrets was created this session as a `docker/.env` variant, not previously covered by the existing `docker/.env` rule.
- Reopens the "Offload service and DB runtime to a WSL host over SSH" story (it had been closed prematurely -- prior work never actually deployed to Newton itself), adds and closes the "Deploy and verify service runtime on the Newton WSL host" task with a Result documenting what was found/fixed, and files two follow-up tasks: compass tooling for remote deploys, and support for multiple named `.env` files.

## What was fixed this session (documented in the task's Notes)

- Newton had no podman at all -- installed, enabled lingering, and fixed a one-time `dbus.socket` activation gap so podman uses the `systemd` cgroup manager instead of falling back to `cgroupfs`.
- Networking model: Newton's Postgres is native on the same host the pod runs on, so the existing pod/sidecar netns pattern (which assumes Postgres is remote from both containers) can't reach it -- switched to `--network=host` for both containers instead.
- DB role passwords on Newton's Postgres were stale relative to this environment's current `.env` -- resynced via `compass db recreate` run against Newton (temporarily editing `ORES_DB_HOST`/`PGPORT` in the local `.env`, then restoring it -- fragile, hence the follow-up task).
- Cert-permission/uid mismatch -- fixed with an explicit `--user` matching the host uid that staged the NATS certs.
- The `/app/log`/`/app/run` fix in this diff.

## What's not fixed / carried forward

After all of the above, `start_all()` progressed noticeably (from 0 services ever reaching `running` to a handful) but was unstable across runs and never reached all 18. One run showed corrupted binary bytes in a DB query string where a service name should have been, immediately followed by a permission-denied re-reading a cert that had already been read successfully earlier in the same run -- looks like a real memory-safety issue in `process_supervisor`'s child-process bookkeeping under real load. This lives in the single-container process-supervisor architecture that the sibling "Split ORE Studio services into one container per service" story is about to replace, so it's flagged there (in that story's Decisions) rather than debugged in code on its way out.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Deploy and verify service runtime on the Newton WSL host](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/offload-services-to-wsl-host/task_deploy-verify-newton-wsl-host.html) | 3627D272-CC9F-4A45-B5F4-0B5A5B1E71AE |


mcraveiro:

mcraveiro:
Fixed in 662708e64 -- removed the redundant docker/.env.* line (and .env.old, also covered by .env.*), and updated the story's Now/Next to reflect the closed task and the two open follow-ups. Left the chmod 777 as-is per your own note that it's a reasonable tradeoff given the uid-unknown-at-build-time constraint.

PR #1752, opened 2026-07-29
[docker] Per-service container images: design + IAM pilot, fix silent NATS cert-load errors

mcraveiro:
## Summary

First task of the new "Split ORE Studio services into one container per service" story: writes up the design for moving `process_supervisor`'s responsibilities (startup ordering, readiness gating, process spawn/monitor, restart policy, DB-backed audit trail) to the orchestration layer, and pilots the container side of it on IAM.

Along the way, ships real, working infrastructure (per-service image builds) and fixes a real, previously-silent bug in `ores.nats` (unchecked NATS TLS cert-load errors) discovered while root-causing what initially looked like a second instance of the Newton memory-corruption bug from PR #1747 -- it wasn't; see below.

## Changes

- `docker/stage-runtime.sh`: adds `--service <binary-name>` to stage just one service binary plus its own `ldd`-computed dependency closure (a single `ldd` call resolves the full transitive closure already, no recursion needed), instead of always staging all 18. Default (all-18) behaviour unchanged.
- `docker/service-runtime.Dockerfile`: adds `ARG SERVICE_NAME`, selecting the entrypoint via a relative symlink created in the debian `strip` stage (the final chainguard base has no shell to `ln -s` with itself; a relative target resolves correctly once `COPY`'d into the final stage, unlike an initial attempt with an absolute cross-stage target that buildah's copier rejected as dangling). Replaces an earlier "one shared 775MB image + `--entrypoint` override" approach for per-service images, rejected in review as not idiomatic -- every service would ship every other service's dead code.
- `ores.nats/service/client.cpp`: `natsOptions_LoadCATrustedCertificates`/`LoadCertificatesChain`'s return status was never checked, so a failed cert/key load surfaced only as a generic, much-later "Connection Closed" instead of a clear diagnostic. Now throws immediately, naming the exact file and explaining the likely cause.
- Agile docs: closes the design/pilot task with a Result documenting what was proven (per-service image works; IAM connects to NATS standalone once `--userns=keep-id` is used; the Newton corruption finding does *not* reproduce here) and what's deferred (a distinct JWT-signing failure, the `HEALTHCHECK` probe); files a follow-up task for that remaining work.

## The corruption red herring, and the real root cause

Piloting IAM standalone in a per-service container initially reproduced a "client didn't provide a certificate" NATS server error, closely resembling the unexplained Newton corruption finding from PR #1747. Root-caused by instrumenting `client.cpp` to actually check the ignored return codes: rootless podman, without `--userns=keep-id`, remaps container UIDs through `/etc/subuid` rather than mapping 1:1 to the invoking host user. `--user 1000:1000` therefore does not mean "the same UID as the host" -- reading a host-owned, mode-600 NATS client private key then fails with a plain `EACCES`, which OpenSSL reports as a generic SSL error that our code silently ignored. This is the *exact same* root cause already fixed for Newton in `docker/run-pod.sh` (which already uses `--userns=keep-id`) -- it just hadn't been connected to this pilot. Confirmed it applies uniformly (both IAM and the controller need it), and confirmed the actual Newton corruption finding does not reproduce in this single-container pilot, as expected given there's no in-process multi-child bookkeeping left to race.

## Verification

- Full local build clean (`linux-clang-debug-make`).
- `ores.nats.tests`: 26 test cases, 51 assertions, all pass.
- Native `compass services start`/`stop`: unaffected, no regression.
- IAM-only image built (410MB vs. 2.1GB for the all-18 image) and run standalone via podman with `--userns=keep-id`: connects to NATS successfully and progresses into real application logic (not yet fully healthy -- see follow-up task).

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Split ORE Studio services into one container per service](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/split-services-into-per-service-containers/story.html) | FFC4117A-3D79-4779-A347-92532ED2A19A |
| Task | [Design the per-service container architecture and pilot it with one service](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/split-services-into-per-service-containers/task_design-pilot-per-service-container.html) | 9E08056D-DEC1-43E3-803E-03E5F9808AD5 |

### DIFF
--- a/doc/agile/versions/v0/sprint_24/split-services-into-per-service-containers/story.org
+++ b/doc/agile/versions/v0/sprint_24/split-services-into-per-service-containers/story.org
@@ -7,7 +7,7 @@
 #+level: s2
 #+filetags: :docker:architecture:deployment:sprint_24:v0:
 #+created: 2026-07-27
-#+updated: 2026-07-27
+#+updated: 2026-07-29
 #+environment: swift_curie
 #+todo: BACKLOG STARTED | DONE ABANDONED
 
@@ -42,7 +42,7 @@ other binaries.
 | Now           | Design done, IAM pilot partially proven (connects to NATS standalone); a distinct JWT-signing bug and the HEALTHCHECK probe remain. |
 | Waiting on    | Nothing.                               |
 | Next          | Finish the IAM pilot (JWT bug + healthcheck), then the dependency generator and events watcher. |
-| Last touched  | 2026-07-27                               |
+| Last touched  | 2026-07-29                               |
 
 * Acceptance
 
--- a/doc/agile/versions/v0/sprint_24/split-services-into-per-service-containers/story.org
+++ b/doc/agile/versions/v0/sprint_24/split-services-into-per-service-containers/story.org
@@ -8,7 +8,7 @@
 #+filetags: :docker:architecture:deployment:sprint_24:v0:
 #+created: 2026-07-27
 #+updated: 2026-07-27
-#+environment:
+#+environment: swift_curie
 #+todo: BACKLOG STARTED | DONE ABANDONED
 
 This page documents a [[id:699A8D45-B67E-4D3E-9206-24FA17C51ADA][story]] in [[id:D269862E-2035-4857-A0D0-A6E93B0E24C9][Sprint 24]]. It captures the goal, current status, acceptance criteria, and the tasks that compose it.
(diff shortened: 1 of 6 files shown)


mcraveiro:

mcraveiro:
Fixed in 2612b38df -- --service now uses a clear ${2:?...} error, and story.org's dates are bumped. Left the check_file leak and missing test coverage as-is per your own note that they're pre-existing/out of scope for this fix's stated goal.

mcraveiro:
Fixed in 2612b38df -- --service error message clarified, story.org dates bumped. The check_file leak and missing test coverage noted as follow-up, not addressed here.

mcraveiro:
Fixed in 2612b38df -- added a RUN test -f guard in the Dockerfile before creating the entrypoint symlink (verified it fails the build loudly on a deliberate SERVICE_NAME/staged-binary mismatch), and clarified the --service error message. Left the check_file leak and missing test coverage as follow-up items.

PR #1754, opened 2026-07-29
[synthetic] Migrate ir_curve_generation_config onto oresmd, add curve role

mcraveiro:
## Summary

Replaces index_name (free-text full floating_index_type code) with oresmd-shaped index_family/tenor fields, adds role, and rebuilds curve_feed_controller's collision check so a discount curve and a projection curve for the same currency+index+tenor can coexist. Also fixes two unrelated bugs surfaced along the way: a codegen required-static_combo misclassification, and a macOS-only RTTI mismatch in oresmd_exception.

## Changes

- Migrated the model to the unified Columns + :primary_key:/:natural_key: codegen format (required, discovered mid-task) and regenerated the full entity stack (SQL, domain, repository, Qt)
- index_family: static_combo (fixed 6-value vocabulary); tenor: dynamic_combo over refdata::domain::tenor via a new fetch_tenors; role: static_combo (discount/projection/self_discounting)
- (currency_code, index_family) composite soft FK validated in the C++ service layer (ir_curve_generation_config_service), since the codegen Insert-trigger mechanism can't express a two-column lookup
- curve_feed_controller's collision check now keys on (qualifier, role) via a new pure ir_curve_feeds_conflict predicate; covered by new curve_feed_controller_tests.cpp
- Migrated every index_name consumer: ir_curve_feed.cpp, IrCurveEditor.cpp, IrCurveGenerationConfigHistoryDialog.cpp, MarketSimulatorWindow.cpp, SyntheticBindingDialog.cpp, application.cpp
- Fixed a codegen bug: required static_combo fields generated QComboBox text() calls instead of currentIndex checks
- Fixed a macOS-only CI regression: oresmd_exception missing ORES_MARKETDATA_CORE_EXPORT caused an RTTI mismatch across the shared-library boundary on Mach-O
- Verification: local build clean (linux-clang-debug-make); ctest 74/74 suites (78 test cases) passed

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Migrate to oresmd, delete market_series qualifier and ore_key](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/migrate-to-oresmd-delete-legacy-schemes/story.html) | CA11A754-C153-42A7-AF8B-FA8447CE50B1 |
| Task | [Migrate ir_curve_generation_config onto oresmd, add curve role](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/migrate-to-oresmd-delete-legacy-schemes/task_migrate-ir-curve-generation-config.html) | 9DB7786E-C9BE-467F-A2B9-5A9AFCFD6B45 |


mcraveiro:

mcraveiro:
Fixed in b346730a4 — tenorCombo now uses currentData() (the code) instead of currentText() (the display_name), matching indexFamilyCombo/roleCombo on the same lines.

mcraveiro:
Fixed in b346730a4 — same tenorCombo currentData()/currentText() fix as the other pass, plus: added a blank entry to the tenor combo and a new updateTenorComboForIndexFamily() that disables/blanks it for overnight families (sofr/estr/sonia/tona), gated on indexFamilyCombo's currentIndexChanged and called from updateUiFromConfig()/setReadOnly() too, so saving an overnight-family config no longer sends a non-empty tenor and violates the CHECK constraint.

mcraveiro:
Fixed in b346730a4 — all four items addressed: (1) added a roleCombo_ to IrCurveEditor's Instrument tab, preselected from ir_.role and wired into the save path, so curves created via MarketSimulatorWindow can set role — this was indeed a gap, not deliberate scoping; (2) fixed the generator to use valid fixed values (sofr/self_discounting/empty tenor) instead of suffixing index_family/role with the uniqueness idx, since party_id/config_id already guarantee row uniqueness; (3) fixed the doc-comment typo in the model source (regenerated the domain header); (4) removed the dead crReasonCode/crCommentary lambda capture.

PR #1412, opened 2026-07-03
[marketdata] Fix uninitialized asset_class/series_subclass in market_series generator

mcraveiro:
## Summary

Nightly Valgrind (linux-gcc-debug-ninja) flagged Uninitialized Memory Conditional/Read defects in market_series_mapper::map(); traced to generate_synthetic_market_series() never setting the two enum fields.

## Changes

- generate_synthetic_market_series() now cycles asset_class and series_subclass deterministically through their full enumerator sets, keyed off the existing per-call index counter
- Verified 0 valgrind errors on ores.marketdata.core.tests [repository] with the project's own build/valgrind/custom.supp suppressions and CTest.cmake memcheck flags

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Fix uninitialized asset_class/series_subclass in market_series generator](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/fix_market_series_generator_uninitialized_enums/story.html) | 83B12C2D-5202-4738-8147-5F28E566D997 |
| Task | [Set asset_class/series_subclass in generate_synthetic_market_series](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_22/fix_market_series_generator_uninitialized_enums/task_fix_market_series_generator_uninitialized_enums.html) | B6BAAABC-C602-4413-A223-9AD532384934 |

**Environment:** solid_dirac

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1764, opened 2026-07-30
[marketdata] Fix marketdata entity model drift

mcraveiro:
## Summary

Regenerates every ores.marketdata domain_entity against current codegen templates; 3 of 4 were stuck on the same legacy primary/natural-key org convention already fixed for ores.synthetic.

## Changes

- Migrated market_fixing/market_observation/market_series off the legacy top-level Primary key/Natural keys headings onto current :primary_key:/:natural_key: field flags.
- feed_binding already regenerated clean (zero diff).
- Verification: local build clean (linux-clang-debug-make); full rat 0 failures.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Fix marketdata entity model drift: update to latest format and regenerate](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/codegen_dx_improvements/task_fix-marketdata-model-drift.html) | 121F5727-E69C-4FA6-B8AD-40F286C22CD2 |


mcraveiro:

mcraveiro:
Excellent catch — both real bugs, fixed in 65ee06b63:

1. Restored the asset_class/series_subclass generator assignments (deterministic cycling, keyed off idx, matching PR #1412's fix exactly) — but this time captured in the .org model's generator blocks rather than only hand-written into the .cpp, so a future regen can't silently drop it again.
2. Qualified the log_handler_entry call in market_series_handler.hpp's export_to_storage paste-block (ores::service::messaging::log_handler_entry) — it was broken by the unrelated handler_helpers-to-wire_codec migration already on main; this was the first time anything actually compiled this header since that landed.

Verified: build clean, full rat 0 failures.

mcraveiro:
Went further than the single-instance fix: added a validate_model check (6c69432b2) that errors whenever a non-nullable `is_enum` column has neither `:default_value:` nor `:default:`. The prior fix only added a `:name generator:` block, which feeds the synthetic-data generator but not the domain struct itself — `cpp_domain_type_class.hpp.mustache` only emits a member initializer when `:default_value:` is set, so the struct member was still genuinely uninitialized on default construction independent of the generator. Added `:default_value: fx`/`spot` to market_series's asset_class/series_subclass so the struct itself is now safe too. Checked every `is_enum` column in the repo (only market_series + the already-correct feed_binding) — no other entity affected, but this closes the whole class of bug for anything added later. Verified: build clean, full rat 0 failures.

mcraveiro:
Both gaps fixed in c6d24ccd9:

1. Dropped the `or "default" in col` branch — the validator now requires `:default_value:` specifically, matching what the template actually consumes. Re-verified: a column with only `:default:` now correctly fails validation.
2. Now also scans `de["primary_key"]["columns"]`, closing the primary-key gap.

Re-ran both accept/reject repros to confirm the tightened logic; `market_series` (already using `:default_value:` correctly) unaffected. Build clean, full rat 0 failures.

PR #1767, opened 2026-07-30
[agile] Merge badge-adoption-audit and classification stories

mcraveiro:
## Summary
- Merges two separately-filed stories that turned out to be the same underlying question: which entities should render as badges, images, or plain text, and rolling out badge_key adoption without deciding that first.
- Retitled and broadened "Audit and roll out badge adoption across domain entities" to cover badge vs image vs plain-text classification, reusing its existing audit-then-implement two-task shape.
- Abandoned "Classify FK-like/lookup entities..." as superseded, with cross-reference (marked ABANDONED, not deleted).
- Broadened both tasks: task 1 is audit/classify, task 2 is batched implementation.

## Test plan
- Docs-only change; no build/test needed.


mcraveiro:

mcraveiro:
Added the audit/classification analysis (task 712AA40D) to this same PR rather than opening a new one -- see the task's updated Plan/Result for the full breakdown: 18 new badge classifications (batched), 18 explicit plain-text decisions, ~65 entities out of scope.

mcraveiro:
Fixed in 31261d8ab -- all three issues addressed: (1) updated all 6 stale-link occurrences (including the two pre-existing ones flagged as lower priority) to the retitled story name; (2) currency_pair legitimately has both flag icons (base/quote combos) and a badge (classification field) -- clarified in the audit doc that the exactly-one-treatment policy is per-field, not per-entity, rather than removing either classification; (3) corrected the 108 to 120 entity count in story.org acceptance and the task description. Also fixed a drift-check CI failure from this same push: I had edited the tangled .mustache templates directly instead of their org tangle sources -- moved the wireDetailDialogCommon and paste-seam edits into ores.cpp.qt.controller_header.org, controller_impl.org, and detail_dialog_header.org, then re-tangled, and verified byte-identical .mustache output.

mcraveiro:
Fixed in 5715e1737 -- updated the sprint_23 occurrence too.

PR #1777, opened 2026-07-30
[dq] Fix dq entity model drift

mcraveiro:
## Summary

Regenerate every ores.dq domain_entity/junction against current codegen templates, and fix several codegen infrastructure gaps (junction paste-blocks, Qt-drawer fail-fast, composite-PK handler args, junction generator uniqueness) surfaced along the way.

## Changes

- Regenerated all ores.dq domain_entity/junction models (badge_severity, catalog, change_reason, change_reason_category, code_domain, data_domain, dataset_bundle, subject_area, badge_mapping, dataset_bundle_member)
- Added junction paste-block support (paste marker syntax) to nats-handler/protocol templates; restored dataset_bundle_member's hand-written list() operation via a paste block, migrated by_bundle onto the existing generic list_by mechanism
- Fixed root-cause class of bug: junction/domain_entity models with no Qt drawer used to abort the whole generate run or silently emit broken Qt output instead of failing/skipping cleanly
- Fixed composite text primary keys being truncated to their first column in generated NATS handler history/remove calls
- Added counter/idx uniqueness to junction generators (previously only domain_entity had it), fixing a real duplicate-key collision
- Fixed stale ores.dq include paths, invalid Qt Icon enum values, a change_reason_fn name collision, and a dataset_bundle seed upsert ON CONFLICT mismatch
- Verification: local build clean (linux-clang-debug-make); full rat (build + ctest) passed after db recreate

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Fix dq entity model drift: update to latest format and regenerate](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/codegen_dx_improvements/task_fix-dq-model-drift.html) | B8FE7DA1-18D7-4F33-8402-21411A51A8E2 |


mcraveiro:

mcraveiro:
Fixed in a1c52579a — added a needs_counter flag for junctions (mirroring domain_entity's), computed from whether any left/right/columns generator_expr references idx, and gated the counter/idx declaration behind it. Confirmed badge_mapping no longer emits it, and dataset_bundle_member (which does use idx) still does. Also added projects/ores.codegen/tests/test_dq_drift_fixes.py covering the composite-PK extra-args fix and this counter/idx gating, per the request for regression coverage in this same PR.

mcraveiro:
Fixed in a1c52579a — added projects/ores.codegen/tests/test_dq_drift_fixes.py with tests for the composite-text-PK history_request_extra_args/delete_request_extra_args fix and the junction counter/idx needs_counter gating (round-1 medium finding). Didn't add a dedicated test for the Qt fail-fast skip logic or paste-block substitution given time constraints, but the two added tests cover the two fixes most at risk of silent regression.

mcraveiro:
Finding 1 (unwired badge_mapping messaging stack): fixed in a1c52579a — disabled ores.cpp.nats-handler/nats-sub-registrar/protocol/service for badge_mapping via :ores.*.enabled: false overrides (it already routes through the bespoke badge_handler/badge_service per registrar.cpp's own comment) and removed the dead generated files.

Finding 2 (junction boolean :default: not rendering): not changed — this is pre-existing (also affects refdata's calendar_date before this PR), and every construction site already sets the field explicitly, so it's latent rather than a live bug. Out of scope for this drift-fix task; worth a follow-up template fix on its own.

Finding 3 (NATS subject/field rename, informational): not changed — no action needed, producer and consumer are updated together in this PR.

PR #1778, opened 2026-07-30
[codegen,refdata] Fix calendar_rule/calendar_exception soft-FK check that failed every INSERT

mcraveiro:
## Summary

calendar_rule and calendar_exception's soft-FK check against calendars.code actually checked a nonexistent id column, because their .org models used :referenced_column:, a property name the codegen template never reads (only :target_column: is consumed) -- a dead-key trap in org_loader.py. Every calendar_rule/calendar_exception INSERT was failing outright. Reported independently by another environment via a from-clean db recreate reproduction.

## Changes

- Removed the dead, unread referenced_column key from org_loader.py so no other entity can be silently misconfigured against it
- Corrected calendar_rule.org/calendar_exception.org to the real target_column property and regenerated their SQL facet
- Verification: local build clean linux-clang-debug-make; ctest 72 of 74 passed (excluding the two pre-existing, already-captured ores.iam failures), verified against a freshly-recreated database via compass db recreate, not a stale local schema

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Fix calendar_rule/calendar_exception soft-FK check: wrong property name silently no-ops the code override](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/calendar-followups/task_fix-calendar-rule-exception-soft-fk-target-column.html) | DDA34695-4AB2-4791-874C-90A3F904A6F1 |


mcraveiro:

mcraveiro:
Fixed in a88f77b4 -- added test_org_loader_foreign_keys.py covering target_column defaulting, explicit override, and an explicit assertion that referenced_column is not a recognised override. All 93 ores.codegen pytest cases pass.

mcraveiro:
Fixed in a88f77b4 -- same fix as the other reviewers' identical finding. Added test_org_loader_foreign_keys.py.

mcraveiro:
Fixed in a88f77b4 -- same fix as the other reviewers' identical finding. Added test_org_loader_foreign_keys.py.

PR #1779, opened 2026-07-30
[docker] Get all 18 services running end-to-end on Newton

mcraveiro:
## Summary

Reaches full success where PR #1747 closed with a partial-success caveat: all 18 services (22 rows counting `ores.compute.wrapper`'s 5 replicas) reach `running` on Newton, reproducibly across 3+ restarts. The apparent `process_supervisor` corruption bug that caveat was attributed to was **not** a distinct bug -- it resolved once the `--userns=keep-id` uid-remap fix (PR #1752) and several further real bugs found in this session were fixed. The Acme end-to-end smoke test scenario passed in full against this deployment, including currency CRUD/eventing/history and a compute job run to completion.

## Bugs found and fixed

- **Newton-specific storage `EIO`** writing to the image's own baked-in `/app/log`, `/app/run`: fixed by host-bind-mounting those paths (plus `/app/storage`, needed once `ores.http.server` required it) instead of relying on the image's own overlay copy.
- **DB connection exhaustion**: repeated redeploys leaked connections faster than Newton's default `max_connections=100` could absorb; bumped to 800, stale connections terminated once.
- **Stale DB schema** on Newton (`marketdata.feed_bindings.asset_class` missing): refreshed via a full `compass db recreate` against Newton.
- **`docker/stage-runtime.sh`'s default glob** (`ores.*.service`) silently never staged `ores.http.server` or `ores.compute.wrapper` -- both now explicitly included.
- **`ores.http.server` advertising the wrong URL**: defaults `--http-base-url` to `http://localhost:{port}` when unset, and it was never set in its DB-driven `args_template` -- remote clients got told to reach the HTTP server on their own `localhost`. Fixed by setting it explicitly in that `args_template` row (deployment-specific DB data, not a code fix).
- **`podman --env-file` doesn't strip quotes** the way bash's `source` does, corrupting the JWT private key's quoted, embedded-newline `.env` line. Fixed `docker/generate-env.sh` to strip quotes generically, and to pass this one variable via `--env "$(cat ...)"` instead of the env-file at all, since podman's env-file format can't hold a multi-line value regardless of quoting (mirrors what `compass_services.py`'s `child_env()` already does for the native launch path).
- **`compass nats certs --hostname`** added an IP address as a `DNS:` SAN entry instead of `IP Address:`, failing strict TLS hostname verification for any client connecting by IP. Fixed `nats_certs.py` to detect an IP-shaped hostname and emit the correct SAN type.
- **Compute dispatch tenant mismatch**: `process_supervisor.cpp` substituted `{tenant_id}` with the NATS `subject_prefix` string (not a UUID) when launching `ores.compute.wrapper`, so its NATS subscription for dispatched work (`compute.v1.work.assignments.{tenant_id}.{platform}`) never matched what the server actually publishes to. Root-caused: the whole compute domain (hosts, batches, workunits) is consistently created under the system tenant by design -- now passes `tenant_id::system()` explicitly.

## Deliberately not fixed here -- filed as captures

- **Compute repositories should query the system tenant, not the caller's** -- even with the dispatch fix above, `host_repository.cpp` (and likely `batch_repository.cpp`/`workunit_repository.cpp`) filter reads/writes by the caller's own session tenant rather than the system tenant the data actually lives under, so a real business-tenant login sees 0 compute hosts/tasks (only `super_admin` happens to work). Bigger fix across 3 repositories' read+write paths, explicitly scoped out of this task.
- **Vendored ORE engine package missing its own shared libraries** -- `external/ore/packages/ore-1.8.15.0-x64-linux.tar.gz` never bundled `libOREAnalytics.so`/`libQuantExt.so`/`libQuantLib.so.1`, so a dispatched compute job's engine process fails at startup. Confirmed reproducible natively (not container/Newton-specific) -- a packaging gap upstream of this repo.

## Verification

- Full local build clean (`linux-clang-debug-make`), including after rebasing onto latest `main`.
- All 18 services (22 rows) confirmed `running` via direct DB query, reproducibly across 3+ consecutive container restarts on Newton.
- Acme end-to-end smoke test scenario passed in full (currency CRUD/eventing/history + a compute job submitted, dispatched, and run to completion) via the local Qt client connected to Newton's remote NATS/DB.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Get all 18 services running end-to-end on Newton](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/offload-services-to-wsl-host/task_finish-all-18-services-on-newton.html) | 82F1F41E-3BA4-45EA-AABE-30CA969E2A1A |


mcraveiro:

mcraveiro:
Fixed in 4a63a1e — added an --env passthrough for ORES_IAM_SERVICE_JWT_PRIVATE_KEY in docker/run-pod.sh, reading the real PEM file (mirrors compass_services.py's child_env() for the native path). Also reworded the stage-runtime.sh comment and added a warning on a missing extra binary.

mcraveiro:
Fixed in 4a63a1e — same fix: docker/run-pod.sh now passes ORES_IAM_SERVICE_JWT_PRIVATE_KEY via --env with the real PEM contents (real newlines), overriding the escaped-\n value from --env-file docker/.env. Good catch on the description/diff mismatch.

PR #1773, opened 2026-07-30
Tidy up connections.db, decommission Barclays/NatWest, fix Acme bundle publishing

mcraveiro:
## Summary

- Bookkeeping: closes out several completed Acme Corporation holding-group tasks now that PR #1758 merged (org-chart and market-feed tasks remain open, tracked separately).
- Tidies the Qt client's saved `connections.db`: adds a `description` field, curates ~10 Acme connections (department heads + representative sample) tagged per entity, removes stale Barclays/NatWest connections, and decommissions Barclays as the default demo entity across skills, recipes, and `compass.py`'s scaffold hints (in favour of Acme Corporation).
- Fixes a real, pre-existing bug: bundle publish never respected the `optional` dataset-member flag — `opted_in_datasets` acted as a strict whitelist over the whole bundle instead of an additive opt-in on top of always-included members. This silently starved both Acme's and Barclays' `base` bundle publish of badges/fpml codes/countries/calendars whenever any dataset was opted in.
- Gives Acme full `base`-bundle reference-data parity (badges, fpml.* codes, countries, currencies, calendars, GLEIF) and publishes CRM Cross-Rates Matrix topology data per office.
- Adds a "Before you start" section to the `test_scenario` codegen template so system-readiness prerequisites are separated from tester-facing steps.
- Files a new task tracking a pre-existing, unrelated bug in `ores_synthetic_publish_theme_from_dq_fn` (owned by Bright Faraday) and works around it in the shared party-provisioning plan so provisioning no longer logs a hard failure for it.

## Test plan

- [x] Full rebuild (`compass build`) succeeds
- [x] `compass db recreate -y -k` + Acme holding-group provisioning script completes with zero failed steps
- [x] Verified via SQL: badges (62 rows), CRM topology configs (3 offices × 3 topologies), login_info rows (59/59 accounts) all present for the Acme tenant
- [x] QA Validation Runner scenario "Verify connections database tidy-up and Barclays decommission" — PASSED
- [ ] CI (drift/roundtrip/site checks)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1771, opened 2026-07-30
[synthetic,refdata] Fix index_name/index_family regression across IR curve themes

mcraveiro:
## Summary
Restores end-to-end provisioning of every IR curve theme (`ore_samples_2016`, `realistic_2026`, `uniform_demo`), broken by `df398d5f1`'s `index_name` → `index_family`/`tenor`/`role` migration — which touched the target table and every C++/Qt consumer, but never the DQ publish pipeline that actually writes rows.

## Changes
- Added `index_family`/`tenor` to both DQ artefact tables (replacing `index_name`); fixed all `index_name` references (plus the previously-missing `role` column) in `ores_synthetic_publish_theme_from_dq_fn`.
- Preserved full 20-currency `realistic_2026`/`uniform_demo` coverage (not trimmed) by extending `index_family`'s check constraint via the codegen `.org` model (SQL-only regen, avoiding pre-existing undocumented hand-patches elsewhere in the generated files) and sourcing real market conventions from ORE's own vendored QuantLib/QuantExt index definitions.
- Added `refdata_overnight_index_conventions_populate.sql` and `refdata_ibor_index_conventions_populate.sql` (24 rows total), wired into `foundation_populate.sql`. Caught and fixed a pre-existing seed-data typo: `NZD-NZIONA` should be `NZD-NZONIA`.
- Filed 2 follow-up captures for out-of-scope issues found along the way: the Qt create-dialog's static combo not yet listing the new families, and an unrelated `calendar_rules` regression.
- Verification: local build clean (linux-clang-debug-make); ctest 74/74 passed; full `barclays_system_provision.ores` run plus direct SQL invocation of all 3 IR curve themes (33/17/31 rows inserted, zero errors).

Story-ID: C29FE5A3-61F0-4D2F-BE4F-8A02223EABEF
Task-ID: C80D94BB-DA80-4871-AAF0-D4F852F58831

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1781, opened 2026-07-30
Restore synthetic themes, decouple CRM from vintage bundles

mcraveiro:
## Summary

- The root cause blocking `synthetic.themes.realistic_2026`/`synthetic.themes.ore_samples_2016` (`ores_synthetic_publish_theme_from_dq_fn`: "column existing.index_name does not exist") is now fixed on `main` (commit `6d3b76ef0`, PR #1771) -- reverts the workaround from PR #1773 that had dropped `synthetic_realistic_2026` from the shared `party_provisioning_bundle_plan()`.
- Goes further than a plain revert: also adds `synthetic_ore_samples_2016` to the same shared plan, so a freshly provisioned party can test either current-regime (2026) or legacy ORE Samples (2016) pricing scenarios without re-provisioning, once feed-starting and workspace-scoped pricing land.
- Splits `refdata.crm_topology_bundles` out of both vintage bundles into its own standalone, vintage-independent `crm_topology` bundle -- the CRM Cross-Rates Matrix should always be available regardless of which synthetic-data theme/vintage a party publishes.
- All additions are to the existing, generic `party_provisioning_bundle_plan()` shared by both Acme's provisioner and `ores.shell`'s generic "provision party" flow -- no Acme-specific C++ added.
- Along the way, discovered and filed (not fixed here) a real gap: there is no in-session party switcher (`PartyPickerDialog` is defined but never instantiated anywhere) -- an account's party context comes only from which specific account you log in as.

## Test plan

- [x] Full rebuild succeeds
- [x] `compass db recreate -y -k` + Acme holding-group provisioning completes with zero failed steps -- `synthetic_realistic_2026`, `synthetic_ore_samples_2016`, and `crm_topology` all show `completed` for every office
- [x] Verified via SQL: `refdata_crm_topology_configs` populated for all 3 offices
- [x] QA Validation Runner scenario "Verify IR, FX, and CRM synthetic data for Acme" -- all 4 steps PASSED (FX spot grid populated, IR curve chart populated, CRM Cross-Rates Matrix renders for majors/exotics/scandies)


mcraveiro:
Thanks for confirming. The pre-existing note about ores_dq_dataset_bundle_members_upsert_fn having no delete/retire path is correct and out of scope here -- as you say, this is a pre-production seed-data codebase where compass db recreate -y -k is the standard way to apply seed-data changes, not incremental migration. No action needed.

PR #1789, opened 2026-07-31
[iam,qt] Add in-session party switching

mcraveiro:
## Summary
- Adds a "Switch Party..." action (File > Current User) that reopens `PartyPickerDialog` mid-session for accounts with cross-entity access to more than one party.
- Backed by a new `switch_party` NATS endpoint (`iam.v1.accounts.switch-party`): `select_party`'s handler is deliberately gated on a single-use, login-only token audience, so switching mid-session needed its own endpoint rather than relaxing that check.
- Fixes UI feedback found during manual QA: the switch confirmation used the wrong status-bar accessor, and the persistent party chip wasn't refreshed on switch (only at login).

## Test plan
- [x] `compass build` — clean
- [x] `compass db recreate -y -k` + `provision tenant --source acme` — all steps completed, including Step 7 (cross-entity access)
- [x] `scenario_verify_switch_party.org` — all 4 steps PASSED via QA Validation Runner


mcraveiro:
Fixed in 81d4cc765 — added the same post-switch checks the login flow already had: showPartyProvisioningWizard() when lastPartySetupRequired(), otherwise MessageBoxHelper::warning() when lastPartySetupWarning() is non-empty.

PR #1800, opened 2026-08-01
[compute] Design and build ACME compute package publishing (client lib, shell, Qt)

mcraveiro:
## Summary

New ores.compute.client package_publisher shared by ores.shell's compute publish-package command and ores.qt's AppProvisionerWizard/UploadEnginesDialog: resolves platform, computes/verifies SHA256, uploads to the canonical per-platform storage key, then upserts app/app_version/app_version_platform. Vendors the ORE binary into publish/vendor-packages/ at build time. Also fixes 8 real bugs surfaced by manually testing the full flow end-to-end (shell publish, wizard, standalone dialog, and an actual dispatched compute job) against a real build rather than assuming correctness.

## Changes

- New ores.compute.client component: package_publisher (upload + client/server SHA256 verification)
- AppProvisionerWizard::PackageUploadPage retargeted to one upload per selected platform via the shared publisher
- New ores.shell compute publish-package command; new ores.qt UploadEnginesDialog for publishing against an existing app/version
- Fix: save_app_version_request replaces every platform row wholesale -- both new callers now fetch-then-merge instead of silently wiping out other platforms on republish
- Fix: ores.http.server (1MB default body_limit) and ores.compute.wrapper's own http_client (same bug on the response side) both rejected the ~57MB ORE package; raised/streamed properly in both
- Fix: ComputeConsoleWindow's detail dialogs now dock into the main window's MDI area via DetachableMdiSubWindow instead of popping up detached
- Fix: AppVersionDetailDialog's platforms/packages silently failed to load (setVersion() called before setCreateMode(false)); AppProvisionerWizard's Audit page queried the wrong change-reason category; both new Qt call sites used the raw login string instead of the resolved account username
- Fix: QaValidationRunnerWidget's screenshot capture now resolves the active window at capture time instead of always grabbing the MainWindow
- Verification: local build clean (linux-clang-debug-make, full clean rebuild); ctest 72/74 clean pass + the other 2 independently re-verified (ores.ore.core.tests fails identically on main, unrelated XML schema issue; ores.analytics.service.tests passes in isolation, was a concurrent-build false negative); manual QA test scenario run to PASSED across shell CLI, wizard, and standalone dialog; also manually submitted and ran a real compute job end-to-end against the published ORE package

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Update vendored ORE Engine to 1.8.16.0](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/update-vendored-ore-engine/story.html) | 9468ACFD-8A99-430E-8E8E-6A828EF48C04 |
| Task | [Design and build ACME compute package publishing (client lib, shell, Qt)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/update-vendored-ore-engine/task_design-acme-compute-package-publishing-flow.html) | 0F768B80-184A-44C7-B725-A594A873AB4A |


mcraveiro:

mcraveiro:
Fixed in fbb5ef660 — existing_platforms() now returns std::optional; both callers (compute_commands.cpp, UploadEnginesDialog.cpp) abort the publish rather than silently proceed with an empty platform list on fetch failure. Include-order nit also fixed. See the task's * Review table for how the remaining findings from this and later review passes were triaged.

mcraveiro:
SHA256-mismatch orphan object and unsynchronized read-modify-write are both real gaps I'm declining to fix in this PR (recorded with rationale in the task's * Review table) -- the former needs a DELETE-on-mismatch or temp-key-rename change to storage_routes, the latter is an admin-only low-frequency operation. Agreed the vendor-package tests gap is worth a follow-up rather than blocking here, given the manual E2E QA already recorded.

mcraveiro:
Findings 1-4 (UI-thread blocking, CMake glob timing, upload-before-DB-registration, fetch-then-merge race) triaged in the task's * Review table -- fetch-then-merge's fetch-failure case is fixed in fbb5ef660 (distinct from the race concern, which I'm declining as admin-only/low-frequency, matching your own conclusion). UI-thread blocking on large uploads is real but scoped as a follow-up rather than expanding this PR with a QtConcurrent refactor. max_body_size scoping and no-tests-for-ores.compute.client both declined for now, the former already covered by the existing stream_large_http_uploads_to_disk capture.

mcraveiro:
Confirmed -- the four still-open findings from the earlier passes are unchanged by this push and are now recorded with explicit accept/decline decisions in the task's * Review table (0F768B80). The top finding across all four passes (fetch-then-merge not guarding against fetch failure) is fixed in fbb5ef660.

mcraveiro:
Thanks for confirming. This round is complete -- all findings from the four review passes are either fixed (fbb5ef660) or recorded with rationale in the task's * Review table.

PR #1808, opened 2026-08-02
[refdata,codegen] Bind ores.refdata entities to profiles; verify zero-diff regen

mcraveiro:
## Summary

Implements real :profile: binding in the code generator for both domain_entity and junction metatypes, rebinds every matching ores.refdata entity/junction (100% coverage, 7 profiles), and proves it with a zero-diff regen against the pre-rebind checked-in output. Escalated mid-session into a full zero-drift hotfix: brought party's previously-BLOCKED SQL layer fully under codegen, fixed a case-sensitive paste-marker matching bug in the codegen substitution mechanism, fixed a composite-component scaffold bug, fixed a NATS wire_codec template bug, and eliminated all remaining ores.refdata generation drift. Rebased onto latest main and root-caused/fixed two post-rebase test failures rather than dismissing them as flaky.

## Changes

- org_loader.py resolves :profile: bindings for domain_entity and junction models from the profile's own variability profile org file Assignments table
- Added tenant-scoped-junction as a 7th profile; 100% profile coverage across ores.refdata's 67 codegen-eligible entities/junctions
- Brought party's SQL layer (codename generation/immutability/queue provisioning, sequence lifecycle) fully under codegen via two new template paste points
- Fixed case-sensitive paste-marker UUID matching in the codegen paste-substitution mechanism (core.py) -- found via full compass db recreate -y, not isolated diffing
- Fixed composite-component archetype matching (aggregator components like ores.refdata no longer incorrectly scaffold as flat single-module)
- Fixed NATS event-cache templates to use ores::nats::default_wire_codec() instead of raw rfl::json
- Eliminated all remaining ores.refdata generation drift (~130 regenerated files, byte-identical to current templates)
- Fixed ores.qt.headless.tests: pinned test binary's default_wire_codec to json to match json-encoded fixtures
- Fixed ores.iam.core.tests: mint_token() now stamps issued_at/expires_at immediately before signing, not before the DB round-trip, closing a short-ttl expiry race
- Verification: local build clean (linux-clang-debug-make); validate_schemas.sh green; ctest 74/74 passed; ores.iam.core.tests additionally stress-verified with 5 consecutive runs

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Entity classification and drift baseline across all components](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/entity-classification-drift-baseline/story.html) | 623CFC6A-320A-4F71-A595-B3A47FBD06B8 |
| Task | [Bind ores.refdata entities to profiles; verify zero-diff regen](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_24/entity-classification-drift-baseline/task_pilot-refdata-classification.html) | AD953FB1-65AF-4475-BF67-E964DA6CA5D4 |


mcraveiro:

mcraveiro:
Fixed in 6f44cc3b4 -- removed the stale, contradictory validation_ignore.txt entry. Also added a new CI job (refdata-codegen-drift, 4a1bba138) that regenerates ores.refdata and fails on any diff, at the user's request during this review round.

mcraveiro:
Fixed in 6f44cc3b4 -- removed the duplicate/contradictory validation_ignore.txt entry, keeping only the correctly-explained one. The legacy is_domain_entity_model fallback note: not changed, agreed it's a future cleanup item not a defect here.

mcraveiro:
Fixed in 6f44cc3b4 -- removed the duplicate validation_ignore.txt entry. The impls_lower collision note: not changed, agreed it can't happen today since UUIDs are canonically case-insensitive, so two such keys are genuinely the same paste point.

mcraveiro:
Round 2 review confirmed both fixes are correct and found no new blocking issues — noted only a purely-cosmetic CI-latency nit (refdata-codegen-drift's redundant `needs: classify` edge) which isn't worth its own CI cycle. All checks green (including the new refdata-codegen-drift job). Merging.

PR #1819, opened 2026-08-04
[ores.utility,ores.nats] Add generic shared-domain fallback tier to environment_mapper_factory

mcraveiro:
## Summary

A new pure-passthrough NATS knob (ORES_NATS_WIRE_FORMAT) required manual argv threading through five independent places to reach every service. This adds a generic, domain-agnostic fallback tier to environment_mapper_factory, so any config module can register a shared domain once and have every service pick up its options straight from .env, with zero argv threading. NATS registers as the first domain.

## Changes

- Added ores::utility::program_options::shared_domain_registry: register_domain(prefix)/domains(), Meyer's-singleton-backed, populated via explicit bootstrap calls (no static-init-order dependency).
- Extended environment_mapper_factory::make_mapper with a fallback tier: the per-app ORES_APP_* prefix still wins, then falls back to any registered shared domain generically (prefix-strip-and-kebab-case) -- no domain name appears in this file.
- Added nats_configuration::register_shared_domain(), wired into all 20 service parsers that build NATS options (17 via a scripted edit, 3 handled individually for their differing local variable names).
- Added unit tests for both the registry and the mapper's two-tier resolution (precedence, generic fallback, unregistered-domain safety, a hypothetical second domain).
- Regenerated component_files.cmake via regenerate_cmake_component_files.py --all.
- Verification: local compass build clean (linux-clang-debug-make, all targets); ctest -R ores.utility.tests and ores.nats.tests both passed (ores.utility.tests 138 cases, 471 assertions).

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Collapse NATS config passthrough to a single environment-mapper fallback tier](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/process-supervisor-config-passthrough-gap/story.html) | 9E3EC9E2-742F-4830-839D-99EF2C2BC626 |
| Task | [Add a generic shared-domain fallback tier to environment_mapper_factory](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/process-supervisor-config-passthrough-gap/task_add_shared_nats_env_fallback_tier.html) | A091A4B9-E54D-4D4F-8F7E-8C6D2437B660 |


mcraveiro:

mcraveiro:
Confirmed and fixed in d5f9bf4a1. You were right that duplicate occurrence was a real risk for ores.shell -- I removed the register_shared_domain() call from shell's parser.cpp entirely, since shell already has its own full ORES_SHELL_NATS_* mirror block and doesn't need the shared fallback. While reproducing this I also found and fixed a second, more severe bug the duplicate-occurrence scenario didn't quite capture: the shared-domain fallback matched any ORES_DOMAIN_* variable (whatever the domain), including ones that aren't options at all (e.g. ORES_NATS_PORT, ORES_NATS_MONITOR_PORT), which made parse_environment reject them as unrecognised options -- breaking every service that registered a domain, verified by running the rebuilt binaries directly. shared_domain_registry now stores exact allowed suffixes per domain instead of a blanket prefix. Added a regression test in ores.nats exercising the real parse_environment+store path end to end.

mcraveiro:
Thanks for the thorough pass -- your prefix-boundary check was correct for the original design, but a related issue surfaced when I reproduced the duplicate-occurrence bug flagged in the parallel review: the shared-domain matching needed to move from a blanket prefix to explicit allowed suffixes per domain (fixed in d5f9bf4a1). See the reply on that thread for details.

mcraveiro:
Noted -- the codegen template gap is intentional sequencing, tracked in task_consolidate_domain_service_parser_boilerplate.org as you found. Also fixed a real bug this round (see reply on the parallel review thread): the shared-domain fallback needed to scope to explicit allowed suffixes per domain, not a blanket prefix, and preserve the domain name in the derived option name. Local build + ctest -R for ores.utility.tests/ores.nats.tests verified green after the fix (476/141 and 53/27 respectively), and I additionally ran every rebuilt service binary directly (env-only, no CLI flags) to confirm the fix holds fleet-wide.

mcraveiro:
@claude please re-review — pushed a fix for the duplicate-occurrence bug you flagged, plus a related bug I found while reproducing it (shared-domain fallback matching non-option ORES_NATS_* vars, and dropping the domain name from derived option names). See the inline replies above for details.

mcraveiro:
Confirmed and fixed in 086d20a4e. Verified directly: today's live systemd-style invocation (CLI-supplied cert/key/ca + full .env) still parses and connects fine, so no current deployment is affected -- but you're right this set a trap for task 3, which plans to remove that CLI threading. Dropped TLS_CA from the shared allowed suffixes; the whole nats-tls-* trio now stays CLI-supplied/excluded from the shared domain until task 3 designs per-service cert/key env wiring alongside it. Added a regression test pinning that TLS_CA does not resolve via the shared fallback, and updated the task doc's Review table and acceptance criteria.

mcraveiro:
@claude please re-review — pushed a fix for the mTLS-gap you flagged (TLS_CA dropped from the shared domain, whole trio deferred to task 3). See the inline reply above for details.

PR #1820, opened 2026-08-04
[iam] Fix flaky mint_token_respects_the_requested_ttl (1s JWT TTL)

mcraveiro:
## Summary

Fixes a nightly-only test failure: mint_token_respects_the_requested_ttl minted a JWT with a 1-second TTL and validated it immediately against a verifier with zero leeway. Reproduced deterministically under valgrind memcheck (0 errors, 0 leaks, but claims.has_value() still false), confirming a pure timing bug — under enough slowdown/load, the mint-to-validate gap exceeds 1 second and the token is already expired by the time it's checked.

## Changes

- Root cause: 1s JWT TTL + zero verifier leeway makes the test inherently racy against wall-clock time between mint and validate.
- Fix: widened the test's requested TTL from 1s to 60s (tolerance check from 2s to 61s), per direction to never mint sub-minute-TTL JWTs even in tests.
- Left jwt_authenticator::validate() untouched — no leeway added to production JWT expiry checking, keeping the fix test-only.
- Verification: reproduced failure under valgrind memcheck before the fix; after the fix, test passes standalone and under valgrind; full ores.iam.core.tests suite (91 cases, 184 assertions) passes; local build clean (linux-clang-debug-make) and full ctest 71/71 passed.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Hotfix: mint_token_respects_the_requested_ttl flaky on nightly (1s JWT TTL)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/jwt_test_ttl_too_short/story.html) | 536A6313-18D1-49BD-9CA0-EB0D2E3B2D3C |
| Task | [Implement Hotfix: mint_token_respects_the_requested_ttl flaky on nightly (1s JWT TTL)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/jwt_test_ttl_too_short/task_implement_jwt_test_ttl_too_short.html) | 0D055056-3497-43C1-8024-4C1B1A532DD7 |

### DIFF
--- a/.github/workflows/nightly-format.yml
+++ b/.github/workflows/nightly-format.yml
@@ -39,8 +39,16 @@ jobs:
       - name: Install clang-format
         run: sudo apt-get install -y clang-format
 
-      - name: Run clang-format in-place
+      - name: Run clang-format in-place (twice)
+        # clang-format is not idempotent in a single pass for some
+        # multi-line trailing-comment patterns: the first pass can land
+        # on an intermediate state that a second pass reformats further.
+        # Two passes here matches the `format` CMake target and
+        # ores.codegen's clang_format_files(), so none of the three
+        # drift apart from one another.
         run: |
+          find projects -type f \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" \) \
+            | xargs clang-format -i
           find projects -type f \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" \) \
             | xargs clang-format -i
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -507,11 +507,21 @@ find_package(ClangTools)
 if (CLANG_FORMAT_FOUND)
     message(STATUS "Found clang-format (${CLANG_FORMAT_BIN}).")
     #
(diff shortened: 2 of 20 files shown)


mcraveiro:

mcraveiro:
Fixed in bc0e0ad50 — thanks for the optional lower-bound suggestion; deferred as noted (non-blocking, pre-existing pattern). Root cause and fix approach confirmed.

mcraveiro:
Not changed. The comment explaining the 60s/120s convention is a reasonable nit but the PR description already covers it; task doc now also has the rationale in * Review and * Result.

mcraveiro:
Fixed in bc0e0ad50 — changed the requested TTL from 60s (which matched mint_token's own default, weakening the test's ability to catch a dropped/ignored ttl argument) to 120s, distinct from the default. Tolerance bound updated to 121s accordingly. Good catch.

mcraveiro:
Correction: this is actually addressed in bc0e0ad50 — added a comment above the mint_token call explaining both the 120s-vs-default rationale and the no-sub-minute-TTL convention, so a future reader won't reintroduce a short TTL here.

PR #1826, opened 2026-08-04
[synthetic] Add scope and binding_mode to market_data_generation_config

mcraveiro:
## Summary

Adds two orthogonal primitives to synthetic market data generation configs: scope (system/tenant/party, who shares the same generated reality) and binding_mode (bound/sandboxed, whether the data is authoritative). First task of the 'Synthetic data scope and binding' story, which addresses ACME's parties diverging on synthetic setup instead of sharing one reality, and the lack of a safe way to consume synthetic data for testing without affecting real feeds.

## Changes

- Added domain::scope and domain::binding_mode enums to ores.synthetic.api (hand-written, following the existing domain::asset_class precedent)
- Extended market_data_generation_config entity model: nullable party_id/tenant_id, new scope/binding_mode columns, regenerated across all layers (SQL, domain, mapper, generator, Qt)
- Added Qt formatter functions (ScopeLabel, BindingModeLabel) since the generated Qt list-model code has no enum-to-QString path yet
- Added a database insert-trigger guard rejecting an enabled bound config that conflicts with another already-enabled bound config at an equal-or-wider scope; sandboxed configs are exempt
- Verification: local build clean (linux-clang-debug-make); ctest 4/4 passed (ores.synthetic.api.tests, ores.synthetic.core.tests, ores.synthetic.service.tests, ores.qt.tests); conflict-rejection and sandbox-exemption verified live against a recreated DB via direct SQL inserts

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Add scope (system/tenant/party) and binding_mode (bound/sandboxed) to market_data_generation_config](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/synthetic_data_scope_and_binding/task_add_scope_and_binding_mode_to_synthetic_config.html) | 63F39350-B419-4863-A276-6EF29439F93B |


mcraveiro:

mcraveiro:
Thanks for the thorough review — addressed in the latest commit:

**Fixed:**
1. **RLS policies not updated for scope (High)** — both `tenant_isolation_policy` and `party_isolation_policy` now also allow the wider-scope NULL case, so tenant/system-scope rows stay visible to narrower-scope sessions.
2. **Scope/binding_mode missing from history/diff view** — turned out to be a pre-existing codegen gap (`render_is_enum` never existed in the history-field-mapper template, and `has_enum_columns` was wrongly nested inside a `natural_keys`-only conditional). Fixed both in the shared template/codegen and regenerated this entity. Left `feed_binding`/`market_series` (the two pre-existing `is_enum` users) untouched — out of scope for this story.
3. **No CHECK constraint tying scope to nullability** — added via the entity model's new `SQL / Checks` section.
4. **Qt Add flow missing party_id** — `setCreateMode()` now sets it from `clientManager_->currentPartyId()`, matching sibling dialogs.
5. **No automated test coverage** — added a pgTAP test (`synthetic_market_data_generation_config_scope_test.sql`) covering the conflict guard, sandbox exemption, disabled-config non-conflict, and the new CHECK constraint. Verified live.

**Declined (reasoning in the task's `* Review` table):**
- `tenant_id` staying non-optional in the domain/entity/mapper — matches the story's own documented out-of-scope (system-scope isn't exercised end-to-end yet). The suggested `job_definition` precedent is drifted/stale code, not what the current codegen template actually emits (confirmed by reading `cpp_domain_type_class.hpp.mustache`) — fixing it is a template change with cross-component blast radius, better as its own follow-up.
- Generator comment placement nit — tried moving it; the template's appended trailing `;` gets absorbed into the comment, producing invalid C++. Current placement is required, not a style slip.
- `QPointer` lambda note — no code change; the reviewer's own comment already confirms it's not a regression.

Local verification: `compass build` clean, `ores.synthetic.*`/`ores.qt.tests` C++ suites pass, new pgTAP test passes (5/5), full `ores.sql/test` suite runs clean modulo pre-existing unrelated failures that need Acme fixture data this fresh DB doesn't have.

mcraveiro:
Good catch on the WITH CHECK gap — fixed in the latest commit.

Split USING (read visibility, keeps the null-widening so wider-scope rows stay visible to narrower sessions) from WITH CHECK (writes, stays strict): the tenant policy's WITH CHECK now requires the writer be in system-tenant context to write a null-tenant row, and the party policy's WITH CHECK drops the null-shortcut entirely so only a session with no party restriction can write a null-party row. Mirrors `scheduler_rls_policies_create.sql`'s `job_instances_write_policy` precedent, as suggested.

One caveat noted in the task doc: the new pgTAP test runs as the DDL user, which bypasses RLS, so it doesn't exercise this write-boundary directly — full coverage would need session-context (SET ROLE + GUC) test setup, which I'm deferring as disproportionate for this task.

Local verification: `compass build` clean, `ores.synthetic.*`/`ores.qt.tests` pass (one unrelated, pre-existing environment-config test failure in `ores.synthetic.service.tests` — not touched by this diff, confirmed via `git status`), pgTAP test still 5/5 against a freshly recreated DB.

mcraveiro:
Noted — the commit message cited the wrong scheduler precedent (job_instances_write_policy instead of job_definitions_tenant_isolation_policy). The code itself is correct as confirmed. Thanks for the thorough review across both rounds.

PR #1835, opened 2026-08-04
[marketdata,refdata,codegen] Shared derivation lineage: schema + entities

mcraveiro:
## Summary

Implements the shared-derivation-lineage task from the IR curve bootstrapping story: a derivation_kind marker on market_series plus a new observation_lineage table, so any published series/observation can answer observed-vs-derived without guessing from the source string. Also fixes a genuine codegen gap (required plain timestamp columns had no working entity-layer representation) surfaced while building this.

## Changes

- New ores.refdata.derivation_kind entity (simple-lookup, seeded OBSERVED/IR_CURVE_BOOTSTRAP/CRM_DERIVATION), matching the curve_role NONE sentinel convention
- market_series gains derivation_kind/derivation_config_id/derivation_config_version, all not null, paired via a check constraint: OBSERVED requires nil-uuid/zero, any other kind requires both set
- New ores.marketdata.observation_lineage entity: standard bitemporal, not a hypertable, keyed on tenant/party/series/observation_datetime/point_id so reruns don't collide with or overwrite a prior generation's provenance; custom read_latest_by_observation repository method
- Wired into create/drop/populate orchestration and RLS policies matching each table's nearest sibling
- codegen: fixed a genuine template gap -- a required, not-null plain timestamp column had no working entity-layer representation and fell through to a raw time_point field, which sqlgen cannot serialise, failing at compile time. Added the missing required-timestamp path mirroring the existing natural-key string treatment, rather than weakening the column to nullable
- Verification: local build clean on preset linux-clang-debug-make; DB recreate plus SQL schema validator clean (one pre-existing-pattern RLS warning, same as market_observations own sibling gap); ctest restricted to marketdata and refdata tests, 8 of 8 test binaries passed

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Shared derivation lineage: market_series marker + observation lineage table](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ir-curve-bootstrapping-and-republish/task_shared-derivation-lineage.html) | 18B74DD5-0A67-4873-A700-B006D4A99FDD |


mcraveiro:

mcraveiro:
Fixed in 40ba606b0 -- this initial comment was just the in-progress todo list; the substantive findings landed in the two follow-up comments, both addressed.

mcraveiro:
Fixed in 40ba606b0 -- (1) added a derivation_config_version not-equal-to-zero requirement to the non-OBSERVED branch of the check constraint; (2) overrode natural_keys_composite_name to a short explicit name (party_series_obs_point), confirmed the generated index name is now well under 63 bytes. Left the read_latest_by_observation party_id note as-is per your own assessment that it's safe today (the active-row unique index already scopes to one row).

mcraveiro:
Fixed in 40ba606b0 -- (1) ported the same is_required_timestamp mustache blocks into both literate .org tangle sources (ores.cpp.repository.entity_header.org, ores.cpp.repository.mapper_impl.org), re-tangled, and confirmed byte-identical output to what was already committed; this was a real gap -- the drift CI check was red exactly as you predicted. (2) Root-caused the derivation_kind validation gap: the Validations table needs to sit under a top-level Insert trigger heading, not nested under SQL -- market_series.org had this structural mistake pre-existing for asset_class, which I copied when adding derivation_kind. Moved it, fixing validation for both columns.

PR #1844, opened 2026-08-04
[synthetic] Run sandboxed configs end-to-end under a distinct namespace

mcraveiro:
## Summary

Second task of the 'Synthetic data scope and binding' story. Wires ores.synthetic.service to run binding_mode=sandboxed configs end-to-end: generate, publish under a distinct namespace, and be structurally excluded from the bound-feed resolution path real consumers use.

## Changes

- feed_controller::start() gained a binding_mode parameter (default bound, every existing caller unchanged); sandboxed configs publish on a distinct 'synthetic.v1.sandbox.tick.' subject via the new synthetic_producer_subject() free function
- start() skips creating a feed_binding entirely for sandboxed configs, so the marketdata ingest loop -- which always derives its subscribe subject from a feed_binding's source_name -- never subscribes to them regardless of any stray binding
- auto_start_enabled_feeds forwards the container's persisted binding_mode automatically
- Live-testing bug found and fixed: publishing to a subject with no matching JetStream stream crashed the service (SIGABRT, silent crash-loop); added a dedicated synthetic_sandbox_ticks stream mirroring the existing synthetic_curve_ticks precedent
- Added feed_controller_tests.cpp (4 cases) covering the subject-naming primitive directly
- Verification: local build clean (linux-clang-debug-make); ctest passes across ores.synthetic.*/ores.marketdata.*/ores.qt.tests (one unrelated pre-existing environment-config test failure); live end-to-end verification against a running fleet -- sandboxed feed auto-starts and publishes, ingest loop stays at zero subscriptions, no feed_binding/market_series/market_observation ever created for it

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Run sandboxed synthetic configs end-to-end under a distinct namespace](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/synthetic_data_scope_and_binding/task_run_sandboxed_synthetic_configs_end_to_end.html) | 45AB05B7-45D4-4F94-8011-8F293B9C676B |


mcraveiro:

mcraveiro:
Thanks — both real findings fixed in the latest commit.

**Fixed:**
1. **folder_feed_control_handler never forwarded binding_mode** — every feed started via "start everything under this folder" (the path Qt, ores.shell, and a wt workflow step all go through) silently defaulted to bound. Fixed by mirroring `auto_start_enabled_feeds`: queries `market_data_generation_config_repository`, maps by container id, forwards `container.binding_mode` into `ctrl_->start()`.
2. **Already-running branch re-derived the binding decision from the wrong input** — `start()` gated `ensure_feed_binding` on the *current call's* `binding_mode` argument, not the feed's actual stored mode. Since `market_feed_config_handler` and the (now-fixed) folder handler default to `bound` when they don't know better, a caller restarting an already-sandboxed feed through either path would have created a real `feed_binding` for it — silently defeating the guarantee for a feed that's still, in fact, publishing on the sandbox subject.

Extracted the decision into a small free function, `should_ensure_feed_binding(already_running, requested, stored)`, so this exact interaction is now directly unit-tested (3 new cases) rather than only covered by the original live verification, which — as flagged — only exercised the auto-start path in isolation.

**Minor:** removed the now-dead `#include <set>` in `application.cpp`.

Local verification: `compass build` clean, `ores.synthetic.*` suite passes (7 cases in `feed_controller_tests.cpp` now), and I re-ran the live sandboxed-auto-start verification against a running fleet after these fixes to confirm nothing regressed.

mcraveiro:
Round 2 confirms both fixes hold across all three call sites. Thanks for the thorough review.

PR #1845, opened 2026-08-05
[compass] WS-1: per-environment Claude slices with limits

mcraveiro:
## Summary

Second staged PR of the per-environment systemd resource limits task: nests each Claude session in a per-environment slice and gives it real memory/CPU limits via a template drop-in.

## Changes

- _SLICE_NAME constant replaced with _slice_name(env_name); systemd-run now targets the per-environment slice
- New app-claude-.slice.d/50-limits.conf template drop-in (dash-truncated, covers every environment with one file)
- _ensure_slice_deployed() generalised to sync a manifest of source and dest pairs instead of a single file
- Verification: called the deploy function directly to avoid disrupting this session's own running scope; confirmed both files deployed; started a throwaway systemd-run scope and confirmed via systemctl show that all four limits apply; confirmed the parent slice is unaffected and pre-existing running sessions undisturbed

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Systemd resource management and per-environment isolation](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/systemd-resource-management-and-isolation/story.html) | 344FF2CC-E27B-4EE7-AAED-F6E2BE77FABE |
| Task | [Per-environment systemd resource limits (Claude slices, services, builds)](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/systemd-resource-management-and-isolation/task_systemd-per-environment-resource-limits.html) | 9913E08D-977A-4C36-8376-62BCEA9A41CB |

### DIFF
--- a/doc/agile/versions/v0/sprint_25/systemd-resource-management-and-isolation/story.org
+++ b/doc/agile/versions/v0/sprint_25/systemd-resource-management-and-isolation/story.org
@@ -40,9 +40,9 @@ implementation.
 |---------------+----------------------------------------|
 | State         | STARTED                              |
 | Parent sprint | [[id:0E7321A7-A66D-4CE6-B3E7-89F2E35CA48B][Sprint 25]] |
-| Now           | Per-environment resource limits task STARTED -- WS-7 change 1 (drop build lock slot c) implemented, first PR in flight. |
+| Now           | Per-environment resource limits task STARTED -- WS-7 change 1 (merged) and WS-1 (per-environment Claude slices) implemented, second PR in flight. |
 | Waiting on    | Nothing.                               |
-| Next          | WS-1 (per-environment Claude slices with limits). |
+| Next          | WS-5 (unprivileged half): emacs.service protection drop-in. |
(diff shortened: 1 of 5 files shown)


mcraveiro:

mcraveiro:
Fixed in 8ca1153ee — corrected the docstring's drop-in name to app-claude-.slice.d/ (trailing dash). Declined the dash-env-name observation: genuinely latent, no current environment name uses a dash (all use underscores), and the drop-in still applies at every implicit level so it's harmless if it ever does occur.

mcraveiro:
Fixed in 8ca1153ee — same docstring fix. Declined both the dash-env-name edge case and unit-test suggestion as non-blocking for this change (no test file precedent in this module, per the WS-7 PR too).

mcraveiro:
Fixed in 8ca1153ee — corrected the docstring's drop-in dir name. Declined the dash-env-name edge case as non-blocking (harmless, and not hit by any real environment today).

PR #1846, opened 2026-08-05
[codegen,refdata] Fix workspace inheritance: exact-match to resolution chain

mcraveiro:
## Summary

The has_workspace_id repository template block generated an exact-match workspace_id == wid predicate, not the resolution-order query the architecture doc describes -- so single-parent workspace inheritance (a child workspace seeing its parent's data) did not function for any entity. The end-to-end plumbing already existed (context::workspace_resolution(), X-Workspace-Resolution NATS header, request_context.cpp); only the template's query was wrong.

## Changes

- Fixed cpp_domain_type_repository.cpp.mustache's read_latest(context ctx) list-all method to use workspace_id IN (chain) when a resolution chain is set, falling back to exact match otherwise. Edited via the literate .org source and re-tangled, not the generated mustache directly.
- Left find-by-id/batch/count/remove/as-of methods on exact match: writes/deletes should never span the inheritance chain, and UUID primary-key lookups already fully identify the intended row.
- Regenerated and verified against the 8 profile-based refdata convention entities (workspace-scoped-lookup profile): zero-diff, idempotent regeneration.
- Added a new permanent repository test proving a child workspace with no local rows inherits Live's rows, plus a fallback test confirming empty-chain exact-match is unchanged.
- The 13 trading has_workspace_id entities could not be regenerated in this pass -- blocked on a separate, already-tracked task (836BB4CE) fixing a legacy primary-key model format repo-wide. The template fix is generic and needs no further work once that lands.
- Verification: full local build clean (linux-clang-debug-make); full ctest 71/71 passed; ores.refdata.core.tests 178/178 passed (176 existing + 2 new); ores.codegen pytest 108/108 passed.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Workspaces: finish the feature](https://orestudio.github.io/OreStudio/doc/agile/product_backlog/inbox/finish_workspaces_feature/story.html) | 603F4845-B4A0-41EB-ACF9-B0FC1387E1C4 |
| Task | [Upgrade repository template: exact-match workspace_id to resolution-order inheritance](https://orestudio.github.io/OreStudio/doc/agile/product_backlog/inbox/finish_workspaces_feature/task_fix_workspace_inheritance_query.html) | 3B718FE6-7CE1-43E1-93D0-E1957A59106F |


mcraveiro:

PR #1847, opened 2026-08-05
[refdata,sql] IR curve bootstrap config recipe entities

mcraveiro:
## Summary

Implements the ir-curve-bootstrap-config task from the IR curve bootstrapping story: two new ores.refdata entities recording the recipe (source/output series, curve family role, build-order dependency, interpolation method, day-count convention, split tenor) that bootstraps a raw instrument grid into a discount curve.

## Changes

- New ir_curve_bootstrap_config entity: soft cross-component references to marketdata.market_series, curve_family_role (FUNDING/PROJECTION), self-referencing discount_curve_config_id encoding the Funding-before-Projection build order, interpolation_method, day_count_convention (validated), split_tenor_code
- New ir_curve_bootstrap_pillar entity: one-parent-many-children pillar list, FK to the parent config, curve_role_code validated -- same shape as ir_curve_template_entry but source-independent
- output_series_id is a natural key alongside party_id (exactly one active config per output series) -- caught and fixed after the first codegen pass produced a wrong one-config-per-party unique index
- Two-tier discount_curve_config_id enforcement: a check constraint shape-validates, and a custom insert-trigger block validates the referenced config is itself FUNDING (no chaining), reusing the same per-entity implements-marker mechanism curve_role's Extra drops already uses
- Dual RLS (tenant plus party isolation) on both new tables, matching ir_curve_template_entry's own precedent
- Also closed out the observation_lineages RLS gap left over from the previous task, and filed a capture plus a documented validation_ignore.txt entry for a pre-existing, unrelated, by-design warning on synthetic_market_data_generation_configs_tbl rather than leaving it unexplained
- Two follow-ons flagged explicitly in the task's Result section rather than silently claimed done: atomic output-series minting is schema-only here (the actual cross-component orchestration belongs to a later task), and the two-tier rejection paths were verified by reading generated SQL plus a clean db recreate, not yet by an automated test
- Verification: local build clean (linux-clang-debug-make); compass db recreate plus SQL schema validator clean, 0 warnings; refdata-codegen-drift check clean; ctest -R refdata 4 of 4 test binaries passed

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [IR curve bootstrap config: refdata entities for the bootstrap recipe](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ir-curve-bootstrapping-and-republish/task_ir-curve-bootstrap-config.html) | 5255E8A6-9EE6-44FA-A5CB-41ABD508ED37 |


mcraveiro:

mcraveiro:
Fixed in a51ec8745 -- (1) reworded the org source to avoid embedding a literal equals sign inside verbatim spans, confirmed the regenerated header comment is clean. (2) Added a note to the task's Follow-ons section: pillar tenor validation is a soft, application-layer reference, matching ir_curve_template_entry's own precedent, not something the acceptance bullet should have implied is DB-enforced.

mcraveiro:
Fixed in a51ec8745 -- (1) added a second guard to the same insert-trigger block: editing an existing FUNDING config's own curve_family_role away from FUNDING now fails if one or more active PROJECTION configs still reference it. Confirmed this actually fires on every edit, not just fresh inserts, since a version bump on this bitemporal table is itself a new INSERT through the same trigger. (2) reworded the org source to avoid a literal equals sign inside a verbatim span; regenerated header comment is now clean. (3) not changed -- agreed this is a pre-existing, shared codegen-template gap (confirmed the sibling table you cited has the exact same issue), tracked as a follow-up rather than fixed one-off here.

mcraveiro:
Fixed in a51ec8745 -- (2) you're right this wasn't fully closed: dropped party_id from the natural key entirely rather than leaving output_series_id uniqueness scoped by party. It's now the sole natural key, tenant-wide, so two parties genuinely cannot each claim the same output series at the DB level -- doesn't rely on the soft cross-component invariant anymore. (1) and (3) not changed -- (1) is a pre-existing, shared codegen-template naming gap (same as the sibling table you cited), tracked separately; (3) matches current project convention for other recent codegen entities, agreed non-blocking.

PR #1851, opened 2026-08-05
[agile] Capture: Add server-side search/filter to list windows

mcraveiro:
## Summary
List windows only support offset/limit pagination with no search field, making it impractical to find a specific record in large tables (e.g. ~13k GLEIF-imported counterparties) without paging through hundreds of pages. Scoped as a cross-cutting codegen capability across all entities, not a one-off fix.

- New inbox capture describing the gap, likely shape (filter_text field + SQL WHERE/full-text search + toolbar search box), and references to existing partial precedent.

## Test plan
- Docs-only, no build/test impact.


mcraveiro:

PR #1855, opened 2026-08-05
[ores.service,codegen] Extract shared host.cpp envelope, fix drift bug

mcraveiro:
## Summary

Audits and unifies application.cpp/host.cpp/config layers across all services (task_unify_application_cpp_shape), prompted by earlier tasks tripping on ores.cli's divergent shape. Analysis found 17 of 20 host.cpp files byte-identical modulo naming -- a genuine shared-code candidate, not codegen. Extracts the parse/log/construct/run/catch envelope into ores::service::run_host_async(), migrates all 18 async services to use it, and updates the codegen archetype so future services do too.

## Changes

- Add ores::service::service::run_host_async() (ores.service/include/ores.service/service/host_runner.hpp + host_runner_impl.hpp) with an optional host_runner_options for custom startup/shutdown/failure messages
- Migrate all 18 async-shaped host.cpp files (17 domain services + ores.http.server) to the shared helper; fixes a real drift bug in ore.service/workflow.service (silently skipped logging e.what() for non-boost exceptions) as a side effect
- Update the component-service codegen archetype (ores.cpp.service-app.host_impl.org) so future-generated services use the shared helper from generation
- Leave ores.cli/ores.shell/ores.compute.wrapper hand-written -- their differences (sync execution, custom constructor args, cli's DB-free interception branch) are real behavioural divergence, not boilerplate drift
- Fix a critical bug found only by live verification: host_runner_options was originally taken by const reference with a default value, a dangling reference the moment the lazily-started boost::asio coroutine resumes -- every migrated service SEGV'd on startup. ctest never exercises host::execute so this passed unit tests cleanly; fixed by taking the options struct by value
- Fix an unrelated crash hit during full-suite verification: ores.synthetic.service.tests aborted on a feed_controller test (added by PR #1844, unrelated to this diff) whose 289-byte name exceeded the 255-byte NAME_MAX filesystem limit when used as a raw log filename; shortened the name and filed a capture for the general fix
- Files a follow-up capture recommending E3647ADD's standard_service_options design become composable opt-in blocks rather than assuming nats+database always present, so the 5 tools it currently excludes could adopt it too
- Verification: ctest 71/71 passed (multiple times, across the coroutine-lifetime fix and the test-name-length fix); live systemd verification -- full stop/start cycle, 22/22 units ready with 0 retries, http.server's custom banner/shutdown confirmed, ore.service/workflow.service confirmed healthy

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Reduce and configure per-service DB connection pool size](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/reduce_db_connection_pool_load/story.html) | 1B68F0E3-59FE-4BBE-B0BD-7FA80D9713B3 |
| Task | [Unify application.cpp and service initialisation across services](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/reduce_db_connection_pool_load/task_unify_application_cpp_shape.html) | 4544572F-7C42-4A37-8440-C11B58511C94 |


mcraveiro:

mcraveiro:
Fixed in 8766f75e8 -- corrected task_unify_application_cpp_shape.org and story.org to list only ores.cli/ores.shell as hand-written, and filed a follow-up capture (review_cli_shell_host_divergence.org) to properly review whether their remaining divergence is load-bearing or just historical, since the ores.compute/wrapper mis-scoping suggests the original judgment wasn't rigorous.

mcraveiro:
Fixed in 8766f75e8 -- same finding as the other passes, corrected both docs and filed a follow-up capture for a proper cli/shell divergence review.

mcraveiro:
Fixed in 8766f75e8 -- corrected both docs to say ores.cli/ores.shell only, and noted ores.compute/wrapper's host.cpp joined the shared helper while its config/options composition (no database) remains the real, still-tracked divergence per design_standard_service_options_composable.org. Also filed a new capture to properly review whether cli/shell's divergence is load-bearing or historical, since this mistake suggests the original judgment call wasn't rigorous enough.

PR #1859, opened 2026-08-05
[ores.compass] Wire systemd generate to the codegen model, strip redundant NATS CLI threading

mcraveiro:
## Summary

compass systemd generate/quadlet queried postgres for a fleet definition that now lives, more completely, in service_registry.org's deployment aspect. Migrated to load that org model directly (no DB round-trip), and stripped the now-redundant --nats-url/--nats-subject-prefix/--nats-wire-format CLI threading from generated units and compass shell, since every service already resolves these from .env via the shared-domain fallback tier.

## Changes

- systemd_generate.py: fetch_service_definitions/fetch_dependencies now load service_registry.org via codegen.org_loader instead of querying ores_controller_service_definitions_tbl/ores_controller_service_dependencies_tbl
- Stripped --nats-url/--nats-subject-prefix/--nats-wire-format from DEFAULT_ARGS_TEMPLATE and compass shell's flag_for; kept the nats-tls-* trio CLI-supplied per the mapper task's Decisions
- Fixed a second consumer (compass_services.py's _service_units()) that called the old fetch_service_definitions(env) signature directly and would have broken on this migration
- Verification: local build clean; ctest not applicable (Python-only change) -- verified live instead: generated/deployed/started the full fleet (23 units + nats-server), confirmed no NATS URL/prefix/wire-format CLI flags anywhere, all services connect correctly; full ores.codegen (117/117) and ores.compass (43/43) pytest suites pass

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Collapse NATS config passthrough to a single environment-mapper fallback tier](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/process-supervisor-config-passthrough-gap/story.html) | 9E3EC9E2-742F-4830-839D-99EF2C2BC626 |
| Task | [Wire systemd generate to the codegen model and strip redundant NATS CLI threading](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/process-supervisor-config-passthrough-gap/task_wire_systemd_generate_to_codegen_model.html) | ED259270-CEAF-4456-A6C9-8974E9E5B8B2 |


mcraveiro:

mcraveiro:
Not changed. All three minor observations are accepted-as-is: (1) the args_template-full-override capability is intentionally narrowed to additive Extra-args per the org model's design (no service currently needs a full override); (2) missing/malformed service_registry.org surfacing as a raw traceback is acceptable since the file is checked into the repo and always present in a normal checkout; (3) binary_name==service_name is an intentional, verified simplification per the task doc's Plan (confirmed against the old seed data).

mcraveiro:
Not changed. The extra_args-containing-nats-placeholders scenario is correctly identified as theoretical -- grep confirms no such bullets exist in service_registry.org today, and adding a guard for a non-existent case would be speculative. Agreed a fixture-based unit test would be valuable, but this migration relies on the live fleet verification described in the PR body as a reasonable tradeoff given no prior test coverage existed for this module.

mcraveiro:
Thanks for verifying the org_loader/systemd_generate data-shape mapping and the shared-domain fallback claim against parser.cpp directly -- no changes needed.

PR #1860, opened 2026-08-05
[synthetic] Consolidate ACME's UK/US/HK configs into one tenant-scoped config

mcraveiro:
## Summary

Third task of the 'Synthetic data scope and binding' story. Consolidates ACME's six party-level synthetic containers (two themes x three parties) into one tenant-scoped bound config, so all three parties share one synthetic reality instead of independently-seeded copies.

## Changes

- Fixed two pre-existing schema-drift bugs in ores.marketdata blocking ACME provisioning entirely: marketdata_publish_from_dq_create.sql hardcoded stale lowercase asset_class literals rejected by current title-case reference data, and never set the derivation_kind/derivation_config_id/derivation_config_version columns added by the shared-derivation-lineage work
- Fixed a regression in this story's own first task: synthetic_publish_from_dq_create.sql's insert into market_data_generation_config never set the new scope/binding_mode columns, so every DQ-published config silently defaulted to nulls, violating the NOT NULL constraint
- Fixed the conflict-guard trigger (also from the first task) to no longer block two enabled bound configs at the same party -- an established, legitimate, pre-existing pattern (multiple named containers per party) unrelated to per-instrument conflict; added pgTAP coverage for it
- Extended the RLS read-widening from the first task down into fx_spot_generation_config and gmm_component -- promoting a container's scope alone doesn't make its sub-configs visible to other parties without the same treatment one/two hops down
- Consolidated ACME's six party-level containers into one tenant-scoped bound config via the same versioned-insert path the app's own writes use (exercising the real conflict guard, not bypassing it)
- Verification: local build clean (linux-clang-debug-make); ctest passes across ores.synthetic.*/ores.marketdata.*/ores.qt.tests; pgTAP suite (6 cases) passes; full ACME provisioning runs end-to-end cleanly for the first time since the story's first task landed; live-verified all three ACME parties see the identical FX sub-config set via RLS-scoped queries as the real, unprivileged readonly role

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Consolidate ACME's UK/US/HK party configs into one tenant-level bound config](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/synthetic_data_scope_and_binding/task_consolidate_acme_synthetic_configs_to_tenant_scope.html) | 78B24A7B-8FBC-4B7C-AABB-5DF11E205F58 |


mcraveiro:

PR #1877, opened 2026-08-06
[compass] Support multiple named .env files via --env / --env-file

mcraveiro:
## Summary

Add -env NAME (resolves to .env.NAME) and -env-file PATH flags to every compass command. Works before or after sub-command.

## Changes

- Flag parsed early in compass.py main() before any command dispatch. _resolve_env_file() converts -env newton shorthand to .env.newton.

Two env-reading functions modified: _read_env_map() for internal cmd_* and load_env() for db/services/systemd/client/claude sub-commands. env_file parameter threaded through five run() functions.

Backwards compatible: no flag means default .env. Working .env never mutated. Recipe included.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Compass quality-of-life improvements: remaining follow-ups](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/compass_quality_of_life_followups/story.html) | E599E5A0-84C9-4636-B478-5D1478688B9B |
| Task | [Support multiple named .env files, e.g. per remote host](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/compass_quality_of_life_followups/task_support-multiple-env-files.html) | 4E613004-5331-42C2-A654-56292F763095 |


mcraveiro:
Fixed in 8a215b9d9:

1. **Missing value**: _pop_env_file_arg() now errors with "--env requires a value" when the flag is last or followed by another flag.
2. **Both flags together**: now errors "cannot use both --env and --env-file together" instead of leaving a dangling flag.
3. **nats init**: Added a Note to the task org clarifying nats and other env-reading commands are explicitly out of scope for this iteration (Acceptance only required db + services). A follow-up task can extend coverage.
4. **--env-file with bare name**: By design — _resolve_env_file() doesn't know which flag was used, and the recipe documents the expected usage for each.

mcraveiro:
Fixed in 8a215b9d9:

1. **Both flags together**: _pop_env_file_arg() now detects and errors with "cannot use both --env and --env-file together" before any argv mutation occurs — strict-argparse sub-commands are protected.
2. **Missing value**: now errors with "--env requires a value" when the flag has no value, rather than silently falling back to default .env.

Test scenarios table left empty per project convention (manual QA scenarios are scaffolded separately).

PR #1898, opened 2026-08-07
[dq] Bind ores.dq entities to profiles; commission full C++/Qt stack

mcraveiro:
## Summary

Rounds out the ores.dq entity-classification-drift work: binds SQL entities to variability profiles, migrates lei_entity/lei_relationship/report_definition/synthetic_fx_spot_config onto full codegen, consolidates ores.qt.data_management into ores.qt.dq (deleting the stepping-stone plugin), migrates catalog/change_reason/change_reason_category/data_domain/dataset_bundle onto full codegen with the generic HistoryDialog, and fixes several bugs surfaced by end-to-end GUI test scenarios.

## Changes

- Bind dq SQL entities to variability profiles (simple-lookup, uuid-surrogate-lookup, artefact-staging-only); verify zero-diff regen
- Commission full C++/Qt for lei_entity, lei_relationship, report_definition, synthetic_fx_spot_config
- Consolidate all ores.dq entities into ores.qt.dq; decommission ores.qt.data_management
- Migrate catalog/change_reason/change_reason_category/data_domain/dataset_bundle onto full codegen; migrate subject_area's history dialog by hand onto the generic HistoryDialog
- Fix artefact_type's missing Detail fields and its NOTIFY trigger never being wired into dq_create.sql (same gap found and fixed for code_domain and dataset_bundle)
- Fix change_reason/change_reason_category showing duplicate rows: drop the redundant system-tenant RLS read-policy fallback for tables the tenant provisioner already copies per-tenant

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Entity classification and drift baseline: ores.dq](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-dq/story.html) | A452F2A7-D9CB-48E6-8B23-4509AFFC5699 |
| Task | [Bind ores.dq entities to profiles; verify zero-diff regen](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-dq/task_bind-dq-entities-to-profiles.html) | 163AC07E-19E4-4486-A503-8170D9B46108 |
| Environment | jolly_knuth | |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1915, opened 2026-08-08
[dq,qt,sql] Fix shared-tenant edit duplicates; remove broken LEI browsing UI

mcraveiro:
## Summary

Fixes two real bugs surfaced by GUI retesting after PR #1898: artefact_type/dataset_bundle producing duplicate rows when a tenant admin edits a DDL-seeded shared row, and the LEI entities/relationships read-only list windows erroring on open because they query a main table that deliberately does not exist.

## Changes

- Extend ores_iam_provision_tenant_fn to copy artefact_types/dataset_bundles per-tenant at provisioning time, same pattern as change_reason/change_reason_category
- Add missing RLS to artefact_types_tbl (had none at all); drop the redundant system-tenant read fallback on dataset_bundles_tbl
- Remove the LEI entities/relationships Explorer list windows and LEI Registry menu entirely -- their data has no working same-component read path by design

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Entity classification and drift baseline: ores.dq](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-dq/story.html) | A452F2A7-D9CB-48E6-8B23-4509AFFC5699 |
| Task | [Bind ores.dq entities to profiles; verify zero-diff regen](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-dq/task_bind-dq-entities-to-profiles.html) | 163AC07E-19E4-4486-A503-8170D9B46108 |


mcraveiro:
Fixed in ce83e21d0 -- extended the copy fix to all 10 tables sharing both true preconditions (DDL-seeded system tenant AND Qt-editable), which I audited precisely: your finding correctly flagged the RLS-pattern list, but that list wasn't itself proof of the bug -- catalogs turned out not to be DDL-seeded at all (verified via grep, no populate script inserts into it), and badge_mappings/dataset_bundle_members turned out to have no save path in the Qt UI (read-only display), so neither is actually reachable. datasets is deliberately excluded still -- its insert trigger validates 7 FK references (catalog, subject_area, coding_scheme, 3 dimensions, methodology, artefact_type), too high-risk to rush in this round. Also added the missing drop-policy pair and fixed the stale LEI Registry doc references.

mcraveiro:
No additional issues raised beyond what's covered in my reply to the parallel review comment -- addressed in ce83e21d0.

mcraveiro:
Fixed in ce83e21d0 -- catalogs turned out not to be DDL-seeded (verified via grep across projects/ores.sql/populate/dq/, no populate script inserts into it), so it's correctly excluded. subject_areas and data_domains are both fixed, along with 8 more confirmed positives found via the same audit. See my reply to the other review comment for the full list and rationale.

PR #1919, opened 2026-08-08
[qt.refdata] Curve Builder Workbench: Save/Bootstrap/Publish

mcraveiro:
## Summary

Non-modal Curve Builder Workbench under the Market Data menu (Conventions/Pillars/Build & Diagnostics tabs) guiding a user through defining an IR curve bootstrap recipe, previewing its computed output, and explicitly publishing it -- three distinct actions (Save/Bootstrap/Publish), never fused. Plus the backend compute/publish split and analytics.quant building blocks (forward_rate_calculator, curve_health_checker) the diagnostics tab needs.

## Changes

- CurveBuilderWorkbench: non-modal MDI window, three tabs, Currency/Index-gated Source/Output Series and Discount Curve Config pickers (with inline New Series creation), tenor/curve-role combos backed by seeded reference data, forward-rate/discount-factor charts, health-check findings
- Backend: curve_republish_service split into compute() (preview, pure) and republish() (write); compute_curve_request/response NATS protocol pair; marketdata::curve_bootstrap:compute permission
- ores.analytics.quant: forward_rate_calculator and curve_health_checker building blocks, fully unit tested
- Market Data menu wiring: Curve Bootstrapping submenu (Bootstrap Configs.../Build Curve...), IrCurveBootstrapConfigController wired into RefdataPlugin for the first time
- Fixed a real backend gap found along the way: ir_curve_bootstrap_config/pillar's NATS handlers existed but were never registered, so the backend never subscribed to their subjects
- Fixed a source/output series clobbering data-integrity bug (client+server guards against source_series_id == output_series_id)
- Fixed pillar sequence_index being derived from table row order instead of actual maturity, and split_tenor_code never being validated/auto-derived
- Fixed two hardcoded, never-seeded change_reason_code literals blocking every first Publish, plus a duplicate-key bug on re-publishing the same as-of
- Standardised error display (MessageBoxHelper::critical) across the new picker dialogs and the workbench itself
- Verification: local build clean (linux-clang-debug-make); rat 0 failures across the whole suite; manually verified end to end via the Verify Curve Builder Workbench test scenario through several real Save/Bootstrap/Publish rounds against seeded Acme Corporation data

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Curve builder UI: bootstrap config CRUD + a guided Save/Bootstrap/Publish workbench](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ir-curve-bootstrapping-and-republish/task_curve_builder_ui.html) | 794D9D94-8E62-4232-A8FD-769B518DB04C |


mcraveiro:

mcraveiro:
Addressed in a1e07d006:
1. Fixed - hasBootstrapped_ is now reset to false in onSaveClicked's success handler, so Publish is re-disabled until a fresh Bootstrap runs against the saved recipe.
2. Fixed - removed the second collectPillarsFromTable() call; the change-reason fields are now stamped directly onto the already-collected config_/pillars_ instead of re-deriving from the table (which was the source of the row-index/id scrambling).
3. Fixed - onBrowseSourceSeriesClicked now excludes an already-picked Output series, symmetric with onBrowseOutputSeriesClicked.
4. Fixed - added CurveBuilderWorkbench::markAsStale() and a matching qobject_cast branch in notifyOpenDialogs, so stale-data warnings now fire for open curve-recipe windows too.
5. Fixed (UI-side) - onBrowseDiscountCurveConfigClicked now guards against self-reference. Confirmed the server already rejects it independently via the ir_curve_bootstrap_configs table's own discount_curve_config_id-not-equal-id check constraint.
6. Fixed - forward_rate_calculator::calculate now throws std::invalid_argument on a non-positive discount factor, with a new test case covering it.
7. Not changed. Both curve_bootstrap:compute and :republish work via the blanket marketdata wildcard already used throughout marketdata's permission model; scoping to per-action grants is a cross-cutting IAM concern, not specific to this PR.
8. Not changed. Low-risk in practice (republish is user-triggered, not high-frequency) - a race here degrades to the exact pre-fix duplicate-key error this PR resolves, not silent data corruption. Worth revisiting if republish is ever automated/scheduled.
9. Fixed - system.curve_bootstrap's applies_to_amend flipped to true (a rerun over the same as_of is a version-bump/amend of the lineage row).
10. Not changed. Pre-existing, not introduced by this PR, and out of scope here - worth its own follow-up ticket.

mcraveiro:
Addressed in f07dce370:
1. Fixed - wrapped the forward_rate_calculator::calculate() call in renderBootstrapResults in try/catch, degrading to a banner ("Bootstrap succeeded, but instantaneous forward rates could not be computed") instead of letting the exception escape the Qt slot and terminate the app. Good catch - this was a genuine regression from the previous fix.
2. Fixed - dropped the dead change-reason stamping from both collectPillarsFromTable() and collectConfigFromUi() (the latter had the identical issue, just not called out explicitly); onSaveClicked's single post-prompt stamp is now the only place change_reason_code/change_commentary get set.

PR #1936, opened 2026-08-09
[codegen] Add oresmd quote-type codegen: metatype, facet, archetypes, templates

mcraveiro:
## Summary

Infrastructure PR (PR 1 of the task's two-PR plan) for the oresmd quote-type codegen port. Adds the `oresmd_quote_type` metatype and org model loader, the `ores.cpp.oresmd` facet with 8 archetype nodes and their tangled templates, the 7 asset-class org specs (ir, credit, equity, commodity, fx, inflation, correlation) with a batch manifest entry point, and the enums archetype verified **zero-diff** against the hand-crafted `oresmd_enums.hpp`. No generated file replaces hand-crafted output yet — PR 2 wires the generated files as canonical and adds drift CI (`cmake-sources-drift`).

## Changes

- [agile] Analysis and plan for the oresmd codegen port; scaffold the codegen task and the YAML prototype generator.
- [codegen] Register the `oresmd_quote_type` metatype, add the org model loader (single spec + batch manifest), route through codegen core.
- [codegen] Add the `ores.cpp.oresmd` facet (`#+model_types: oresmd_quote_type`) and the enums archetype at the correct physical-space locations.
- [codegen] Add the remaining 7 archetypes (identifiers, requirements, parser, projections, resolver, parser_tests, projection_tests) as tangled templates, and the batch loader driven by the manifest's Spec files table.
- [codegen] Add all 7 asset-class specs with `comment` column, `* Enum brief` / `* Enum footer` sections; the enums template renders `oresmd_enums.hpp` byte-identical to the hand-crafted file (clang-format normalizes whitespace; verified via `compass codegen entity generate model --address ores.cpp.oresmd --diff`).
- [codegen] Guard bulk `regenerate --component marketdata` from the not-yet-implemented archetypes via the catalogue `exclude_org_types` column; drop stale un-referenced fixture copies.

The branch was rebased on latest `main` (post codegen-catalogue refactor); full build of all targets succeeds, 129 codegen tests pass (2 pre-existing profile-binding failures unrelated).

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Extend oresmd to full ORE quote-type coverage](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/extend-oresmd-full-quote-type-coverage/story.html) | D566131C-D08C-4AFE-950E-B3DD26EB2C24 |
| Task | [Codegen oresmd quote types](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/extend-oresmd-full-quote-type-coverage/task_oresmd-codegen.html) | 602454B1-2D6E-4E40-BBD0-DA3F6446A569 |


mcraveiro:
Fixed in 4d90aff9b and 9acd412c9.

1. Trailing comma: switched the variant loops to the populated {{^last}} sigil (nothing ever set {{^-last}}) and added _mark_last_item(specs) on the outer list in core.py so the final variant member renders comma-free. Verified: the identifiers variant renders all 7 members with no trailing comma. While in these templates I also found and fixed a related escaping bug: the {{cpp_type}} double braces HTML-escaped angle brackets in Fields cells, so a member declared std::optional of cpp_type came out with entity escapes in the generated header — switched to the triple-brace {{{cpp_type}}} unescaped form in both templates.

2. oresmd_verify.py: agreed it was dead on arrival — it referenced a template path that does not exist, applied none of the context shaping, and nothing referenced it. Removed in 9acd412c9.

mcraveiro:
Follow-up fix in bc990693e: the review-table note added in 4d90aff9b contained a literal triple-brace cpp_type tag in the task doc. The site build expands triple braces as Org macros, so deploy_site aborted with 'Undefined Org macro: cpp_type' — that was the site/CI failure. Reworded the note and switched backtick markup to org =monospace= form; deploy_site builds cleanly locally now.

mcraveiro:
Agreed — not blocking this infrastructure PR; only the enums archetype is claimed wired up here. Verified your claims against the tree before deciding: 6 of 7 specs indeed lack a type Fields row, the requirements template does double-wrap the already-optional quote_type/point/vol rows, and the has_quote_type/has_point/has_vol/generated guards are template-only, never populated by the loader or core.

Decision: accepted, deferred to PR 2 — recorded as row 3 in the task's Review table (pushed in f1cd8bd75). PR 2 wires these archetypes up with the same zero-diff verification used for enums, which is where the loader-side guard population, default-value support, and the missing type rows land.

mcraveiro:
Done — the round-2 identifiers/requirements finding is now row 3 in the task's Review table (Accepted — deferred to PR 2), pushed in f1cd8bd75. Thanks for re-verifying the earlier fixes against the tree.

PR #1944, opened 2026-08-10
[ci] Run the vendored nats-server binary on all CI platforms

mcraveiro:
## Summary

Continuous MacOS CI aborts at the Start NATS server step: the nats-server binary was only installed by brew inside the Run CTest workflow step, which runs after the step that launches it, so the launch emits command not found and the 30s readiness poll fails. Linux CI installed apt nats-server via env install-packages and Windows CI downloaded a GitHub release pinned v2.10.24: three platforms, three sources, three versions. Every CI workflow now gets an identical Setup NATS server step that unpacks the vendored nats-server v2.14.3 release (external/nats/packages) onto PATH, so every platform runs the same pinned binary.

## Changes

- All four CI workflows (continuous-linux, continuous-macos, continuous-windows, nightly-linux) gain an identical Setup NATS server step before the Start NATS server step
- New shared build/scripts/setup_nats_server.sh (Linux/macOS) and its PowerShell equivalent setup_nats_server.ps1 (Windows) unpack the vendored nats-server v2.14.3 release into build/nats/vendored and prepend it to PATH via GITHUB_PATH
- macOS fix: nats-server is removed from the Run CTest step brew list, which ran after Start NATS server and caused the command-not-found failure
- Windows CI drops its GitHub-release download pinned v2.10.24 for the vendored v2.14.3 archive
- Verification: bash -n plus functional smoke tests of both scripts on a Linux host (unpack, nats-server reports v2.14.3, second run idempotent); yamllint and YAML parse clean on all four workflows. No C++ or SQL changed, so no build or test suite was run; the four workflows only run on schedule and tags, so the next scheduled runs are the on-runner validation

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Hotfix: macOS CI fails to start NATS server, nats-server not installed](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/fix_macos_ci_nats_server_not_found/story.html) | 69D2C83C-F79B-4B52-98B9-F3A8664B0BD8 |
| Task | [Scaffold story: Hotfix: macOS CI fails to start NATS server, nats-server not installed](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/fix_macos_ci_nats_server_not_found/task_scaffold_fix_macos_ci_nats_server_not_found.html) | CBAD58B1-26F3-405E-9155-EF039B429B30 |


mcraveiro:

mcraveiro:
Round addressed in 9721f817.

1. Fixed. The Linux branch now checks arch like the Darwin branch (arm64 selects linux-arm64). Since the vendored set has no linux-arm64 archive, I also added a missing-archive guard: an unsupported combo now fails at setup time with a clear 'no vendored nats-server archive' diagnostic instead of an opaque tar or exec-format error. Verified with bash -n and by faking uname -m (exit 1 with the clear diagnostic).
2. Not changed. Removing nats-server from _BASELINE would break local dev: the generated systemd unit for this environment runs ExecStart=/usr/sbin/nats-server (systemd/units/nats-server-bright_faraday.service), a path only the apt package provides, and compass env install-packages is shared between local dev and Linux CI. On CI the GITHUB_PATH prepend makes the vendored binary win, so the apt copy is shadowed, not double-installed to any effect. Unifying local dev onto the vendored binary is a deliberate follow-up.
3. Fixed. Add-Content now uses -Encoding utf8NoBOM. Parsed under pwsh locally.
4. Not changed. The idempotency skip is a deliberate guard for local/self-hosted reuse; no workflow caches build/nats/vendored by design.

mcraveiro:
Round addressed in 9721f817.

Not changed (nats-server in _BASELINE). Removing it would break local dev: the generated systemd unit for this environment runs ExecStart=/usr/sbin/nats-server (systemd/units/nats-server-bright_faraday.service), a path only the apt package provides, and compass env install-packages is shared between local dev and Linux CI. The vendored binary shadows the apt copy on CI via the GITHUB_PATH prepend, so the apt install is harmless there. Unifying local dev onto the vendored binary is a deliberate follow-up.

Agreed on the shared composite action; noted as a follow-up, out of scope for this hotfix.

mcraveiro:
Round addressed in 9721f817.

1. Not changed. The doc's a49329949 is correct: git rev-parse --short=11 a49329949 resolves to a49329949bb, so the 9-character short SHA in the doc is a valid, unambiguous prefix of the actual commit. The review's a4932994 drops a digit.
2. Not changed. The single-entry glob matches the real NATS release layout, and a layout change fails fast with a clear shell error, as the review notes.
3. Understood. The change is verified locally (bash -n, pwsh parse, functional smoke run unpacking v2.14.3, idempotency, GITHUB_PATH content) and will be confirmed by the next scheduled runs.

mcraveiro:
Agreed — the line 61 re-spelling of the archive path is a harmless nit, and per your own assessment it is not worth a follow-up commit: any push re-runs the full check set (the site job alone is about 19 minutes) for a cosmetic change. Leaving it as-is; it can ride a future change to these scripts. Thanks for the re-review.

PR #1964, opened 2026-08-11
[marketdata,synthetic,ores.sql] Folder cascade covers every asset class via the feed factory

mcraveiro:
## Summary

The start/stop feeds under folder cascade resolves the folder subtree once and dispatches every feed row beneath it -- fx_spot and ir_curve alike -- through the feed factory to the per-kind controller, fixing the reported defect that the cascade only walked fx_spot_generation_config rows. Responses now carry per-kind outcome counts, and the integration tests prove the whole path over real NATS and the real database: start, repeat (already-running), skip (disabled), stop, permission rejection, and malformed input.

## Changes

- start()/stop() dispatch every config family through make_default_feed_factory(): fx rows via feed_controller, ir_curve rows via curve_feed_controller, with per-kind skip reasons counted in the reply
- start/stop folder responses gain per-kind counts (by_kind started/already_running/skipped; stopped_by_kind)
- The values party-isolation policy's unqualified config_id reference is table-qualified -- the reference bound to the inner config table's own config_id column, so every value row was rejected on write and invisible on read under party isolation
- New folder feed control handler integration tests: real NATS + real database, per-run tenant, requests on test-scoped subjects so the running service's verifier never answers them

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Folder cascade covers every asset class via the factory](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/feed-lifecycle-harmonization/task_folder-cascade-all-asset-classes.html) | 9E2344CF-9A66-4742-9DE6-6E2AFBF8B3B1 |


mcraveiro:
Fixed in c7f9ad53b3 - the FX factory.make()/ctrl_.add() path is wrapped in try/catch; a malformed row counts as skipped with a warn log, mirroring the IR pass.

mcraveiro:
Not changed. The cascade aligns with the auto-start path it replaces; the vintage check stays in the client-supplied-params path until the control-plane task deletes it. Recorded in the task Plan (lines 74-77).

mcraveiro:
Fixed in c7f9ad53b3 - add() now takes a conflicting-source-name out param; an empty conflict (concurrent duplicate) counts as already_running, a named conflict as skipped with a warn log.

mcraveiro:
Not changed. The uniform gate is by design: the cascade now touches both config families, so both read permissions are required; the handler's doc comment states the gate.

mcraveiro:
Fixed in d3f2c370d1 - comment moved above the field.

mcraveiro:
Fixed in 7b657a0543 - add() returns false for an already-running source_name, as start() always did; the duplicate no longer emplace-collides and destroys a joinable thread.

mcraveiro:
Fixed in c7f9ad53b3 - the call site now passes the conflicting-source-name out param and counts the false return: empty conflict as already_running, named conflict as skipped with a warn log, matching the application.cpp auto-start path.

mcraveiro:
Not changed. Accepted in the task Plan: the cascade aligns with the auto-start path it replaces; the vintage check stays in the client-supplied-params path until the config-id control-plane task deletes it.

mcraveiro:
Fixed in d3f2c370d1 - comment moved above the field.

mcraveiro:
Not changed. Story task 4 (uniform-auto-start) plans the unification of both walks.

PR #1969, opened 2026-08-11
[synthetic] Collapse curve_feed_controller into the unified feed_controller

mcraveiro:
## Summary

One lifecycle controller serves every feed kind. The running map keyed by source_name, thread-per-feed, and start/stop/list/running_count/join are unchanged; the IR qualifier+role conflict semantics now flow through the feed's own IFeed::conflict_key() on every registration path (add, IFeed start, and the FX client-supplied-params start), reported with the holding source_name. The FX extras stay FX-shaped: binding auto-creation and vintage availability only in the ad-hoc start() overload, binding_mode stored per running feed purely as that overload's restart gate, and a uniform status loop. Dead code (resolve_series and its members) is dropped.

## Changes

- One feed_controller owns both kinds; curve_feed_controller and its test file deleted
- Uniform conflict rule via conflict_key() on all paths; feeds_conflict helper (renamed from ir_curve_feeds_conflict) gains the empty-qualifier guard and now also rejects two FX feeds on the same pair; conflict reported via running_source_name_for_conflict_key()
- FX extras stay FX-shaped: binding auto-creation + vintage availability only in the ad-hoc start() overload; add() keeps binding_mode (two overloads) purely as that overload's restart gate; status loop uniform (source_name + publish_count)
- Dead code dropped: resolve_series, md_client_, tenant_id_, uuid_gen_; ctor now takes (nats, auth_nats)
- Callers collapse to one ctrl: registrar, application, folder handler (IR pre-check dropped), IR on-demand handler, FX ad-hoc handler (gains qualifier_conflict case); stale comments updated
- Verification: local build clean (linux-clang-debug-make); DB recreated (schema 0.0.25); rat green on the fresh DB; synthetic service component tests 195 assertions in 39 cases

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [One lifecycle controller for all feed kinds](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/feed-lifecycle-harmonization/task_single-feed-controller.html) | 1D756D47-9473-4A99-8A47-E3E4906DA4C3 |


mcraveiro:

mcraveiro:
Fixed in 8245ecdc5f — the folder-cascade test now seeds a second enabled FX config on the same EUR/USD pair (under its own container: the schema forbids two FX configs on the same pair per container) and asserts only one of the two starts while the other is counted as skipped, driving the qualifier_conflict rejection through the real controller path, as suggested. The kind-scoped list filter added alongside (fd5a2155c2) is asserted there too.

mcraveiro:
Both points fixed. 1) FX auto-start: 9e2d0faaa4 — auto_start_enabled_feeds now captures out_conflicting_source_name and logs an error when add() rejects an enabled config whose ore_key is already held, mirroring auto_start_enabled_ir_curve_feeds(); an already-running same-source row (a duplicate config) stays silent, matching the old behaviour. 2) Doc/implementation mismatch: fd5a2155c2 — the class doc now states the check compares the (qualifier, role) pair via feeds_conflict() rather than the key string, and why (the empty-qualifier carve-out is key-string-blind).

mcraveiro:
All four points fixed. 1) list() kind scoping: fd5a2155c2 — IFeed gains kind() (the factory kind string, now defined in each producer header) and feed_controller::list(kind) filters the running snapshot by it; each handler passes its own kind, restoring the per-kind list responses. 2) FX auto-start silent drop: 9e2d0faaa4 — captures out_conflicting_source_name and logs an error on collision, mirroring the IR walk. 3) Class doc contradiction: fd5a2155c2 — the class doc now reads 'at most one per (qualifier, role) pair — two FX feeds included'. 4) Lookup carve-out: fd5a2155c2 — running_source_name_for_conflict_key() skips empty-qualifier feeds, so its string-equality match cannot false-positive on a key that carries no conflict semantics.

PR #1973, opened 2026-08-11
[shell] Feed verbs for every asset class

mcraveiro:
## Summary

(Fill in: what changes, why.)

## Changes

- (Fill in.)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [ores.shell: feed verbs for every asset class](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/feed-lifecycle-harmonization/task_shell-feed-verbs.html) | 26D67D72-1136-4100-BFED-C081246B5ADB |


mcraveiro:
Fixed in 51c097f01e and 0f41482ebb. 1. Success check, fixed: each probe now fails with 'Failed to list feeds: ' plus the backend message instead of falling through as an empty list (51c097f01e). 2. Cross-family ambiguity, not changed: FX-first resolution mirrors the server's own kind resolution (feed_config_handler probes the repositories FX before IR), and the function comment documents that a token matching both families resolves to the feed the server would start -- the server resolves the same config_id to the FX feed, so the FX win is consistent. A cross-family check would add an IR round-trip to every FX match to defend a collision the naming conventions make unreachable. 3. Duplicated blocks, fixed: factored into exact_feed_match and feed_by_id templates over the config vector plus id/name/ore-key accessors; a third asset class is one call per family.

mcraveiro:
Fixed in 51c097f01e and 0f41482ebb. 1. Success check, fixed: each probe now surfaces the backend message ('Failed to list feeds: ' plus message) instead of reading as zero configs (51c097f01e). 2. Duplicated blocks, fixed: exact-match and ambiguity handling factored into an exact_feed_match helper, UUID search into feed_by_id (0f41482ebb). 3. try_uuid parsed twice, fixed: hoisted above both probes in the same refactor.

PR #1976, opened 2026-08-12
[ores.analytics.quant] Widen statistical test tolerances to multi-sigma bounds

mcraveiro:
## Summary

Two Monte Carlo statistical tests in ores.analytics.quant fail on macOS CI: black_karasinski_process_tests.cpp:255-256 (sample variance, got 0.044853 vs 0.045) and heath_jarrow_morton_process_tests.cpp:188 (HJM drift mean, got 0.04733 vs 0.045). Root cause: the tolerances are tighter than the estimators' own standard error (BK variance check about 0.95 sigma, HJM drift check about 0.6 sigma), and std::normal_distribution is implementation-defined, so the same mt19937 seeds produce different draws under libc++ (macOS) than under libstdc++ (Linux). The fix widens the bounds to multi-sigma headroom, matching sibling conventions in the same files; no source changes, so genuine model errors of about 3 sigma or more are still rejected.

## Changes

- Widen the Black-Karasinski sample-variance epsilon from 3e-3 to 1.5e-2 (about 4.7 sigma at 200000 paths, matching the Gaussian-OU moments test in the same file).
- Widen the HJM drift-mean margins from 2e-3 to 1e-2 (at least 3.2 sigma at the noisiest rate, sigma = 0.5; the previous margin was about 0.6 sigma).
- Add comments documenting the sigma headroom and the per-standard-library draw difference.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Hotfix: stochastic process statistical tests fail on macOS CI](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/osx-stochastic-process-test-failures/story.html) | 4A047F18-0AE2-4DDF-9AC9-04AFAAF233AE |
| Task | [Implement Hotfix: stochastic process statistical tests fail on macOS CI](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/osx-stochastic-process-test-failures/task_implement_osx-stochastic-process-test-failures.html) | 3DE5BBB9-13CD-41FF-9F69-855CF34A7FAA |

### DIFF
--- a/doc/agile/versions/v0/sprint_25/osx-stochastic-process-test-failures/story.org
+++ b/doc/agile/versions/v0/sprint_25/osx-stochastic-process-test-failures/story.org
@@ -29,11 +29,11 @@ genuine model errors.
 
 | Field         | Value                                  |
 |---------------+----------------------------------------|
-| State         | STARTED                              |
+| State         | DONE                                 |
 | Parent sprint | [[id:0E7321A7-A66D-4CE6-B3E7-89F2E35CA48B][Sprint 25]] |
-| Now           | Diagnosed: sub-sigma tolerances.       |
+| Now           | Nothing.                               |
 | Waiting on    | Nothing.                               |
-| Next          | Implement and verify the fix.          |
+| Next          | Nothing.                               |
 | Last touched  | 2026-08-12                               |
 
 * Acceptance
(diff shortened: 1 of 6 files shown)


mcraveiro:
Not changed. A drift-formula error scales each leg's error with that leg's volatility (mu_i = sigma_i * integral of sigma), and the standard error of each sample mean scales the same way, so in sigma units the detection sensitivity is already leg-independent: the shared 1e-2 margin sets one common threshold, bounded by the noisiest leg at 3.2 sigma. Per-leg margins would equalize around the same threshold (rate 3 needs 1.26e-2 for 4 sigma) without improving worst-case detection, and 1e-2 matches the file's existing variance-check convention (epsilon 1e-2).

PR #1979, opened 2026-08-12
[marketdata,synthetic,qt] Unified feed ingest: kind in the tick subject, one loop, one stream

mcraveiro:
## Summary

Task 8 of the feed-lifecycle-harmonization story (71B00A0E), closing the story: one ingest loop over a unified tick subject scheme.

What ships:
- Kind as a subject token: producers emit synthetic.v1.tick.<kind>.<source_name> (fx_spot / ir_curve, the factory kind strings) through the new shared wire contract ores.marketdata.api/domain/tick_subjects.hpp; the curve subject family and synthetic_curve_ticks stream are retired (stop-ensuring; no delete-stream API, a stale stream is inert).
- One loop: curve_feed_ingest_loop deleted; feed_ingest_loop is the single loop with one synthetic.v1.tick wildcard and kind-token dispatch. The per-source subscribe/unsubscribe lifecycle became a binding cache rebuilt by refresh() (still driven by feed_binding_changed_event); unbound fx_spot ticks are dropped with a one-time warn — the old structural unreachability, now explicit.
- Uniform republish: FX semantics applied to every kind (raw payload to the lowercased marketdata.v1.tick.<tenant>.<series> subject), no per-kind flag; CRM stays FX-only.
- Consumers unchanged: payloads and the republished scheme are untouched, so curve_republish_resolver point-id lookups and the Qt tick chart keep working; MarketSimulatorWindow::synthetic_subject() now emits the kind-prefixed subject via the shared tokens.

Verification (recorded in the task's Result): full build green (exit 0, 100%); no SQL change, so no DB recreate; compass build rat green (exit 0) after the documented iam CD-residue cleanup.

## Changes

- (Fill in.)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [One feed ingest loop over a unified subject scheme](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/feed-lifecycle-harmonization/task_unified-feed-ingest.html) | F550327A-001F-460F-A490-4EBF94642F31 |


mcraveiro:
Addressed in 354538356a.

1. Fixed. The ir_curve republish (and the fx_spot one, for uniformity) is now gated on the persist helper returning true; a tick that failed to persist is not republished, so the republished stream cannot diverge from the observations table.
2. Fixed. Stats are keyed by (kind, source_name); a cross-kind source-name collision no longer merges counters or the write-once series identity.
3. Fixed. refresh() prunes the stats of fx_spot sources that left the bindings; ir_curve entries have no binding lifecycle and stay.
4. Fixed. One uuid generator member, reused across ticks. The loop's single subscription dispatches callbacks serially, so a single generator is safe.
5. Fixed. The algorithm include is added.
6. Fixed. The two persist paths are one persist_tick_observation() helper; asset_class now derives uniformly from the lowercased series_type (FX to fx, RATES to rates), so the cited drift is gone.
7. Fixed. The tick_subjects.hpp doc now states the dispatch reality (one ingest branch per kind; producers reuse the factory vocabulary), and the tests tie fx_spot_feed_kind and ir_curve_feed_kind to the marketdata kind tokens with static_asserts plus an ir_curve subject-shape check. A kind-to-handler registry is not built: ores.marketdata.service cannot depend on ores.synthetic.api, so the dispatch stays an explicit branch.
8. Not changed. isIrCurveSourceName() is a pre-existing client-side resolution (refreshTickChart has only the source_name in scope; the client has no kind on the wire). A cross-kind source-name collision is a seed-data concern, not enforced anywhere today, and this PR does not change the resolution mechanism.
9. Fixed. The class doc states the current design without edit-history narration.

PR #1988, opened 2026-08-17
[ores.marketdata] Export IFeed default ctor and virtual dtor out-of-line

mcraveiro:
## Summary

Fixes the Windows build: `ores.synthetic.api.dll` fails to link in all three configs (clang-debug, msvc-debug, msvc-release) with undefined symbols `__declspec(dllimport) ores::marketdata::domain::IFeed::IFeed()` and `~IFeed()`, referenced from the derived feed destructors (`feed_factory.cpp`, `fx_spot_feed.cpp`). Windows has been red for 8 consecutive scheduled runs.

**Root cause**: `IFeed` is `ORES_MARKETDATA_API_EXPORT` (dllimport in consumers) with inline-defaulted default ctor and virtual dtor in `i_feed.hpp`. The class is abstract and never constructed or destroyed inside the producer, so the producer never emits or exports the special members; consumers import them.

**Fix**: define both out-of-line in the producer:
- `i_feed.hpp`: `IFeed();` and `virtual ~IFeed();` declarations (no inline defaults)
- New `domain/i_feed.cpp`: `IFeed::IFeed() = default;` and `IFeed::~IFeed() = default;` — the class-level export macro exports them from `ores.marketdata.api.dll`
- Registered in `component_files.cmake`

Consumers (ores.synthetic) are unchanged.

## Verification

- `ores.marketdata.api.lib` builds; `nm` shows `IFeed::IFeed()` and both `~IFeed()` variants exported (T) from `libores.marketdata.api.so`
- `ores.synthetic.api.lib` (the Windows-failing target) builds and links
- `ores.synthetic.api.tests` pass (incl. `feed_factory_tests`, which construct/destroy the derived feeds)
- `ores.synthetic.service.tests` pass
- Note: an initial `ores.synthetic.api.tests` run showed unrelated `std::bad_alloc` failures from a stale test binary (ABI mismatch with the rebuilt .so after the FOMC merge); rebuilding the test binary makes them pass.

Windows-side export behavior cannot be exercised on Linux; the scheduled Continuous Windows run after merge is the confirmation.


mcraveiro:
Not changed. Agreed it is a genuine same-shape follow-up candidate, but it is out of this hotfix's scope: the in-tree Windows build is unaffected (PluginBase.cpp defines its ctor out-of-line in the producer, odr-using IPlugin's implicit ctor, and nothing in-tree destroys IPlugin instances — plugins are owned by QPluginLoader), and the failing target is only ores.synthetic.api.dll. Tracked in the task Notes as a follow-up for a separate change if it ever manifests.

PR #1992, opened 2026-08-25
[ores.shell] Add compute grid management verbs to the shell

mcraveiro:
## Summary

The shell's `compute` menu gains the full compute grid management layer: five read verbs (`list-apps`, `list-app-versions`, `list-platforms`, `list-hosts`, `list-batches`), the job lifecycle (`add-batch`, `dispatch-batch`, `list-workunits`, `list-results`), grid monitoring (`grid-stats` with `--watch`/`--smoke`, `delete-host`), artifact download (`download-input`, `download-output`) and the grid smoke-test script. Every verb reuses existing NATS subjects and the codegen table renderers in `ores.compute.api`; no backend or codegen changes.

This branch carries the five tasks of the "Manage the compute grid from the shell" story (read verbs, job lifecycle, monitoring, artifact download, smoke script — all closed DONE with this PR; the reconciliation handoff remains BLOCKED on the ores.compute drift story).

## Changes

- **Read verbs** — `list-apps`, `list-app-versions`, `list-platforms`, `list-hosts`, `list-batches` render the app catalogue, versions, platforms (provisional raw rows until the drift story lands the generated `table_io`), hosts and batches as tables; empty domains render a header-only table, not an error.
- **Job lifecycle** — `add-batch` creates a batch (with `--smoke` bounding the job count to 10..20). `dispatch-batch` uploads the input `tar.gz` tarball verbatim as the batch's shared input bundle (matching the Qt UI convention of uploading the file as-is; directory packing was removed from the shell), then saves one workunit per job — each save dispatches to the grid (the backend creates the result rows and publishes the JetStream assignments). The batch is marked "dispatched" only after every workunit save succeeds, so a partial failure leaves it open and re-dispatchable; each new workunit sets the nil `canonical_result_id` sentinel. `list-workunits` and `list-results` render the tables, optionally filtered to one batch.
- **Grid monitoring** — `grid-stats` renders the grid summary and per-node rows; `--watch` polls until the batch drains; `--smoke` asserts every result has outcome Success and every host online at the drain transition has at least one result (the host set is captured at the drain transition, and the online window matches the server's grid-stats definition). `delete-host` removes an offline host row and refuses an online host.
- **Artifact download** — `download-input` and `download-output` fetch a workunit's input bundle or a result's output bundle from compute storage and unpack the tar.gz into a target directory.
- **Smoke script** — a library `.ores` script (tangled from `how_do_i_smoke_test_the_compute_grid_from_the_shell`) drives the whole lifecycle: add a batch with `--smoke`, dispatch the jobs, watch `grid-stats --watch --smoke` until the batch drains.
- **Recipes** — one recipe per verb under `doc/recipes/shell/compute/`, tangled into `projects/ores.shell/scripts/library/compute/`, indexed under the new `Compute` section of `doc/recipes/shell/shell.org`.

## Recent changes on this branch

- **NATS subject logging** — the NATS client now logs connect, publish, request, subscribe, queue-subscribe, js-publish, js-subscribe and js-queue-subscribe with the full prefixed subject, durable and queue group (the publish log is throttled: first subject at info, later at trace). Services drop their own redundant "Connected to NATS" lines, and `remote-run.sh` mounts the wrapper log directory and enables wrapper logging so remote nodes log to files instead of container stdout only. Used to verify stream, durable, queue-group and filter-subject naming end-to-end on a live run.
- **`dispatch-batch` tarball input** — the command now requires a `tar.gz` tarball and uploads it verbatim; directory packing was removed from the shell (`pack_and_upload` remains in `ores.storage` for its other users). The dispatch and smoke recipes document the tarball input.
- **Capture** — consolidated the two duplicated libarchive archivers into one shared utility (inbox capture for future on-the-fly tarball creation); `.gitignore` now covers the `.compass.db*` sidecar files.

## Verification

- Local build clean (preset `linux-clang-debug-make`); `ores.shell.tests` 26/26 and `ores.cli.tests` 125/125 test cases passed (via `compass build test_ores.shell.tests test_ores.cli.tests`).
- All verbs proven live against brave_hopper: read verbs returned 2 apps, 2 app versions, 6 platforms, 6 hosts and 1 batch; a 10-job batch was created, dispatched and drained in ~2 seconds with all results Success; input and output bundles round-tripped byte-for-byte.
- **Smoke test fully green live** — three consecutive smoke runs passed on 2026-08-25/26 with the 6-host grid (including the remote Newton node): 15/15 results Success, 6/6 online hosts exercised each time, with the tarball passed directly to `dispatch-batch`.

## Traceability

Story: Manage the compute grid from the shell (36E9BEDD-4804-4822-AFD8-8AF961978398)

| Task | ID |
|------+----|
| Add compute read verbs to the shell | 7FB32716-261C-45D9-A5C9-6A43FF2AD6F5 |
| Add job lifecycle verbs to the shell | 05D09703-6899-43A1-A9F8-997AE79D6436 |
| Add grid monitoring verbs to the shell | 90383934-D78D-41A2-88C3-FF369CB01830 |
| Add artifact download verbs to the shell | 5908A274-E089-4D2B-9982-68ED6F779B9B |
| Add a compute grid smoke-test script and prove it live | 292AB16F-3785-4D1F-AEAC-BF6B59390078 |


mcraveiro:

mcraveiro:
Findings addressed:
1. Fixed — wu.canonical_result_id = boost::uuids::uuid{} set alongside the other field assignments; the mapper stores nil as NULL (dddb3ca710).
2. Fixed — the batch is marked dispatched only after every workunit save succeeds; a partial failure leaves the batch open and re-dispatchable (dddb3ca710).
3. Not changed — the 1000-row limit matches the existing convention in the crm/provision/accounts commands, and reviewer 2 explicitly checked it as not a concern. Tracked as a follow-up.
4. Partially addressed — a comment now documents the batch-scoped key convention (dispatch uploads once per batch; download-input reverses it), since compute_storage::input_key is workunit-scoped and does not apply. A batch-scoped helper is tracked as a follow-up.
5. Fixed — the online host set is now captured at the drain transition (67406e0ea2).

mcraveiro:
1. Fixed — the batch status is saved as dispatched only after the workunit loop completes; a partial failure leaves the batch open so dispatch-batch can be re-run (dddb3ca710). One caveat: re-running after a partial failure may duplicate the workunits already created before the failure — inherent to both approaches; the shared input bundle is keyed by batch, so it is uploaded once regardless.
2. Fixed — the Traceability table now cites the read-verbs task (7FB32716-261C-45D9-A5C9-6A43FF2AD6F5) via gh pr edit.
Thanks for the not-a-concern checks — they match our decisions (limit stays as-is; no new unit tests matches this layer's convention).

mcraveiro:
Fixed — the process_list_platforms doc comment now states it prints provisional raw rows (id, code, display name, OS, CPU arch) until the drift story lands the generated platform table_io (01ad6a94e3). And thanks for the observation on the broader diff — intentional: the branch carries the sibling tasks of the same story.

mcraveiro:
1. Fixed — the batch is marked dispatched only after every workunit has been saved; a mid-loop failure leaves it open and re-dispatchable (dddb3ca710).
2. Fixed — is_online now uses a 300s window matching ores_compute_grid_stats_fn_create.sql, so the smoke assertion and the delete-host guard agree with the server's online_hosts count (67406e0ea2).
3. Partially addressed — the comment now documents the batch-scoped key convention; compute_storage::input_key is workunit-scoped, so it does not apply to the shared bundle. A batch-scoped helper is a follow-up.
4. Fixed — results_of_batch builds a std::unordered_set of workunit ids, O(n+m) instead of O(n×m) (01ad6a94e3).

mcraveiro:
1. Not changed — re-running dispatch-batch after a partial failure re-mints job_count new workunits rather than only the missing ones; this is the inherent trade-off of the leave-open-and-redispatch design, acknowledged in my earlier reply. An operator can create a fresh batch instead of retrying a partially-dispatched one. 2. Not changed — the 1000-row limit matches the existing convention in the crm/provision/accounts commands; tracked as a follow-up (paginate_compute_list_requests.org). 3. Fixed — the delete-host docstring now says 300s, matching online_window{300} (0dc9f4a13b).

mcraveiro:
Fixed — the stale 90s heartbeat text in compute_commands.hpp:186 and the delete-host recipe now state 300 seconds (5 minutes), matching online_window{300} (0dc9f4a13b). Thanks for re-verifying the five earlier findings as fixed.

mcraveiro:
1. Fixed — split_storage_uri now derives its prefix from storage_paths::prefix, so download-input/download-output stay in sync with make_object_path if the API path ever changes (58761e0ffa). 2. Fixed — the monitoring task doc's Acceptance and Plan sections now state 300 seconds (5 minutes), consistent with its Result section (0dc9f4a13b).

mcraveiro:
1. Fixed — all three stale 90s references (header doc, delete-host recipe, monitoring task doc Acceptance/Plan) now state 300 seconds (5 minutes), matching online_window{300} (0dc9f4a13b). 2. Fixed — split_storage_uri now derives its prefix from storage_paths::prefix, so no backlog capture was needed (58761e0ffa). 3. Fixed — the smoke and dispatch recipes now name the variable INPUT_TARBALL, retangled into the .ores scripts (bb0b13bcd7).

PR #1998, opened 2026-08-28
[ores.eventing,ores.database] Harden the eventing chain against transient notification loss

mcraveiro:
## Summary

Hardens the eventing chain so a notification is either delivered or its loss is reported. The listener reconnect window (consume_input failure, backoff, re-LISTEN) drops NOTIFYs in flight with no replay possible, and NATS publish failures were caught and swallowed end to end.

## Changes

- Listener: each detected connection loss increments connection_loss_count and logs the affected channels at error; the successful reconnect logs the outage duration and channels.
- publish_entity_event rethrows NATS failures with the subject in the message; the event_bus delivery summary logs at error when delivery is incomplete.
- Tests: listener reconnect test kills the session server-side and proves loss surfacing and post-reconnect delivery; publisher test proves a NATS_NOT_CONNECTED publish throws. Fixed invalid DO $ blocks in the listener tests (de-vacuated connect_forces_utc_session).
- Verification: full build; ores.database.tests 17/17; eventing api/core suites; eventing integration family (85 tests); full ctest 71/71 on a freshly recreated database.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Harden the eventing chain against transient notification loss](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/eventing_chain_hardening/story.html) | 639008BB-03A6-4675-AF2F-90A4A699729C |
| Task | [Implement eventing chain hardening](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/eventing_chain_hardening/task_implement_eventing_chain_hardening.html) | C9B8F903-27A4-456E-90E7-157507FFC024 |

### DIFF
--- a/doc/agile/product_backlog/inbox/limit_concurrent_running_fleets.org
+++ b/doc/agile/product_backlog/inbox/limit_concurrent_running_fleets.org
@@ -1,0 +1,48 @@
+:PROPERTIES:
+:ID: BF8714BC-B214-448E-A16C-66EB36C967B1
+:END:
+#+title: Limit concurrent running fleets via fleet slots
+#+description: Bound concurrent fleets to a fixed slot count so service fleets cannot exhaust host memory.
+#+type: capture
+#+level: cross
+#+filetags: :compass:memory:host:
+#+created: 2026-08-28
+#+updated: 2026-08-28
+
+This page is a [[id:671F18E4-E09C-4B3B-BD24-D33DF8AE38A6][capture]] in the *inbox* bucket of the product backlog — a pre-sprint idea, not yet pulled into a sprint as a story.
+
+* What
+
+Add a host-wide fleet-slot limit to =compass services start= so only a
+fixed number of environments can run their service fleets at once.
+The mechanism counts active systemd user units matching
+=ores-<env>.target= (e.g. =ores-brave_hopper.target=,
+=ores-jolly_knuth.target=) before starting a new fleet, and refuses
+with a message naming the running fleets when the limit is reached.
+The check lives in =compass_services.py=, in =_cmd_start=, before
+=systemd_generate.cmd_generate=; the count is a constant in the style
+of =BUILD_LOCK_SLOTS= in =compass.py=. A count-based check (not a
+flock like the build lock) because a fleet outlives the =services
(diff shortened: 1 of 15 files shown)


mcraveiro:
Fixed in f7c789c5e5 and 9e351613dc. Each service application.cpp now delegates its local publish_entity_event to the shared ores::eventing::service::publish_entity_event, which rethrows — so the event_bus delivery summary fires at error for hand-registered entities too. The now-unused pub_lg() loggers and wire_codec includes are dropped. The hand-rolled comma-join is replaced with boost::algorithm::join. Verified locally: full build clean, ores.database.tests green.

mcraveiro:
Fixed in b20138f7cf — the kill is now scoped to the listener's own backend. connect_utc stamps the session's application_name with its backend pid via set_config (ores.database.listener dot pid), and the test's pg_terminate_backend matches only backends carrying that name. On the suggested mechanism: sqlgen's Connection has no scalar-query API (execute returns Result of Nothing; read needs a dynamic SelectFrom), so the pid is exposed through the stamped name rather than a backend_pid accessor — the name embeds the exact pid, and only reconnect-capable listener sessions of this user can match. The two minor notes are acknowledged and intentionally unchanged: wait_for's extra predicate eval is harmless with the idempotent-read predicates used here, and the reconnect message logs at error deliberately — that window lost notifications, and the error record is the loss surfacing.

mcraveiro:
Fixed in 4f7e000d0f. Each listener instance now stamps a per-instance unique application_name (process pid plus an instance counter), generated in the constructor and exposed via application_name(); the test's backend_filter matches that exact name with equality, so it can never match a sibling listener from a concurrently-running suite — only this test's own backend. On the suggested pid accessor: sqlgen's Connection cannot read scalar results (execute returns Result of Nothing and read requires a dynamic SelectFrom), so the backend pid itself cannot be surfaced from the service; the unique name scopes pg_terminate_backend to the same single backend, which is the property the finding requires. Verified locally: full build clean, ores.database.tests green. On the unrelated capture file: acknowledged — it is an agile backlog capture from the working session and rides harmlessly on the topic branch.

PR #1999, opened 2026-08-31
[ores.platform] Add process::current_pid and migrate getpid call sites

mcraveiro:
## Summary

The Windows CI build broke at postgres_listener_service.cpp:27 with
C1083 (cannot open include file 'unistd.h') after PR #1998 introduced
the POSIX-only include together with `::getpid()`.

Instead of copying the ad-hoc `#ifdef getpid/_getpid` pattern a third
time, this PR puts the portability in ores.platform, the component that
already owns this problem class (setenv/unsetenv live in
ores.platform::environment).

## Changes

- **ores.platform::process::current_pid()** — new process/pid module
  wrapping `boost::process::v2::current_pid()` (Boost.Process v2 is
  already a private link dependency of ores.platform). Returns a
  stable int64_t so callers are not tied to per-platform pid types.
  Unit tests included.
- **Migrate call sites** — postgres_listener_service.cpp (the break)
  and test_database_manager.cpp (the original ad-hoc pattern) now call
  the wrapper; both `#ifdef` patterns are retired.
- **Retire the duplicate** — the misplaced
  `ores::platform::net::get_process_id()` had the identical
  boost-backed implementation. It and its tests are removed;
  ores.telemetry.core resource now uses `process::current_pid()`.

## Verification

- ores.platform.lib, ores.database.lib, ores.testing.lib and
  ores.telemetry.core.lib build and link on linux-clang-debug-make.
- ores.platform.tests: 201 assertions in 88 test cases, all pass.
- ores.telemetry.core.tests: 2133 assertions in 57 test cases, all
  pass.
- The Windows CI build now compiles postgres_listener_service.cpp —
  boost::process::v2 is the cross-platform implementation.

Closes the Windows build break from #1998.


mcraveiro:
Fixed — the rework of this PR replaced the duplicated getpid/_getpid ifdef blocks with ores::platform::process::current_pid(), a boost::process::v2-backed wrapper. Neither file carries an ad-hoc block anymore.

mcraveiro:
Fixed in a1aaa61d0a — the story Status block is now DONE with Now/Waiting on/Next set to Nothing, and the sprint.org row is DONE, following the windows-ifeed-export-hotfix pattern.

mcraveiro:
Fixed in af31e35572 — lifecycle_manager.cpp now calls ores::platform::process::current_pid(); ores.logging already links ores.platform.lib, so no dependency change was needed.

PR #2000, opened 2026-09-03
[ores.database,ores.synthetic,ores.qt] Black-Karasinski config test: pool backoff, data-driven parameter labels, editor and tree fixes

mcraveiro:
## Summary

Feature-test work for the Black-Karasinski log-normal short-rate engine in the IR curve config editor (scenario 795B5273). The test run drove out the changes in this PR: a DB pool acquisition backoff that makes the composite config save reliable under contention, data-driven process-parameter labels (English name, Greek symbol, short label) seeded for all parameter definitions, and a set of IR curve editor and Market Simulator tree fixes. The agile captures for the scenario and for the deferred IR/FX client cleanup are included.

## Changes

- ores.database: configurable pool acquisition backoff in common code (tenant_aware_pool). Exponential is the default: 1 s base, 10 s cap, per-thread jitter so waiters do not re-probe in lockstep; the retry limit is kept. Linear backoff remains available via `pool_acquire_policy` and the `--db-pool-backoff <exponential|linear>` option. The external sqlgen pool runs in probe mode; the project-side loop owns the wait policy. Documented in postgresql_architecture.org.
- ores.synthetic / ores.sql: three label columns on yield_curve_process_parameter_definitions — display_name, symbol (nullable), short_label — added to the org model and regenerated through codegen (domain, protocol, repository, Qt model, schema DDL), with hand-written idempotent populate SQL seeding all 23 rows. Parameter definitions resolve at the system tenant.
- ores.qt: the IR curve editor's parameter rows are data-driven from the definitions table — symbol plus English name in Advanced mode, short layperson labels in Simple mode, full description as tooltip. The segmented Simple/Advanced toggle is restored, Curve Template rows seed the refdata codes (DEPO/FRA/IRS), both parameter tables hold a fixed 40/60 label/value ratio re-applied on resize, and the chart and save paths fall back to the stored frequency code while the frequency combo still populates from reference data.
- ores.qt: the Market Simulator tree shows editor-created IR curves under folder-less feeds (buildTree's fallback appended FX pairs only).
- agile: BK feature test scenario recorded and closed (FAILED); asset-class-agnostic Market Simulator client debt captured as story E95A3D20.

## Verification

Full build exit 0; services restarted on the eager_maxwell environment; database recreated 2026-08-31 (schema 0.0.25, 275 tables, zero validation warnings); Acme Corporation holding group provisioned.

Scenario steps 1-5 PASS: the Black-Karasinski engine appears in the engine combo, its four parameters show the seeded defaults and bounds, Simple and Advanced labels render as described, and the live curve preview draws through the mapping layer.

Scenario step 6 FAILED, and the work is closed as FAILED: the sigma 0.02 save completes with no timeout (~200 ms with the backoff fix, previously timing out at the 30 s NATS budget) and the config row persists, but a reload does not show the new config in the tree. Steps 7-9 were not executable (reopen needs the tree item; IR curve tree delete is broken client-side — the delete dispatch always sends the FX delete request, which no-ops silently). The failing paths are pre-existing IR/FX client defects, tracked in story E95A3D20 rather than fixed here.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Add new stochastic processes](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/add_new_stochastic_processes/story.html) | C603D702-DE7E-4DA4-8409-6FCB0F64808F |
| Task | [Implement Black-Karasinski log-normal short-rate process](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/add_new_stochastic_processes/task_implement_black_karasinski_process.html) | CDB12DB0-B0FF-4507-B94F-17540F6792B3 |
| Scenario | [Verify Black-Karasinski IR curve config](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/add_new_stochastic_processes/scenario_verify_black_karasinski_ir_curve_config.html) | 795B5273-58CF-40BF-9F01-E83CF82F1891 |
| Story | [Asset-class-agnostic Market Simulator client: shared IR/FX config UX](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/asset-class-agnostic-simulator-client/story.html) | E95A3D20-7BE2-4126-89F6-896F0F965C90 |


mcraveiro:
Round reviewed against the full diff. Dispositions:

1. Pool-rebuild path lost its retry/backoff — Fixed in 91d0fde83c. The rebuild's make_connection_pool and the acquire that follows it now run under the same pool_acquire_policy backoff loop as the initial acquire (factored into acquire_with_backoff()); the sqlgen pool stays in probe mode since the project-side loop owns the wait policy. A database still restarting is now retried with backoff instead of failing the request on the first refused connection.

2. Simple-mode slider math duplicates FxSpotRateEditor — Not changed. Unifying the two slider mappings is exactly the consolidation story E95A3D20 (captured in this PR) covers; the 1e6 fixed-point scheme with clamping is deliberate and documented above the functions.

3. New code grows the duplication its own story flags — Not changed. Same E95A3D20 scope: the Simple/Advanced surface is that story's consolidation target.

4. End-of-line comments — Fixed in feb2048e0a. All five moved above their statements.

5. Redundant Simple-table rebuild on toggle — Not changed. Minor widget churn confined to the Simple/Advanced toggle surface E95A3D20 rewrites into the shared editor base; no data or correctness impact.

6. Table rename without migration — Not changed. The rename is deliberate (d933f15284: the conventional long name pushes tenant-scoped derived object names past PostgreSQL's 63-byte identifier limit — noted in the org model). No non-throwaway environment exists: every worktree DB is recreated from the branch DDL (compass db recreate); eager_maxwell already carries the new name.

Minor 1 (-Wreorder) — Fixed in a46cc8cb76. Init lists now follow declaration order.
Minor 2 (permission heads-up) — Not changed. Confirmed pre-existing (your second pass verified this PR leaves it untouched); acknowledged as a follow-up.

Round recorded in the task's * Review table (eb6cb08c68, same push).

mcraveiro:
Round 2 reviewed against the full diff. Dispositions:

1. Simple-mode vintage lock does not disable the row — Not changed. The Simple/Advanced parameter page is exactly the surface story E95A3D20 (captured in this PR) consolidates into a shared editor base keyed on parameter_name; the defect is recorded there (Motivation, 2026-09-04) so the data-driven base fixes it in the right layer. The BK scenario never exercises the vintage price-source lock.

2. Initial pool creation at boot lost its retry budget — Not changed. sqlgen's ConnectionPoolConfig is consulted only by acquire(); ConnectionPool's constructor creates 'size' connections in a plain loop with no retry (ConnectionPool.hpp:30-38) and connection creation receives only the credentials — the attempt/wait values never reach it. The pre-PR boot tolerance therefore never existed: a refused connection failed the bootstrap identically before and after this PR, and the systemd restart policy is unchanged. Probe-mode config is a no-op for boot semantics; only the project-side policy loop (added around acquire and the rebuild) retries.

3. Rebuild holds the reconnect mutex across the backoff budget, no already-rebuilt check — Fixed in 6472e0ce78. The mutex now guards each make attempt and the pool swap only; the backoff waits happen outside it, so threads queued behind a rebuild are not blocked for the whole budget. A pool generation counter, bumped on each swap under the mutex, provides the already-rebuilt check: a thread that detected a dead connection skips re-making when the pool it probed is no longer current.

4. Rebuild retry loop duplicates acquire_with_backoff() — Fixed in 6472e0ce78. One retry_probe() helper owns the probe/backoff/log cadence; the acquire path probes pool_.acquire() and the rebuild probes make_connection_pool under the mutex. Finding 3's already-rebuilt check is applied once, in the helper's caller.

5. backoff_delay() reimplements retry_strategy — Not changed. retry_strategy is a stateful restart calculator (on_start/on_failure, reset_threshold, seconds granularity) with no linear mode and no jitter; backoff_delay is a stateless per-attempt delay in milliseconds with deterministic per-thread jitter. Adding a jitter option does not bridge the stateful/stateless or seconds/milliseconds gaps — the contracts do not map.

6. Duplicated widget-sync blocks — Not changed. Same class as the round-1 declines (#2/#3/#5, not re-raised this round): the mirroring code sits on the Simple/Advanced surface E95A3D20 rewrites into the shared editor base.

Also: the round-1 re-acquire-first step was removed in 6472e0ce78 — sqlgen sessions co-own their connection and in-use flag (rfl::Ref over shared_ptr), so replacing the pool object is safe while the dead session is outstanding; the step was also the one pool access in the rebuild path outside the mutex.

Round recorded in the task's * Review table (56dea1f3ab, same push).

mcraveiro:
Round 3 reviewed against the full diff. Dispositions:

1. Rebuild retries no longer log the actual per-attempt error — Fixed in 684c601163. retry_probe() gained an on_failure hook: the rebuild probe passes a callback logging each failed attempt's real sqlgen error ('Pool rebuild attempt K of N failed: ...') before the backoff wait, while the acquire probe passes a no-op. Operators see the actual cause as it happens, not only after the full budget is exhausted.

2. acquire_with_backoff() discards the real sqlgen error on exhaustion — Fixed in 684c601163. The function now returns the last probe's outcome directly — whether a success or the error that exhausted the budget — matching its doc comment; the synthesized fallback is gone.

Round recorded in the task's * Review table (74c4b6d251, same push).

PR #2005, opened 2026-09-04
[ores.codegen] Protocol/service templates iterate primary_key.columns for key gates

mcraveiro:
## Summary

cpp_protocol.hpp.mustache and cpp_service.cpp.mustache gated their delete/history/save is_uuid branches on primary_key.is_uuid alone, which reflects only the first key column. A compound key whose leading column is uuid would silently drop every trailing key column from the request structs and the save validations. The org sources now loop primary_key.columns with the per-column is_uuid form, matching the member-name derivation (ids for a uuid column, {{column}}s for a text column, is_nil() vs empty() checks), so no key column can be dropped. The {{#entity}} and {{#junction}} sections are untouched: schema primary keys carry no columns list and the junction section has no key gates.

## Changes

- Protocol delete/history and service save-single output is byte-identical for every entity
- Regenerated output confined to uuid-keyed *_service.cpp batch-save loops gaining braces (37 files); three full regenerations idempotent (runs 2-3 byte-identical)
- Verification: ci+code+docs; site build clean; full build clean (linux-clang-debug-make); ctest 71/71 passed on recreated DB; drift checks (refdata, reporting, marketdata) clean except three recorded pre-existing files; ore domain roundtrip report unchanged from recorded pre-existing state

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Fix codegen template drift](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/story.html) | F5532C5C-7A8D-461D-8E0A-CACB602AB3DF |
| Task | [Compound-key is_uuid gating drops trailing columns in Qt/service templates](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/task_compound-key-is-uuid-gating-drops-trailing-columns.html) | 453FA225-F6FF-4095-AFB7-4D59E3FDB357 |

### DIFF
--- a/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/story.org
+++ b/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/story.org
@@ -41,9 +41,9 @@ not waste the sweep.
 |---------------+----------------------------------------|
 | State         | STARTED                              |
 | Parent sprint | [[id:0E7321A7-A66D-4CE6-B3E7-89F2E35CA48B][Sprint 25]] |
-| Now           | Task 3F3559A2 (history mapper provenance fields) in progress. |
+| Now           | Task 453FA225 (compound-key is_uuid gating) done, delivered in commit 28d36860dc. |
 | Waiting on    | Nothing.                               |
-| Next          | Remaining three template-bug tasks, in story order. |
+| Next          | Remaining two template-bug tasks, in story order. |
 | Last touched  | 2026-09-04                               |
 
 * Acceptance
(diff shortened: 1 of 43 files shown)


mcraveiro:

mcraveiro:
Review round dispositions (both reviews: no correctness issues found):

1. Regression-test observation (no fixture exercises a uuid-leading compound key): Not changed in this PR. The 122-entity key census records no such shape today, and a synthetic codegen fixture would widen this fix beyond its scope. Kept as durable input for the codegen test suite; the task's Plan documents the fix's evidence so the shape is reconstructible.

2. Scalar primary_key.is_uuid gates in other templates (history mapper, nats eventing, generator/domain-type): Confirmed, no action. Those templates never iterate primary_key.columns and handle only the single-column case, so they are a pre-existing compound-key limitation, not a regression of this change. The cpp_nats_* templates named are part of the messaging/eventing surface owned by this story's junction follow-up task (D8301A1C); the observation is recorded there as input.

PR #2006, opened 2026-09-04
[ores.codegen] Qt client model fetches paginated unconditionally

mcraveiro:
## Summary

The Qt client model template gates real offset/limit pagination on the :has_pagination: facet flag, which defaults to off with no signal: an entity with the flag unset sends the request with its defaults (offset=0, limit=100), fetches the server's default page, and reports the page's size as the total, silently truncating any list past 100 rows behind fully functional-looking pagination controls. This deletes the fork: the org archetype and the tangled mustache generate the paginated fetch unconditionally, so the request always carries offset/limit and the model always takes total_available_count from the response. The false branch never was a small-table optimisation — it fetched at most the same default page, differing only in whether the model can page onward and report a truthful total — so nothing is lost, and the flag survives only as entity-org documentation. Two regenerated flag-off models are carried (ClientBadgeSeverityModel, ClientReportInstanceModel): entities the narrow-fix census (tracking task 914B206F) missed, now repaired at source.

## Changes

- Delete the has_pagination fork in ores.cpp.qt.client_model_impl.org and cpp_qt_client_model.cpp.mustache: the offset/limit pair and the response-total return are unconditional; the {{^}} truncating branch (defaults, page size reported as total) is gone; the narrative retitles to 'Fetching a page' and records the fork's history.
- Regenerated ClientBadgeSeverityModel.cpp (ores.qt/dq) and ClientReportInstanceModel.cpp (ores.qt/reporting): pure flips from the truncating shape to the paginated shape — both were flag-off entities the 914B206F census missed.
- Verification: code+ci+docs; site build clean; full build clean (linux-clang-debug-make); ctest 71/71 passed against a recreated database; re-tangle changes only the two template files; component drift checks (refdata/reporting/marketdata) reproduce the committed tree except the 3 recorded pre-existing drift files (reverted after the check); regen evidence — pre-edit baseline drift (202 tracked files, recorded composition) reverted or excluded, attributable delta exactly 7 files (2 templates, 2 stub-byproduct cmake reverted, 3 model flips), runs 2-3 byte-identical (idempotent); ore domain roundtrip report clean (rc 0), structurally unchanged.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Fix codegen template drift](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/story.html) | F5532C5C-7A8D-461D-8E0A-CACB602AB3DF |
| Task | [has_pagination knob is a silent-failure footgun in the Qt client model template](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/task_pagination-knob-fragility.html) | 9FF6FAD3-A3C3-4C31-9220-57CAADFDCCA0 |

### DIFF
--- a/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/story.org
+++ b/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/story.org
@@ -103,7 +103,14 @@ not waste the sweep.
   entities — the dq and reporting flips are carried in this PR, and
   the scheduler one (=ClientJobDefinitionModel=) waits for its
   component's drift story because its regeneration bundles unrelated
-  stale drift.
+  stale drift. Review round 1 (PR #2006) completed that record: the
+  compute→reporting and dq→api migrations left stale duplicate
+  client models compiled under =ores.qt/compute/src= and
+  =ores.qt/api/src= that no generator writes to (their deletion
+  belongs to the migration follow-up), and the workspace model is
+  unregenerable until 836BB4CE. The round also retired the
+  =variability_feature_has_pagination.org= feature page, which still
+  described the flag as a live behavioural toggle.
 
 * Out of scope
 
--- a/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/story.org
+++ b/doc/agile/versions/v0/sprint_25/fix-codegen-template-drift/story.org
@@ -41,9 +41,9 @@ not waste the sweep.
 |---------------+----------------------------------------|
 | State         | STARTED                              |
 | Parent sprint | [[id:0E7321A7-A66D-4CE6-B3E7-89F2E35CA48B][Sprint 25]] |
-| Now           | Task 453FA225 (compound-key is_uuid gating) done, delivered in commit 28d36860dc. |
+| Now           | Task 9FF6FAD3 (has_pagination knob footgun) done, delivered in commit 7287a46527. |
 | Waiting on    | Nothing.                               |
-| Next          | Remaining two template-bug tasks, in story order. |
+| Next          | Task D8301A1C (junction codegen template follow-ups), the last in the story. |
 | Last touched  | 2026-09-04                               |
 
 * Acceptance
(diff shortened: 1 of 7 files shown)


mcraveiro:

mcraveiro:
Thanks for verifying the two flips against their repositories. The round is recorded in the task's * Review table and story Decisions (commit 3389f7d7ee). The fold-in suggestions cannot regenerate, for reasons per file: ClientJobDefinitionModel (scheduler) sits on unrelated pre-existing drift (its org evolved after its last regen), so flipping it would pull stale drift into this foundation PR; the deferral was already recorded and the scheduler per-component drift story owns it. ClientWorkspaceModel is not a trivial flip: its org is a legacy-'* Primary key' entity whose codegen validation fails in both regen runs, so it cannot regenerate until the unblocking task 836BB4CE lands. That is the acceptance-recorded class. The compute/src and api/src duplicates are endorsed as a follow-up: no generator writes to ores.qt/compute or ores.qt/api for these entities (canonical output goes to ores.qt/reporting and ores.qt/dq), so no regen can flip them; their deletion belongs to the migration follow-up, and the ODR hazard you noted is recorded in Review row 1.

mcraveiro:
Fixed in 3389f7d7ee: the stale feature page now records the retirement (task 9FF6FAD3). The flag has no behavioural effect any more, the fetch is unconditional, and the literal survives on entity orgs and profiles as documentation only. Recorded, not folded in, per your fallback: the four compute/src models and api/src ClientDatasetBundleModel are migration orphans. No generator writes to ores.qt/compute or ores.qt/api for these entities; canonical output goes to ores.qt/reporting and ores.qt/dq. No regen can flip them, and their deletion belongs to the migration follow-up. Your drift-from-own-flag note on dataset_bundle confirms the orphan record: the api copy froze mid-migration while the canonical dq copy is the flag-on paginated shape. ClientWorkspaceModel is likewise not a trivial flip: its org is a legacy-'* Primary key' entity that fails codegen validation in both regen runs, unregenerable until 836BB4CE. Review rows 1 and 2 record all of this with the scheduler deferral.

PR #2009, opened 2026-09-04
[ores.codegen] Add RLS policy emission and aggregator wiring validation

mcraveiro:
## Summary

Codegen now emits ENABLE ROW LEVEL SECURITY/CREATE POLICY per domain entity instead of leaving every entity author to hand-write the RLS-layer blocks, and a new validation rule closes the silent failure where a generated create/drop file is never \ir-wired into its component aggregator. ores.marketdata is the adoption proof; the remaining components migrate inside their own drift stories, with the pre-existing unwired-file debt recorded in validation_ignore.txt.

## Changes

- Three RLS variability features (rls_tenant_isolation, rls_party_isolation, rls_system_tenant_visible) as ores.sql.schema features with profile defaults
- Domain-entity SQL create emission appends declared policies as a post-table trailer; drop emission carries the policy drops; junction entities excluded
- ores.marketdata adoption: all five entities declare shapes, hand RLS files and their \ir lines deleted, second regen byte-identical
- WIRE_001 reachability rule in plantuml_er_parse_sql.py: every *_create.sql/_drop.sql must be \ir-reachable from create.sql/drop.sql; deliberate exceptions encoded in the rule
- validation_ignore.txt ledger records the pre-existing unwired debt (3 create + 74 drop files) per file for the component drift stories; zero-warning census pass
- Verification: docs+ci+python tooling; site build clean; codegen_templates tangle + component drift check (refdata, reporting, marketdata) no drift; roundtrip exit 0; compass test suite 117 passed/1 skipped; codegen pytest 145 passed; validate_schemas.sh --strict 0 warnings; compass db recreate green (48s)

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Close systemic codegen gaps](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/close-systemic-codegen-gaps/story.html) | 8B4DE7A9-A10A-4F23-B300-D2B98CA8D7FA |
| Task | [Codegen leaves new entities unwired: RLS policies and the create/drop aggregators](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/close-systemic-codegen-gaps/task_codegen_rls_policy_facet.html) | 467F77DB-3829-4F15-8E8E-C64FAA369459 |


mcraveiro:

mcraveiro:
Fixed in 8ac6f77ae1 for the early-return finding: the blanket return is removed, so RLS_001/RLS_002 now scan every *_create.sql file (the inline-emission pattern) and fire regardless of whether any *_rls_policies_create.sql remains; only RLS_003 depends on the legacy files, and its loop no-ops on an empty list. Regression tests added in test_plantuml_er_parse_sql_rls.py, and a real-corpus validation run stays at zero warnings. Not changed for the observation_lineages policy-name finding: no non-recreate upgrade path exists, and a generated drop file cannot carry a hand-maintained one-off rename without re-introducing the hand-maintained state this program removes; the orphan is inert (permissive policies OR together) and disappears on the next full recreate.

mcraveiro:
Fixed in 8ac6f77ae1 for both findings. Finding 1: validate_rls_isolation now raises at the sql_section read point in core.py when rls_party_isolation is set without rls_tenant_isolation, mirroring the house validator idiom, so the misconfiguration points at the entity instead of surfacing two steps away via RLS_001/RLS_002; tests in test_rls_isolation.py. Finding 2: rls_table_base gets its own truncation budget (63 minus the 28-char isolation suffix) applied only when RLS is opted in and no product/component prefix was stripped, so an unconventional :tablename: cannot silently overflow; zero emission change on today's corpus, test in test_sql_name_base.py. Not changed for finding 3: the ledger count in the PR body is illustrative of the migration and cosmetic.

PR #2012, opened 2026-09-05
[compute,codegen] Bind compute entities to profiles, zero-diff regen

mcraveiro:
## Summary

Bind every codegen-eligible ores.compute entity to a variability profile, regenerate, and confirm the output matches the checked-in tree. Six domain entities bind to uuid-identified-lookup and the app_version_platform junction to tenant-scoped-junction; the hand-crafted messaging protocol, entity handlers and hand-written services give way to the standard codegen'd set; the generator defects the rebind surfaced are fixed at source.

## Changes

- Bind the seven ores.compute entities per the survey: app, app_version, batch, host, result and workunit to uuid-identified-lookup; the app_version_platform junction to tenant-scoped-junction (file-level #+profile). result is the one change-reason exception (list-only entity; grid machinery writes).
- Replace the hand-crafted protocol classes, six entity handlers and six hand-written service files with the standard codegen'd messages, handlers and CRUD services; the custom repository reads (list_by_workunit, list_by_batch) generate from FK :list_by: flags. Two documented exclusions remain: the app/app_version system-tenant union filter (:system_tenant_visible: recorded as the template-feature candidate) and report_submit_handler plus registrar.cpp.
- Fix at source, in the generator, the defects the rebind surfaced: derive table-display cell optionality from the raw cpp_type (dccbe5d574); gate the qt detail-dialog display-name row on the name member existing (86393b44a8); count a display name carried as the natural key as a real member (9a5c60a45b). Restore the six legacy entities' Table display sections the JSON-to-org migration dropped (81ddb04d60); regenerate the detail dialogs and adapt the qt and shell call sites.
- Adopt the history-provider registrar gating and its marketdata/reporting byproducts: registrar pairs are dropped for no-audit entities (8d94e6d8fa, 2500c5cdee, 20e3b980f6).
- Docs: close the task and story (entity-classification-drift-compute); fix the two latent org-export defects the site publish surfaced in the task doc (bare ={{{sys_decl}}}= / ={{{tenant_where}}}= splices parsed as org macros; an unresolvable [[id:...]] placeholder link, now pointed at the codegen-fix-drift skill).
- Verification: code + python tooling + ci + docs class; full build clean at 100% with ores.qt.exe linked; ctest 71/71 green; roundtrip clean; codegen drift checks byte-identical for refdata, reporting, marketdata and compute-cpp (check_component_drift.py); compass pytest 117 passed, 1 skipped; site build clean.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
[...]

PR #2015, opened 2026-09-05
[ores.compute] Restore grid dispatch and result-submit channels

mcraveiro:
## Summary

The post-merge system test of the compute/iam bind PRs found that bind PR #2012 dropped two side effects the grid depends on: the workunit dispatch seam and the trusted wrapper result-submit channel. Restore both in the hand-written files that per-entity regeneration cannot remove, and prove the grid drains live.

## Changes

- Finding #4: restore the grid dispatch seam dropped by the bind. The generated workunit CRUD replaced the hand-written save path, and with it the side effect that published JetStream assignments. Add workunit_dispatcher, an event-bus subscriber in the service app layer, constructed before event_source.start() so no change is missed: it creates the result row and publishes the assignment to each platform package the app version ships.
- Finding #5: restore the wrapper result-submit channel. The bind rerouted wrapper reporting onto the JWT-gated save_result CRUD; wrapper nodes hold no JWT, so every submission was rejected and the wrapper logged 'failed to deserialize response'. Restore the trusted channel (heartbeat-grade transport trust, no verifier): submit_result_request/response on compute.v1.results.submit in the hand-written work_protocol.hpp, the new result_submit_handler with the Validator (canonical_result_id at target redundancy) and Assimilator (batch closed when all workunits canonical), its queue subscription in the aggregate registrar, and the reverted wrapper call site.
- Review round (2026-09-05): dispatch now converges on the redundancy target (1b8118df97) — a workunit short of its target tops up the missing assignments on the next event, so a partial publish failure heals instead of stranding the workunit. Declined with recorded reasons: guarded-update hardening of the canonical assignment (a multi-instance scale-out prerequisite), a unit-test seam, and the round-robin nit; confirmed intentional: the trusted submit channel.

## Testing

**Plan.** Code class per the pr-raise rule table: full build and the complete ctest suite via compass, roundtrip, cmake-sources and codegen drift checks; then the live end-to-end drain on brave_hopper through ores.shell. UI sessions were not available during the system test; every flow was driven from ores.shell (shell-only testing directive).

**What was tested.** All green on the fix branch (linux-clang-debug-make):

- Full build clean; ctest 71/71 green — at raise, and again after the review-round fix 1b8118df97 (71/71, 261.75 s at -j 8).
- Roundtrip check exit 0 (informational; committed inputs untouched by this diff).
- cmake-sources drift clean for ores.compute (12 stale lists elsewhere pre-date this branch: untracked codegen residue in six other components, out of scope).
- Codegen drift: refdata/reporting/marketdata regenerated byte-identical (the six-file tree diff the tool lists is this PR's own hand-written content).
- Live smoke on brave_hopper, shell-only: after service restart, a dispatch-batch of 10 workunits at redundancy 1 drained through 5/5 online wrapper nodes — 10/10 results success, canonical_result_id set, batch closed. SMOKE PASS.

**Limitations.**

- No UI driving: all flows were exercised through ores.shell; the UI-only test gaps are tracked in a separate story per the system-test directive.
- Redundancy above 1 was not exercised live; the Validator threshold and the dispatch top-up branch added this round have no unit coverage yet (recorded follow-up).
- Single-instance topology only. Multi-instance scale-out needs the dispatcher eventing and the submit canonical-assignment guarded update redesigned together (review finding 1, declined for that reason).
- The smoke grid is single-platform (x64-linux).

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Manage the compute grid from the shell](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/shell-compute-grid-management/story.html) | 36E9BEDD-4804-4822-AFD8-8AF961978398 |
| Task | [Fix the compute grid dispatch and wrapper result-submit channels dropped by the ores.compute bind](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/shell-compute-grid-management/task_fix-compute-grid-dispatch.html) | 77101559-5AE3-4390-94AC-65D88D866D63 |


mcraveiro:

mcraveiro:
Review round addressed — verdicts per finding:

**1. Validator/Assimilator race (medium) — Not changed.** The competing-consumer premise requires two service instances on one queue group. ores.compute.service runs single-instance, and multi-instance is not a supported topology today: each instance would run its own postgres_event_source into an in-process event_bus, so both would dispatch every workunit — the double-dispatch breaks before the submit race is reachable. The guarded update (UPDATE ... WHERE canonical_result_id IS NULL) is the right hardening once scale-out is designed, and it belongs in that coordinated change together with dispatch idempotency across instances. Recorded as a scale-out prerequisite.

**2. Partial dispatch failure never retried (medium) — Fixed in 1b8118df97.** dispatch_one now reads the existing result count, skips only when the workunit is at or above target_redundancy, and tops up the missing assignments on each event, so a partial js_publish failure heals on the next workunit event instead of stranding the workunit. The dispatch() doc comment in the header now states the convergence semantics. Verified: full build clean + ctest 71/71 green (linux-clang-debug-make) on the fix.

**3. Unauthenticated submit channel (informational) — Confirmed intentional.** The result_submit_handler class docstring records the rationale: wrapper nodes hold no user JWT, so the JWT-gated generated save_result flow rejects them; the channel carries heartbeat-grade transport trust on the internal NATS. It is deliberately kept out of the regenerated result handler, whose per-entity regeneration would also strip it (that is how the original submit path was lost).

**4. No unit tests (minor) — Not changed in this round.** The live smoke on brave_hopper exercised the happy path end to end (10/10 results success, 5/5 hosts, canonical ids set, batch closed). The redundancy-above-1 and top-up branches are not covered: both handlers construct their repositories and NATS client directly with no injection seam, so unit coverage needs a small refactor first. Recorded as a test-coverage follow-up in the post-merge system-test task (#79).

**Nit (round-robin) — No change.** avps[i % avps.size()] deliberately spreads a workunit's replicas across every platform package the app version ships when redundancy is above 1. The 'platform x64-linux' wording in the PR description described the smoke environment's single platform, not a code pin; the publish subject uses each avp's own platform_code.

PR #2016, opened 2026-09-05
[iam,sql] Grant iam_service select on refdata for the account_party soft FK

mcraveiro:
## Summary

Bind PR #2014 regenerated the iam junction templates; the account_party insert function now runs under the plain invoker idiom, so its soft-FK validation subselect against ores_refdata_parties_tbl executes as the iam service user. The registry granted iam_service select on ores_assets_ and ores_dq_ only, so the first live tenant provisioning on a fresh database failed with 'permission denied for table ores_refdata_parties_tbl'; ctest never caught it because the tests run as privileged roles. Express the grant in the service registry model, where the investigate-hand-added-service-grants story added the cross-domain capability, and regenerate so the grants file is generator-owned again.

## Changes

- service_registry.org: add ores_refdata_ to the iam_service Select prefixes
- iam_service_db_grants_create.sql regenerated from the model; the diff is the single grant line

## Testing

**Plan.** Model+SQL regen class per the pr-raise rule table: lossless-regeneration proof, full build and the complete ctest suite via compass, codegen drift and roundtrip checks; then live provisioning on a fresh database under the real service-user path (shell-only).

**What was tested.** All green on the fix branch (linux-clang-debug-make):

- service_registry regeneration lossless, proven twice (no-op before and after the edit).
- Full build clean; ctest 71/71 green.
- Codegen drift byte-identical for refdata/reporting/marketdata; roundtrip check exit 0 (informational).
- Live on a fresh database recreated at this commit: has_table_privilege confirms the iam service user now holds select on ores_refdata_parties_tbl, and the tenant-provisioning leg that failed before the fix (account_party soft-FK insert under the real service-user path) completed end to end.

**Limitations.**

- No UI driving: all flows were exercised through ores.shell; the UI-only test gaps are tracked in a separate story per the system-test directive.
- The ctest suite runs as privileged roles and cannot exercise the service-user grant path; the live provisioning run on the fresh database is the evidence for the fix.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Entity classification and drift baseline: ores.iam](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-iam/story.html) | C7AB2C55-5EF6-4DCE-8000-A2BEB3184971 |
[...]

PR #2020, opened 2026-09-06
[ores.codegen] Emit value_or(0) for nullable numerics in result mappers

mcraveiro:
## Summary

Main's continuous CI has been red on Linux and macOS since bind PR
#2012 regenerated the compute result mapper. The mapper-impl codegen
template emits `v.<field>.value_or({})` for nullable numeric fields,
and a braced-init-list cannot deduce the template parameter of
`std::optional<T>::value_or`. gcc 13 and Apple clang enforce the
standard here and fail to compile `result_mapper.cpp`; the local
toolchains (g++ 16.2, clang 21.1) and MSVC/clang-cl accept the
deduction, which is why the bind PR built green locally and the Windows
jobs compiled the same file.

The template now emits `value_or(0)` at both nullable-numeric sites.
Zero is semantically identical to `{}` for the two fields in play:
`pgmq_msg_id` (nullable bigint to `std::int64_t`) and `outcome`
(nullable int to `int`, generator default 0), both read into
non-optional domain scalars.

## Changes

- **Template fix** — `ores.cpp.repository.mapper_impl.org` emits
  `value_or(0)`, never `value_or({})`, for nullable numeric fields at
  the identity and plain group sites; the mustache templates are
  retangled.
- **Regenerate compute-cpp** — the only generated diff across all eight
  compute models is `result_mapper.cpp` (lines 45 and 47). Refdata,
  reporting and marketdata regenerate byte-identical.
- **Record** — sprint-25 hotfix story and implement task ride this
  branch; the evidence below is recorded in both docs.
- **Verification**: mixed docs+ci+code class. Site build green; codegen
  drift clean (refdata, reporting, marketdata byte-identical after
  regeneration); domain roundtrip exit 0; full build clean and full
  ctest green on linux-clang-debug-make.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Hotfix: CI builds red across all OSs on value_or({}) in the generated result mapper](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ci-builds-red-hotfix/story.html) | D1896FA3-026B-4056-A4FD-8C18BF9DF4C1 |
| Task | [Implement Hotfix: CI builds red across all OSs on value_or({}) in the generated result mapper](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ci-builds-red-hotfix/task_implement-ci-builds-red-hotfix.html) | 670C711E-2DCB-42DD-9F01-EC45FFF5310D |

## Testing

Plan. Remove the ill-formed deduction from the template, regenerate,
and prove the local union of checks (site build, codegen drift,
roundtrip, full build + ctest) stays green. The compile break itself
cannot reproduce on the local toolchains, which accept both forms, so
the fix is additionally proven by the CI records below: the same two
lines fail deterministically under the CI compilers and no tracked
file carries the old idiom.

Evidence. Linux CI run 34028102867 (linux-gcc-debug-ninja) fails at
`result_mapper.cpp:45:43` with "couldn't deduce template parameter
'_Up'" (`/usr/include/c++/13/optional:1039`) and "no matching function
for call to 'std::optional<long int>::value_or(<brace-enclosed
initializer list>) const'". macOS CI run 34033068694
(macos-clang-debug-ninja) fails the same way at `result_mapper.cpp:45:35`
and `47:27` with "no matching member function for call to 'value_or'".
Both runs built commit 1b8f8146 (merge of #2016); the same failure
traces back to 56a483c7 (merge of #2012), where the idiom entered.
Windows is the control: MSVC compiled the same file in run
33972829991 ([2437/3912] building result_mapper.cpp.obj, no error) and
that run failed only on an app_id test seed, since fixed; run
34009148424 failed on a transient file-lock link error. This PR is
expected to turn Linux and macOS green on merge; final confirmation is
the continuous runs on main after merge.

Limitations. No UI, no new behavior. Local compilers accept the old
idiom, so the local build cannot falsify the original failure; the
authoritative check is the post-merge CI on main, gated on the same
workflows that were red.

### DIFF
--- a/doc/agile/versions/v0/sprint_25/ci-builds-red-hotfix/story.org
+++ b/doc/agile/versions/v0/sprint_25/ci-builds-red-hotfix/story.org
@@ -37,11 +37,11 @@ green.
 
 | Field         | Value                                  |
 |---------------+----------------------------------------|
-| State         | STARTED                             |
+| State         | DONE                             |
 | Parent sprint | [[id:0E7321A7-A66D-4CE6-B3E7-89F2E35CA48B][Sprint 25]] |
-| Now           | Template fix in flight on branch hotfix/ci-builds-red-hotfix. |
+| Now           | Nothing. |
 | Waiting on    | Nothing.                               |
-| Next          | Roll the story DONE when the PR merges. |
+| Next          | Nothing. |
 | Last touched  | 2026-09-06                               |
 
 * Acceptance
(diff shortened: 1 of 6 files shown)


mcraveiro:

PR #2021, opened 2026-09-06
[trading,codegen] Bind trading entities to profiles, zero-diff regen

mcraveiro:
## Summary

Bind every codegen-eligible ores.trading entity to a variability profile, regenerate, and confirm the output matches the checked-in tree byte for byte. Three simple-lookup orgs, nine instrument orgs and lifecycle_event bind to profiles under projects/ores.trading/modeling; trade, trade_identifier and trade_party_role are recorded as genuine no-matches with their blocking features documented after org-form PK migration, and trade's five-way whole estate is verified by measured clobber rather than assumed. The generator defects and regeneration losses the rebind surfaced are fixed at source. trade and trading-cpp do not join the known-drift-free registry: registering them would make the drift check fail by construction, so the registry stays seeded with the six re-verified components.

## Changes

- Bind per the survey recorded in the task notes: party_role_type, trade_id_type and trade_type to simple-lookup, the nine instrument orgs (balance_guaranteed_swap, callable_swap, cap_floor, fra, inflation_swap, knock_out_swap, rpa, swaption, vanilla_swap) to trading-instrument, and lifecycle_event to workspace-scoped-lookup; the lookup orgs the bindings replace are retired.
- Record the genuine no-matches instead of forcing a bind: trade_party_role (no uuid detail-form widget in the Qt drawer vocabulary for non-null soft-FK columns), trade_identifier (uuid-identified-lookup features vs string natural key) and trade (whole estate) are migrated to org-form primary keys and regenerated as no-match entities, with trade's regen verified as a restructure by measured clobber (37 outputs).
- Fix at source the losses and generator gaps the rebind surfaced: restore the Table display sections (operator<< on the domain product_type; the trade_party_role org section), move the rates umbrella removes to the plural service API and add delete_fra_instruments to the restored fra service, point the trade_type repository test at the plural generator namespace, and align regenerated generator version expectations in tests.
- Generator changes under this branch: repository entity headers now emit org-declared entity includes after the template's fixed include set (the trade_type product_type column header is the load-bearing need), with a standard-token filter applied at both loader extraction sites so the whole-estate drift gate stays byte-identical (Timestamp arrives via db_types, boost uuid headers are dead because uuid renders as string, cstdint is a stale transcription); the history-provider registrar facet is dropped for audit-group entities; the drift check is restricted to two modes (--all registry or --component).
- Docs: close the bind task and the entity-classification-drift-trading story with Result, acceptance markers and Decisions (registry decision, no-match rationale, entity-includes filter contract); sprint 25 board row dated; the superseded Apply safe drift to trading-cpp ad hoc task marked ABANDONED pointing here; two memories recorded (regen byproduct hygiene, weak-booking model).

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Entity classification and drift baseline: ores.trading](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-trading/story.html) | 914050BE-700A-4799-903F-5E27B44185D1 |
| Task | [Bind ores.trading entities to profiles; verify zero-diff regen](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-trading/task_bind-trading-entities-to-profiles.html) | 62B120E9-AE25-4786-B90B-1BDE8A4ACC34 |

## Testing

Plan. Regenerate every rebound entity and diff against its pre-rebind checked-in output (byte-identical is the pass condition); then run the full build, the full ctest suite on a recreated test database, the roundtrip and drift checks, the compass test suite and the site build. trade_type regeneration is re-verified after the include-filter change because that commit lands after the main regen pass.

Evidence. Class: code + python tooling + ci + docs. Full build clean at 100% with ores.qt.exe linked; ctest 71/71 green on the recreated test DB; roundtrip clean; cmake-sources drift clean; codegen drift checks byte-identical for the six registered components (check_component_drift.py --all exit 0 over refdata, reporting, marketdata, compute-cpp, iam and iam-cpp); compass pytest 117 passed, 1 skipped, covering the loader filter changes; site build clean. The built C++ tree is byte-identical to the branch head: the two post-build commits change the org loader and docs only, and the drift gate re-verified exit 0 after them.

Limitations. trade and trading-cpp are not in the known-drift-free registry: registering them would fail the drift check by construction. No desktop-UI verification of the rebound entities was run (shell and test-suite level only).


mcraveiro:

PR #2030, opened 2026-09-07
[doc] Specify the skills framework as a regulator over an agent

mcraveiro:
## Summary

Specifies the compass skills framework as a regulator over an LLM agent, and derives from that what the previous documents stipulated. Scope follows from the good regulator theorem, the law of requisite variety and the black box. The seven classes of document are derived from the primitives of a control loop rather than chosen. The upstream pstack collection is classified against that derivation rather than forked. Adds a writing standard for the corpus and two lint checks for rules the framework states but nothing enforced.

## Changes

- Adds The Cybernetic Skills Framework as the specification: scope, the seven regulatory functions, ultrastability, and what the framework leaves to the agent
- Adds Grounding, separating the activity from ground truth and from the set of types implementing loading
- Adds The pstack Collection, classifying all 52 upstream skills by regulatory function with a disposition each; upstream is referenced, never forked
- Renames the session role from function to System, which is Beer's term, and the seven classes from trait to regulatory function
- Adds doc-review-academic, a layered writing and review standard grounded in Gopen and Swan (1990) and Swales (1990)
- Adds compass index --org-ids, giving the org-id scan one home shared by six build scripts that each carried a copy with drifted exclusions
- Adds generator-marker and durable-link checks to compass lint; the second is advisory while its 142 existing violations are worked through
- Enables section numbering across the site

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Specify the skills framework as a regulator](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_specify_framework_as_regulator.html) | 39264768-EE92-4DB1-B048-3C960415B358 |

## Testing

Plan. Documentation and tooling, so verification is by build, lint and rendered output rather than by tests. Every page must publish, compass lint must pass its enforced checks, the new command and both new lint checks must be exercised directly, and the four build scripts rewired to share the org-id scan must be run.

Evidence. compass build --direct site completes with zero failures and every page in the cluster returns 200. compass lint passes filetags and generator markers. compass index --org-ids rebuilds the map in 3.1s, compass site page recovers from a deleted map in 8s, and compass build --direct --no-index confirms the bypass reaches the elisp. The manual, help, plan and release-notes builds all run, and the sprint 24 release notes regenerate byte-identical to the committed file. After the prose reflow, link brackets were checked to balance in every file and list items confirmed rendering as list items in the published HTML.

Limitations. No agent-behaviour evaluation was run, so no claim is made that this changes how agents act; the paper states that under Limitations. The durable-link check is advisory and reports 142 pre-existing links, filed as a backlog capture. The story's other tasks remain unstarted: nothing here renames a skill, builds the dispatcher, or attenuates the catalogue by level.


mcraveiro:

mcraveiro:
Local checks, recorded here because the PR body was written before the diff was formally classified against the pr-raise rule table.

The diff spans three classes. `doc/*` is docs. `projects/ores.compass/src/compass.py` is python tooling. `CMakeLists.txt` and `projects/ores.lisp/*` fall to the code class, the latter through the catch-all.

Run and passing:

- Site build. `compass build --direct site` completes with zero failures; every page in the cluster returns 200.
- Compass test suite. 127 passed, 1 skipped.
- The new command and both new lint checks, exercised directly. `compass index --org-ids` rebuilds the map in 3.1s; `compass site page` recovers from a deleted map in 8s; `compass build --direct --no-index` confirms the bypass reaches the elisp.
- The four build scripts rewired to share the org-id scan: manual, help, plan and release-notes. Sprint 24's release notes regenerate byte-identical to the committed file.

Not run, and why:

- Full build and ctest. The code-class trigger here is an `add_custom_target(org_ids ...)` in `CMakeLists.txt` and the elisp catch-all. Neither touches a C++ target, a header, or anything the compiler sees. I judged the full build disproportionate rather than skipped it under time pressure, and I am flagging it so a reviewer can disagree.
- Roundtrip, cmake-sources drift and codegen drift. No C++ sources were added or removed and no `projects/*/modeling/*` path is touched, so none of the triggers apply.

One thing worth a reviewer's attention beyond the diff. The durable-link check added to `compass lint` is advisory rather than enforced, because it reports 142 pre-existing violations across 72 files. That is deliberate: a check failing on its first run teaches people to ignore it. It is filed as a backlog capture with the sweep described, and flipping the default is a one-line change once that lands.

mcraveiro:
Thanks — one finding fixed, plus the regex observation, both in 40fdcff89c.

**skill-add step 9.** Confirmed and fixed. You are right that the heading was migrated and the prose below it was not, and that `doc/functions/` is now a dangling path. It reads System, doc/systems/, and "the System or Systems the skill belongs to" throughout. This is the ninth defect traceable to that rename and the first found by someone other than me, which is a useful datum about how far a mechanical rename's damage spreads past what grep finds.

**`_ID_LINK_RE` and bare links.** Not fine to leave, so fixed. A bare `[[id:UUID]]` is equally valid org and equally capable of pointing into a sprint, so the gap was in what the check *claims* to cover rather than in its strictness. The regex now matches both forms with the description group optional. It was hiding two violations: the count moves 142 to 144. I also stopped the backlog capture quoting a hard number, since it ages every time main merges; it now cites the live count from lint.

**Two changed-pages implementations.** Not changed, and I agree with your reading. `compass.py` always resolves explicit paths, so it never reaches the elisp `--all` branch and the two cannot disagree today. Consolidating on that basis would be speculative work against a divergence that does not exist yet. Recorded in the task's review table so whoever wires the elisp path up meets it.

**Out-of-order generator `END`.** Not changed. Every marker pair in the corpus is flat, never nested, and the check's value is that it is simple enough to trust. Precise reporting for a case that does not occur would add branching without adding coverage.

On the checks you could not run: fair, and the gap you could not close is the one I would want a second opinion on. Round checks here were compass tests (127 passed, 1 skipped), skills bundle rebuilt, 399 site pages published with zero failures after the rebase, and lint clean on both enforced checks.

PR #2043, opened 2026-09-09
[compass] Index doc/knowledge/external and lint dangling id links

mcraveiro:
## Summary

The doc walker pruned every directory named external at any depth. The entry exists for the vendored ./external tree at the repository root, but it also hid all twelve documents under doc/knowledge/external, so that cluster resolved nowhere in the graph and never appeared in compass search. This anchors the large exclusions at the repository root, then turns the invariant into an enforced check so the failure cannot recur unnoticed.

## Changes

- doc_index: split the exclusions. EXCLUDED_DIRS keeps the names that are noise at any depth; EXCLUDED_ROOT_DIRS holds the large repository-root trees and is pruned only at the root.
- compass lint: report every [[id:...]] whose target :ID: is in no document, and fail on it.
- compass lint walks through doc_index.find_org_files, so the lint and the graph share one definition of the corpus; a run drops from over two minutes to one second. Dotfiles are skipped, since the per-worktree .journal.org links to documents on other branches.
- New Doc Lint workflow runs the command on every push and pull request that touches an .org file.
- Regression test for the walker and for the dangling-link check.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Fix the invisible knowledge cluster and audit graph integrity](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_fix_knowledge_graph_visibility.html) | FE910755-16F8-4854-BA3D-D780BE5AFB64 |

## Testing

Plan. Reproduce the missing cluster, fix the walker, prove the new check fires on a dangling link and passes on the clean corpus, and run the checks for the python tooling, ci and docs classes the diff spans.

Evidence. python tooling, ci and docs. Compass suite: 129 passed, 1 skipped. Site build: Build succeeded. Roundtrip check: exit 0. misspell-fixer: exit 0. compass lint over the whole corpus: clean, and it now runs in 1s. Repro before the fix: compass show 0412444A-0A4C-4611-887C-09353A3CB253 answered 'No doc or anchor matches'; after it returns QuantLib with ten inbound links, compass search quantlib ranks it at 84%, and all twelve external documents load. A probe document carrying a bogus link made lint exit 1 with 'dangling id links: 1 link(s) resolve to no :ID: in the corpus'. The walker regression test fails against the old exclusion list with assert set() == {'quantlib.org'}.

Limitations. No C++ was touched, so no build or ctest run. Codegen drift was not run: the diff touches no model, template or generated source. The Doc Lint workflow itself is first exercised by this PR.

### DIFF
--- a/.github/workflows/doc-lint.yml
+++ b/.github/workflows/doc-lint.yml
@@ -1,0 +1,42 @@
+# Copyright (C) 2026 Marco Craveiro <marco.craveiro@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free Software
+# Foundation; either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 51 Franklin
+# Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+---
+name: Doc Lint
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches: [main]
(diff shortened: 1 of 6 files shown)


mcraveiro:

PR #2045, opened 2026-09-10
[ores.trading] Bond relational model, wave 1.6: the loading exercise

mcraveiro:
## Summary

Wave 1.6 of the bond relational model task. The population loader runs at the three standard sizes on a database recreated from HEAD. The eight family tables land at 8864 to 10155 rows/sec, about four times the deliverable's 2200 rows/sec baseline, and all nine family entities regenerate with zero diff.

## Changes

- Record the wave 1.6 loading exercise in the task Notes and Result: six runs at --rows 1000, 10000 and 20000, with the per-table and aggregate rows/sec and the comparison against the deliverable baseline and its sizing guidance.
- Record the attribution evidence for the aggregate split: the isolated against in-run penalty, the notify-trigger readings, and the exclusion constraint that keeps the insert trigger in place.
- Record the family drift gate: all nine entities regenerate with zero diff, and the dependency on the port story and the parent analysis's drift PR.
- Add four rows to the task Review table from the pre-raise local pass, three of them corrections to the record's own claims.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Implement the bond relational model: codegen, SQL and the loader exercise](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_implement-bond-relational-codegen.html) | D7943D7E-2889-4C6B-857A-C382DBF755BA |

## Testing

Plan. Docs class: the diff is one task org file, so the site build is the matching check. The measured work the record reports ran live against the database during the wave: the population loader at the three standard sizes, and the per-entity regeneration for the family.

Evidence. Site build clean: ./compass.sh build --direct site exits 0 with Build succeeded. Loader: 49 tables at 41,800 rows for 1000, 418,000 rows for 10000 and 836,000 rows for 20000, zero findings and clean cleanup at every size. Family drift gate: compass codegen entity generate --diff reports no differences for each of the nine entity slugs. misspell-fixer is clean on the changed file and on doc/.

Limitations. The aggregate timings are not reproducible to better than about 2x, so the record carries all six samples rather than one number. The in-run penalty on the legacy tables is recorded but not diagnosed; that work sits with the parent analysis. The shell system test for the family is wave 1.7. The repo-wide misspell-fixer run exits 1 while reporting no hits and no replacements in verbose mode; that is pre-existing and outside this diff.

### DIFF
--- a/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_implement-bond-relational-codegen.org
+++ b/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_implement-bond-relational-codegen.org
@@ -28,9 +28,9 @@ Bond is the pilot for the relational target model of the trading analysis (dfe15
 |--------------+----------------------------------------|
 | State        | STARTED                              |
 | Parent story | [[id:688959F0-A1AB-4021-A57E-1103AD92773A][Redesign ores.trading on data-oriented principles]] |
-| Now          | Wave 1.5 done and verified. The reshaped instrument table and the nine family tables are live on a recreated database (schema version 0.0.25, commit cc013f1589 stamped), the baseline population runs at the family premise with zero findings, and the verification cycle is green on the new shape: ctest 71 of 71, roundtrip exit 0, codegen drift and cmake lists clean. The migration script converts an old-shape database with its row counts and joins validated on a scratch copy. The wave records are in the Notes. PR #2044 carries the wave; its review round fixed three defects (the shell TRS leg type, the repo and TRS reverse legs, stale fact staging in the qt form) and two minor items, and declined the both-facts guard in favour of wave 1.7's code-vs-fact cross-check. |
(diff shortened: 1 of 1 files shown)


mcraveiro:

mcraveiro:
Fixed in 6f5e65b7a4. Both occurrences now read "eight family tables". The 8864 to 10155 range spans the eight child and fact tables only: they load at 2000 rows each, while bond_instruments loads at the full 20000 and is reported apart at 7948 to 8423. Wave 1.5's own sentence already used "eight family tables" for this set. The nine-entity drift claim and the two nine-table presence claims are left alone, because those are correct as written. The PR body carried the same count and is corrected too.

The comma grouping is not changed. Tables group thousands and the surrounding prose does not. That split predates this wave, as in wave 1.5's "7123 rows/sec", so it is recorded as the convention rather than a defect.

Rows 19 and 20 of the task's Review table record both decisions.

PR #2046, opened 2026-09-10
[ores.trading] Bond relational model: the shell system test, wave 1.7

mcraveiro:
## Summary

Wave 1.7 closes the bond relational model task. It records the live shell round trips through ores.shell against the environment, and closes the task DONE with the Result as the closing statement over the nine-entity family.

## Changes

- Records wave 1.7, the shell system test, in the task Notes: the environment rebuild, eleven probes covering all ten trade type codes and their fact branches, the paginated get, the eleven histories and deletes, and the DB-layer, changed-event and CHECK-gate verification.
- Closes the task DONE: the Result becomes the closing statement over the nine entities, the migration, the loader exercise and the shell round trips, and the parent story Tasks row and Status are synced.
- Corrects two claims the local review caught, recorded as Review rows 21 and 22: the family is nine tables rather than ten, and the trade-type negative does leave its issue row behind.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Implement the bond relational model: codegen, SQL and the loader exercise](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_implement-bond-relational-codegen.html) | D7943D7E-2889-4C6B-857A-C382DBF755BA |

## Testing

Plan. The change is the live system test it records: a scripted ores.shell session per the generic-batch protocol on task C36A4ECA, run as tenant_admin@acme against a database recreated from HEAD, with a NATS subscriber on the prefixed ores.dev.brave_hopper.ores.trading subject and psql reads at the DB layer.

Evidence. docs class; local site build clean (./compass.sh build --direct site, exit 0, with both rendered pages carrying the corrections). Every number in the record was re-verified against the live schema and the captured artifacts: nine bond tables, 14 instrument rows all soft-closed, 15 issues open, one option row (Call, strike 1000) and two TRS rows, 28 event deliveries over 14 instrument ids with two each, and the four constraint names read live.

Limitations. The shell transcript and the event capture are local artifacts and are not committed. The three findings are recorded as boundaries, not fixed: the delete that leaves the fact row open belongs to the parent story service layer, and the codegen sentinel and the non-atomic add are unchanged. Windows is out of scope.


mcraveiro:

mcraveiro:
Fixed in 776339b558. The note now reads "one of the four that the partial-state finding below counts", and Review row 23 records the round. The independent re-derivation of the counts, the four constraint names and the sentinel convention against the source is useful confirmation. The bearings block is a permission setting for the review job rather than anything in this PR; noted for the workflow owners.

PR #2048, opened 2026-09-10
[compass] Instrument skill selection

mcraveiro:
## Summary

We know skill mis-selection is real because it has been corrected by hand for months, but we have no instrument, so we cannot quantify or report it. This builds one that records selections forward rather than scoring past tasks, so the measurement accrues while the migration proceeds and keeps working after it lands.

## Changes

- PreToolUse hook on the Skill tool records every selection: the skill chosen, the session, the environment, how many skills were visible, and the attenuation level from ORES_SKILL_LEVEL. Literate source in doc/llm/claude_code_settings.org, tangled with its tests.
- The hook returns 0 on every failure path, including a malformed payload and an unwritable log, because a hook command that cannot execute blocks its tool outright.
- compass skills correct records the judgement half as a separate append-only event, against the fixed cause vocabulary: description, catalogue-size, level and none.
- compass skills report prints the correction rate, the causes ranked, the split between attenuated and unattenuated sessions, and the most corrected and most chosen skills over a window.
- The log is JSONL in the git common directory, so every worktree in the fleet appends to one file and the baseline never restarts on env provision.
- New recipe: How do I report on skill selection?

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Instrument skill selection so the baseline accrues forward](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_add_selection_instrumentation.html) | 37BC3547-4FE9-4D13-B3E7-5C071C5155B9 |

## Testing

Plan. Drive the real hook end to end with a synthetic PreToolUse payload, record a correction, read the report, and run the checks for the python tooling and docs classes the diff spans.

Evidence. python tooling and docs. Compass suite: 144 passed, 1 skipped, including 16 new tests. Site build: Build succeeded. cmake-sources drift: All component_files.cmake up to date. compass lint: clean, with the durable-link advisory back at its pre-existing 160. End to end: piping a Skill payload into record_skill_selection_hook.py exited 0 and appended a selection reading offered 188, level null; compass skills correct --cause description then made the report show Corrections 1, the cause ranked, and the level split as unattenuated. The probe rows were removed afterwards so the shared baseline starts clean.

Limitations. No C++ was touched, so no build or ctest run. Codegen drift was not run: the diff touches no model, template or generated source. The hook reaches a session only after the settings are tangled, so it records nothing for sessions already running. The first populated report is the gate before the sequencing function lands, not this task.

### DIFF
--- a/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/story.org
+++ b/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/story.org
@@ -73,7 +73,7 @@ Tasks are listed in intended execution order.
 | Task | State | Start | End | Description |
 |------+-------+-------+-----+-------------|
 | [[id:FE910755-16F8-4854-BA3D-D780BE5AFB64][Fix the invisible knowledge cluster and audit graph integrity]] | DONE | 2026-09-09 | 2026-09-09 | The whole doc/knowledge/external cluster does not resolve in the graph; diagnose, fix, and add a continuous S3-star check so it cannot recur. Unblocks domain grounding. |
-| [[id:37BC3547-4FE9-4D13-B3E7-5C071C5155B9][Instrument skill selection so the baseline accrues forward]] | BACKLOG | | | Record selection events as they happen rather than scoring past tasks, so skill-choice quality becomes measurable and reportable from now on. |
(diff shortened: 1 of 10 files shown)


mcraveiro:

PR #2049, opened 2026-09-10
[doc] Land the skill architecture into policy

mcraveiro:
## Summary

The skill architecture states the model but binds nobody. This moves it into the two documents an author actually reads, and into the skill that enforces them. The model itself does not change.

## Changes

- skill_naming_conventions: the compass- prefix and the five name shapes; principle- and method- as second segments; a table mapping each name shape to the regulatory function it offers; the act- and export- verbs; the level register with its advisory and binding strengths; and the placement procedure, reproduced where an author naming something will meet it.
- The reference-only-skills rule now exempts the citation function. A principle is by definition cited and never run, so requiring executable steps of it would forbid the entire function.
- glossary: the terminology register lands here rather than in the skills cluster, because one word per concept governs every document. It names the retired synonyms and both traps.
- glossary: Agent now carries the general definition rather than the delegate's, and Session and Delegate are separate entries.
- skill-add: the naming step states the prefix and the segments and routes an unsure author to the placement procedure, and a new step makes the level a field to fill rather than a rule stated elsewhere.
- Every register extension carries a decision-log entry, as the policy already demanded of itself.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Land the skill architecture into policy](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_design_target_skill_model.html) | F7B5BED7-AB0A-4ECD-A8E4-B31E7A77F2EF |

## Testing

Plan. Check the policy against its own sources, rebuild the deployed skills bundle so the enforcing skill actually ships the new steps, and run the docs-class checks.

Evidence. docs. Site build: Build succeeded. Skills bundle: Build complete, and the deployed skill-add carries the new level step. compass lint: clean, with the durable-link advisory unchanged at its pre-existing 160. misspell-fixer on the three changed documents: nothing to replace. The three new glossary ids were each confirmed to appear exactly once in the corpus.

Limitations. No code was touched, so no build, ctest or drift run. Three source conflicts were resolved in favour of the later document and recorded rather than escalated: seven regulatory functions rather than the six the task names, System rather than function for a session role, and the corrected Agent definition. The level keyword is mandatory from now on while 52 of 83 skills do not yet carry it; that sweep is its own task in this story.


mcraveiro:

mcraveiro:
Fixed in e13e2591a4 — you were right that a register contradicted by its nearest cross-reference is not yet landed, so this went in here rather than as a follow-up. execution_model.org's six role rows now say System, and the Agent row says delegate; s1_agent.org's description called the role a subagent too, so that went with it.

Not changed: the name-shape table listing a representative rather than exhaustive set of types per function. That is deliberate. The table orients an author naming something, and the exhaustive mapping is one link away in Regulatory Functions; copying it in would create a second copy to keep in step.

Noted on the environment limit — since neither lint nor a build ran on your side, the local run for this round was: compass lint clean with the durable-link advisory unchanged at its pre-existing 160, and the site build succeeded.

PR #2050, opened 2026-09-10
[doc] Collapse the uniform verbs into polymorphic actions

mcraveiro:
## Summary

The catalogue was an N-by-M table written out by hand: seven add skills were one verb over seven artefact types, six show skills one verb over six. A verb that behaves uniformly becomes one skill plus a type table, and the test is whether the procedure differs rather than whether the artefact does. Doing this before the imports land avoids porting cells we were about to delete. The catalogue goes from 82 skills to 69.

## Changes

- doc-show absorbs the eight showing skills, each of which held one or two steps: run a compass read command, read its output. Its target table is generated from compass's own subparser registry.
- doc-add absorbs the four document-scaffolding skills, which shared one shape: scaffold, fill, wire the parent index, commit. Its type table is generated from the doc_*.org.mustache templates, so a new document type joins the table by existing, carrying the sections its template declares and the parent-dir default compass applies.
- codegen-add-surface-entity absorbs the CLI, HTTP, shell and Wt entity skills, which differed only in file locations, build target and phase contents. Qt stays separate: it generates before it writes, which is a different procedure rather than a different surface.
- Skills that survive state why their procedure differs: a memory is mostly the judgement of whether to write one, a recipe carries a testable script, a skill has registers to parse and a bundle to rebuild.
- generate_skill_type_tables.py derives both tables and carries a --check mode, wired into the Doc Lint workflow alongside the catalogue check so a stale table fails CI.
- The catalogue no longer carries a ledger of removed skills, and skill-delete no longer writes one. It carries the current set; what a skill used to be called is in git. Two stale deployed documents from earlier renames are purged.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Collapse the uniform verbs into polymorphic actions](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_collapse_polymorphic_actions.html) | 629B48C7-BC7E-44A8-8979-3E05D6DA5BFE |

## Testing

Plan. Retire each skill through the skill-delete procedure, prove the graph is whole afterwards, prove the generated tables regenerate to what is committed, and run the docs and python-tooling checks.

Evidence. docs and python tooling. compass lint: clean, and it caught 22 dangling links left by the retirements in the archived plans and the facet models, all since re-pointed; the durable-link advisory is unchanged at its pre-existing 158. Site build: Build succeeded. Both generators pass --check. Compass suite: 145 passed, 1 skipped. misspell-fixer on the new files: nothing to replace. Deployed bundle rebuilt and the 15 retired directories purged; 69 skill directories deployed, and this session's own catalogue picked up doc-add, doc-show and codegen-add-surface-entity.

Limitations. No C++ was touched, so no build or ctest run. Codegen drift was not run: the diff touches no model or template output. Closed sprint records still name the collapsed skills in their own accounts of what happened at the time; rewriting those would falsify the record, and every live document was re-pointed. The surface table is written by hand rather than generated, because its four rows draw on four documents with no common schema and C++ surfaces do not churn the way document types do.


mcraveiro:

PR #2052, opened 2026-09-10
[doc] Tag every skill with its cybernetic level

mcraveiro:
## Summary

The level turns the cybernetic systems into a variety attenuator. A regulator swamped by variety it cannot use stops regulating, and an agent shown the whole catalogue at once is exactly that. Every skill now declares which System it serves, and a declared System selects the slice in scope by default.

## Changes

- apply_skill_levels.py holds the level table. A level is a judgement about which System a skill serves and a wrong one hides a capability, so it is a reviewed table rather than a rule inferred from names. A skill missing from it is an error rather than a default, so the next skill cannot land untagged.
- The audit channel's slice is derived rather than tagged: show- and find- promise to change nothing, so the auditor sees exactly those five skills. The promise written into the verb register earlier in this story is now load-bearing.
- --prune-deployed removes the out-of-scope skills from the deployed tree, so an agent spawned afterwards cannot load one. That is the spawn-time enforcement a read-only System needs, rather than an instruction in a prompt a delegate under pressure can reinterpret.
- Normal operation deploys the full set, which is what keeps the other five levels advisory: an agent may reach outside its slice because the skill is still there.
- The level-to-scope inventory in Systems and levels is generated from the same table that tags the skills, so the document cannot claim a scope the catalogue does not have. --check runs in the Doc Lint workflow.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Tag every skill with its cybernetic level](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_add_skill_cybernetic_levels.html) | 3969C185-48DE-47F0-A00D-FD28CD58C1B0 |

## Testing

Plan. Apply the table, read every slice back, prove the audit channel is actually enforced rather than merely described, and run the docs and python-tooling checks.

Evidence. docs and python tooling. All 69 skills carry a table level and --check passes. Slices: s1 sees 58, s2 15, s3 18, s3star 5, s4 13, s5 13, with 12 cross skills in every advisory slice. Enforcement proved on a copy of the deployed tree: pruning to s3star removed 64 directories and left exactly the five read-only ones, then the full tree was restored and rebuilt to 69. compass lint: clean, durable-link advisory unchanged at its pre-existing 158. Site build: Build succeeded. Compass suite: 145 passed, 1 skipped. misspell-fixer on the new files: nothing to replace.

Limitations. No C++ was touched, so no build or ctest run. The distribution is lopsided at 46 s1 skills, which reflects that most of what we do is execution rather than a tagging error. --prune-deployed is available at any level while only being appropriate for a read-only spawn: pruning makes attenuation binding, and binding an advisory level would contradict the rule that an agent may reach outside its slice.

### DIFF
--- a/.github/workflows/doc-lint.yml
+++ b/.github/workflows/doc-lint.yml
@@ -60,3 +60,9 @@ jobs:
         # The polymorphic skills dispatch on these. A stale table sends an
         # agent to a command or a document type that has moved.
         run: python3 build/scripts/generate_skill_type_tables.py --check
+
+      - name: Check every skill carries its level
+        # A skill with no level is visible at every level, which defeats
+        # the attenuation. A skill missing from the table is an error
+        # rather than a default, so a new skill needs a considered level.
+        run: python3 build/scripts/apply_skill_levels.py --check
--- a/build/scripts/apply_skill_levels.py
+++ b/build/scripts/apply_skill_levels.py
@@ -1,0 +1,332 @@
+#!/usr/bin/env python3
+"""Set the cybernetic level on every skill, and report the attenuated slices.
+
+The level turns the cybernetic systems into a variety attenuator: a session
+declares its System, and the level selects the slice of the catalogue in
+scope by default. A regulator swamped by variety it cannot use stops
+regulating, and an agent shown every skill at once is exactly that.
+
+The level is a judgement about which System the skill serves, so it is a
+reviewed table rather than a rule over names. LEVELS below is that table:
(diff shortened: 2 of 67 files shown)


mcraveiro:
@claude please review this pull request. The level assignments in build/scripts/apply_skill_levels.py are the substance here — a wrong level hides a capability, so please challenge any assignment you find questionable rather than only checking the mechanism.

PR #2053, opened 2026-09-10
[doc] Separate skill judgement from recipe commands

mcraveiro:
## Summary

The recipe holds the command, tested and runnable by a human; the skill holds the judgement — when to reach for it, what must hold first, what to do when the guard rails refuse. A skill that restates a recipe's command creates a second place for it to drift. The survey changed the shape of the work: of the 17 skills carrying a command with no recipe behind it, 16 already had a recipe nobody had linked, and only one operation had none.

## Changes

- New recipe: How do I rebase my branch on main, covering the rebase, the force-with-lease push, and both conflict paths.
- link_skill_recipes.py points each of the 17 skills at the recipe that owns its commands. The mapping is the judgement, applying it is mechanical, and --check holds it in CI so the delegation cannot rot.
- Every Reference section now names the knowledge the skill assumes. Seventeen were empty.
- Eight skills were a command and nothing else, which under the architecture is a recipe wearing a skill costume. Rather than keep the duplicate they gained the judgement they should have carried: what to check first, what the command does that is not obvious, and how its failure misleads.
- The sequencing rule is documented in Methods and the mode, where a method author meets it: a missing effector becomes a recipe, never an inline command.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Separate skill judgement from recipe commands](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_separate_skill_recipe_boundaries.html) | F4EF0C2D-B32E-4705-AE0F-AAC3E027FF59 |

## Testing

Plan. Survey which skills carry commands and which recipes already own them, apply the links, fill what the skills were missing, and run the docs checks.

Evidence. docs. link_skill_recipes.py --check passes across all 17 skills; every one of the 69 skills now names both its recipes and the knowledge it assumes, verified by counting id-links in both sections. compass lint: clean, durable-link advisory unchanged at its pre-existing 158. Site build: Build succeeded. Skills bundle rebuilt, and this session's harness reloaded the rewritten skills. misspell-fixer on the new files: nothing to replace.

Limitations. The first acceptance criterion is only partly met and is filed rather than counted as done: 105 shell blocks remain in skills where the command sits inside a judged sequence, compass-devops-setup-environment being the shape of it. That is a real remaining violation, so it is a DISCOVERED task in this story with its own acceptance. No code was touched, so no build, ctest or drift run.

### DIFF
--- a/.github/workflows/doc-lint.yml
+++ b/.github/workflows/doc-lint.yml
@@ -66,3 +66,8 @@ jobs:
         # the attenuation. A skill missing from the table is an error
         # rather than a default, so a new skill needs a considered level.
         run: python3 build/scripts/apply_skill_levels.py --check
+
+      - name: Check every skill links the recipes that own its commands
+        # The recipe holds the command and the skill holds the judgement,
+        # so drift has one place to be detected rather than two.
+        run: python3 build/scripts/link_skill_recipes.py --check
--- a/build/scripts/link_skill_recipes.py
+++ b/build/scripts/link_skill_recipes.py
@@ -55,8 +55,10 @@ OWNERS = {
          "the environment lifecycle verbs"),
     ],
     "compass-devops-run-client": [
-        ("compass/how-do-i-use-named-env-files",
-         "launching the client per environment, with colour"),
+        ("ops/how_do_i_ready_up_an_environment",
+         "the readiness the client assumes, and the launch with a scenario"),
+        ("ops/how_do_i_test_a_feature_after_implementation",
+         "handing a readied client to a tester"),
     ],
     "compass-devops-deploy-manual": [
         ("cmake/how_do_i_build_the_system", "the deploy_manual target"),
--- a/build/scripts/link_skill_recipes.py
+++ b/build/scripts/link_skill_recipes.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+"""Point every action skill at the recipes that own its commands.
+
+Under the architecture the recipe holds the command, tested and runnable by
+a human, and the skill holds the judgement: when to reach for it, what must
+hold first, what to do when the guard rails refuse. A skill that restates a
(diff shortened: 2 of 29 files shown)


mcraveiro:
@claude please review this pull request. Two things worth challenging in particular: whether the judgement written into the eight command-only skills is genuinely judgement rather than padding, and whether the 105 shell blocks I deferred to a follow-up task really do sit inside judged sequences rather than being duplicates I should have cleared here.

mcraveiro:
Fixed in a3ed76ad86 — both regressions were real and mine. You were right that they are exactly the judgement this PR's thesis says must survive delegation.

compass-devops-run-client: restored the absolute-path rule for --open-scenario with the reason (the client's working directory is not the repo root), the readiness-and-verify-it-yourself requirement, and the login-as-first-Steps-entry rule with why prose does not do (the Scenario Runner tracks pass and fail per step). Also re-pointed its recipe link: you are right that how-do-i-use-named-env-files covers neither flag, so it now links how_do_i_ready_up_an_environment and how_do_i_test_a_feature_after_implementation, which do. That also heals the cross-reference from those two recipes back to this skill.

compass-devops-deploy-skills: restored the reload reminder, with the reason stated rather than just the instruction.

On challenge #2, you found the flaw in my framing and I have amended the follow-up task. compass-agile-refine-backlog is the counter-example: it already links the recipe that owns its commands and still restates them, git mv and git commit included, adding no ordering. The sweep now starts by sorting the 105 into mechanical duplicates and genuine sequences, because the mechanical half needs no hand-written prose at all. That makes the follow-up cheaper than I had framed it.

Noted that you could not run the interpreter: link_skill_recipes.py --check passes locally, the site build succeeded, and compass lint is clean with the durable-link advisory unchanged at 158.

PR #2055, opened 2026-09-10
[doc] Add structure notes to the knowledge Zettelkasten

mcraveiro:
## Summary

The knowledge index says which pages exist. It cannot say which of forty are load-bearing, nor in what order to read them, and that judgement is authored rather than computed. Six domain clusters now have a structure note that supplies it.

## Changes

- Two clusters already had their entry point: market data resolution was a full structure note under the name hub note, and entity lifecycle already carried the layer ordering per row. Both are now named and tagged as such, and the entity lifecycle states why its order holds rather than only asserting it.
- Interest Rate Curves — pillars before interpolation, because interpolation is only a question once the curve is known to be defined at finitely many points.
- Volatility — realised before implied, because the surface is a price rather than a measurement.
- Identity and Access — the application layer before the database backstop, with the reverse order given for the one case that inverts it, debugging a query that returns too few rows.
- MASD — logical space before physical, because a file layout reads as arbitrary without the model it projects.
- link_structure_notes.py adds a back-link from every page a hub orders, so a reader who lands by search learns the order exists. The hubs declare their own membership and the outward-pointing sections are headings so the scan can stop there. --check runs in Doc Lint as the seventh check.
- Zettelkasten described the theory and said we had none; it now carries the convention and the three rules that keep a note honest. The grounding phase reads a cluster's structure note first where one exists.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Add structure notes to the knowledge Zettelkasten](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_add_knowledge_structure_notes.html) | 6766A23C-0357-4873-B5BF-B5A71BF0B1D5 |

## Testing

Plan. Write the notes, wire them both ways, prove the set is listable as a whole, and run the docs checks over a change touching forty documents.

Evidence. docs. compass list --tag structure_note returns all six. 33 back-links added across the pages the hubs order, and link_structure_notes.py --check passes. compass lint: clean, durable-link advisory unchanged at its pre-existing 158. Site build: Build succeeded, which is the check that matters for a change spanning this many documents. misspell-fixer on the four new notes and the script: nothing to replace.

Limitations. No code was touched, so no build, ctest or drift run. The reading orders are judgements and the reviewer should challenge them rather than check them: I argued each from the dependencies between the pages, not from measured reader behaviour. Interpolation is ordered by two hubs and carries a back-link to each, which is expected in a Zettelkasten rather than a fault. Clusters beyond the six named in the task still have only the index.

### DIFF
--- a/.github/workflows/doc-lint.yml
+++ b/.github/workflows/doc-lint.yml
@@ -71,3 +71,8 @@ jobs:
         # The recipe holds the command and the skill holds the judgement,
         # so drift has one place to be detected rather than two.
         run: python3 build/scripts/link_skill_recipes.py --check
+
+      - name: Check structure notes are reachable from what they order
+        # A hub only helps readers who already found it, unless every page
+        # it orders points back at it.
+        run: python3 build/scripts/link_structure_notes.py --check
--- a/build/scripts/link_structure_notes.py
+++ b/build/scripts/link_structure_notes.py
@@ -1,0 +1,152 @@
+#!/usr/bin/env python3
+"""Make every structure note reachable from the pages it orders.
+
+A structure note is a hub: it names a cluster and puts its pages in an
+argued order. Reachability has to work both ways. A reader who finds the
+hub gets the order; a reader who lands on one page by search needs to be
+told the order exists, or the note only helps people who already knew
+about it.
+
+The hubs are the source of truth: whatever a hub links to, it orders. This
+script reads each hub's outbound links and adds a back-link to the hub in
+each ordered page's See also section, skipping pages that already carry
+one.
+
+Usage: python3 build/scripts/link_structure_notes.py [--check]
+  --check  exit non-zero if an ordered page lacks its back-link.
+"""
+import argparse
+import pathlib
+import re
+import sys
+
+ROOT = pathlib.Path(__file__).resolve().parents[2]
+SEARCH = [ROOT / "doc", ROOT / "projects" / "modeling"]
+TAG = "structure_note"
+
+LINK = re.compile(r"\[\[id:([0-9A-Fa-f-]{36})\]\[([^\]]*)\]\]")
+
+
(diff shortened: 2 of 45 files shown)


mcraveiro:
@claude please review this pull request. The substance is the four reading orders, and they are judgements rather than facts — please challenge any ordering you think is wrong, and say so if a note reads as a table of contents with prose attached rather than an actual argument. I would rather hear that now than ship six hubs nobody trusts.

mcraveiro:
Fixed in ad01adde37 — all three accepted, and the first was a real bug in my own mechanism.

Asides: you are right that the script asserted an order the prose denies. Rather than add those two headings to the stop list, aside sections now mark themselves with a '# aside' line and their pages get distinct wording — 'sits alongside its sequence rather than in it'. I went that way because matching on heading wording asks a script to interpret prose, and it breaks the first time someone phrases a section differently; the marker is explicit and the convention is now documented in zettelkasten.org alongside the other three rules. curve_point_provenance and both papers carry the aside wording; the 30 genuinely ordered pages are unchanged.

GARCH: accepted, and this is the correction I am most glad of. 'Not constant, therefore shaped' does conflate a P-measure time-series property with a Q-measure cross-sectional one. The justification now claims only the leverage asymmetry producing the same qualitative skew, and explicitly states the limit so the next reader does not widen it back. GARCH stays second for the reason you gave.

entity_lifecycle renames: fair — they are stale labels from before the compass- rename, caught while tagging that page as a structure note. Noted in the PR description rather than split out, since reverting them would leave known-wrong names in a file this PR already edits.

Thank you for checking the other three orders against the source pages rather than taking them on trust. A challenge that finds nothing is what makes the two that found something worth acting on.

PR #2056, opened 2026-09-10
[doc] Cite the principles under the citation function

mcraveiro:
## Summary

A principle is a rule an agent cites at a judgement call and never runs. The thirteen adopted from pstack are referenced upstream rather than copied, because forking them would sever the feedback path that lets an upstream rule improve. What this adds is the grouping, the two bindings where our detail differs, and one rule moved out of prose into a check.

## Changes

- Principles is the index: the thirteen grouped by when they are cited, so a phase reaches for the two that bear on its decision rather than reading the set. Upstream attribution is recorded once for the collection, which is the right granularity when nothing is forked.
- compass-principle-type-system-discipline binds the rule to C++. A translation table, plus the two patterns the language makes harder: exhaustive matching, which stops being enforced the moment a default label appears, and branding, which costs a struct and its operators rather than a type alias.
- compass-principle-never-block-on-the-human reconciles the rule with our confirmation conventions on one criterion: proceed where a written record makes the choice reversible, confirm where the cost is paid before anyone reads the trail.
- The audit said eleven of twenty-one principles were adopted; its own table says thirteen. Corrected in the collection page and the two agile documents repeating the figure.
- The encode-as-a-check test moved one rule out of prose. The stash stack is shared by every worktree, so a bare push contends for its top and a bare pop can take another session's work. It is now a PreToolUse hook, literate with its tests, that denies the two bare forms and names the safe ones in the denial.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Cite the principles under the citation function](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_port_principles_layer.html) | A7BCF024-9222-40C3-9151-AFE9C9B4BDBE |

## Testing

Plan. Write the index and bindings, apply the encode-as-a-check test to every candidate, prove the resulting hook both denies the real thing and permits prose about it, and run the docs and python-tooling checks.

Evidence. docs and python tooling. Compass suite: 159 passed, 1 skipped, including 14 new hook tests. The hook was proved live rather than only in tests: running a bare stash in this session was denied with the correction, and the sentence containing the same words passed. compass lint: clean, durable-link advisory unchanged at 158; it also caught two mistyped ids while I wrote the index. All five generator checks pass, and the catalogue generator now knows the principle segment. Site build: Build succeeded. misspell-fixer on the new documents: nothing to replace.

Limitations. The acceptance asks for each principle to be ported as a skill. I did not port them, because forking thirteen upstream documents contradicts the reference-not-fork decision this story already took; the reinterpretation is recorded in the task Result rather than left implicit, and it is the one thing in this PR most worth disagreeing with. No C++ was touched, so no build or ctest run. The groupings are judgements about when a rule is cited and should be challenged as such.

### DIFF
--- a/build/scripts/apply_skill_levels.py
+++ b/build/scripts/apply_skill_levels.py
@@ -115,6 +115,11 @@ LEVELS = {
     # Identity: what belongs in the product at all.
     "compass-agile-brainstorm-idea": "s5",
 
+    # Bindings on an upstream principle. The level is the level of the
+    # judgement the principle is cited at, not of the principle itself.
+    "compass-principle-type-system-discipline": "s1",
+    "compass-principle-never-block-on-the-human": "cross",
+
     # Every level acts on these: the documents and skills themselves, and
     # the orientation any System starts from.
     "compass-doc-add": "cross",
--- a/build/scripts/generate_skills_catalogue.py
+++ b/build/scripts/generate_skills_catalogue.py
@@ -43,6 +43,11 @@ DOMAINS = {
     "devops": ("DevOps (=compass-devops-=)",
                "Environment, database, services, shell, client, site."),
     "skill": ("Skills (=compass-skill-=)", "The skills themselves."),
+    # Not a domain: the reserved second segment for a rule that is cited
+    # rather than run. See the naming conventions.
+    "principle": ("Principles (=compass-principle-=)",
+                  "Local bindings on an upstream principle, supplying the "
+                  "project detail the rule cannot know."),
 }
 
 # Planned-but-not-yet-created skills, shown as prose under each section.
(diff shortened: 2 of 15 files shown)


mcraveiro:
@claude please review this pull request. The thing most worth disagreeing with is that I did not do what the acceptance literally says: it asks for each principle to be ported as a skill, and I cited them instead, on the grounds that forking thirteen upstream documents contradicts the reference-not-fork decision this story already took. If you think the acceptance should have won, say so.

Two other things to challenge: whether the C++ binding's claims about exhaustive matching and branding are actually true of this codebase, and whether the groupings in the index put each principle at the moment it is really cited.

mcraveiro:
Fixed in ae95f524d0 — all three accepted, and the second one is the finding of the session.

The -Wswitch claim was wrong and I verified it before acting rather than taking it on trust: compiling a non-exhaustive switch over an enum class with -Werror alone succeeds. You are right that -Wall is absent from every CMakeLists.txt and .cmake in the tree, so I documented a mechanism that does not exist. The section now states that the warning is off, why the advice still stands without it, and that a default: label on a closed enum is what will hide the missing case the day the flags go on.

Enabling them is filed as its own DISCOVERED task rather than done here, for the reason your 174-file count implies: the flag and the sweep have to arrive together, and each default: needs judging — legitimate over an open set, concealment over a closed enum. Its acceptance requires a case that fails before the change and passes after, so the mechanism gets proved rather than assumed a second time.

Acceptance bullet: accepted and amended in place to say cited rather than ported. You are right that a Result note does not help a reader skimming the table, and leaving a criterion visibly unmet is worse than editing it with the reason attached.

sequence-verifiable-units: accepted, moved to before-writing. Your argument is the better one — it was sitting beside prove-it-works doing a different job, and a split settled as the PR opens has already cost what it was going to cost.

On the first point, thank you for checking the story's decision log instead of taking my framing at face value. A task that reinterprets its own acceptance in the same PR that claims to satisfy it deserves exactly that scepticism, and the verdict is worth more for having been tested that way.

PR #2057, opened 2026-09-10
[doc] Port the sequencing function and bind it to our effectors

mcraveiro:
## Summary

A method is control flow with no effectors; our actions and recipes are effectors with no control flow. This joins them. Six methods, one per shape of work: bug fix, feature, refactoring, investigation, prototype, and figure-it-out for work no method fits.

## Changes

- They share a skeleton because the skeleton is the contract. Ground the concepts the work names; state before anything is built what would make it finished in checkable terms; cite principles at the judgement calls; name the action or recipe for anything touching an artefact.
- Each ends by saying when it is the wrong method. A misjudged shape is the failure a method exists to prevent, and no phase inside one recovers from having picked it.
- The methods carry no shell blocks at all, which is the recipe boundary from earlier in this story applied to control flow.
- Nothing was copied. The upstream playbooks live inside poteto-mode, which is adopted whole, so forking them would contradict the reference-not-fork decision. What is written is the local sequencing that binds our effectors, which upstream cannot supply because it does not know our actions exist.
- Steps assuming a foreign host are dropped rather than translated: Graphite stacks, model routing, and an issue-tracker abstraction, all ground our agile layer already owns.
- Levels for the six, a compass-method- section in the generated catalogue, and the set listed in Methods and the mode with the work each sequences.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Port the sequencing function and bind it to recipes and actions](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_port_method_layer.html) | 08F7F925-49B2-40E0-A298-E300353CAE2E |

## Testing

Plan. Write the six, wire them into the level table, the catalogue and the methods page, and run every corpus check plus the emacs export that the new files have to survive.

Evidence. docs. compass lint: clean, durable-link advisory unchanged at 158. All five generator checks pass: levels for 77 skills, the namespace, the recipe links, the structure-note back-links, and both generated tables. Skills bundle rebuilt and this session's harness reloaded all six under compass-method-, which is the emacs org export applied to every new file.

Limitations. The site build could not be run. It was killed three times by the machine's low-memory guard — twice as a background task, once after a foreground run was moved to the background at its timeout — while the box reported 19G available. Rather than retry a fourth time I covered the org-export risk another way: the skills build runs the same emacs export over doc/llm/skills and completed, exporting all six methods and the amended methods page. The two agile documents here are the only files without that check, and both are conventional edits. Dangling id-links, the site build's other unique catch, are covered by compass lint. No code was touched, so no build or ctest run.

### DIFF
--- a/build/scripts/apply_skill_levels.py
+++ b/build/scripts/apply_skill_levels.py
@@ -115,6 +115,16 @@ LEVELS = {
     # Identity: what belongs in the product at all.
     "compass-agile-brainstorm-idea": "s5",
 
+    # Methods. The level is the level of the work the method sequences:
+    # execution for the four that carry a change, cross for one that only
+    # reads, s2 for the one that plans work spanning many units.
+    "compass-method-bug-fix": "s1",
+    "compass-method-feature": "s1",
+    "compass-method-refactoring": "s1",
+    "compass-method-prototype": "s1",
+    "compass-method-investigation": "cross",
+    "compass-method-figure-it-out": "s2",
+
     # Bindings on an upstream principle. The level is the level of the
     # judgement the principle is cited at, not of the principle itself.
     "compass-principle-type-system-discipline": "s1",
--- a/build/scripts/generate_skills_catalogue.py
+++ b/build/scripts/generate_skills_catalogue.py
@@ -48,6 +48,11 @@ DOMAINS = {
     "principle": ("Principles (=compass-principle-=)",
                   "Local bindings on an upstream principle, supplying the "
                   "project detail the rule cannot know."),
+    # Likewise: a sequence of phases, which spans artefact types rather
+    # than naming one.
+    "method": ("Methods (=compass-method-=)",
+               "How a kind of work is approached, from framing to a "
+               "verified end."),
 }
 
 # Planned-but-not-yet-created skills, shown as prose under each section.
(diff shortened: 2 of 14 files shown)


mcraveiro:
@claude please review this pull request. Three things I want challenged rather than checked.

Is six the right set? I took bug fix, feature, refactoring, investigation, prototype and figure-it-out from the design's adopt list, but a method nobody reaches for is dead weight, and I have no usage evidence for prototype in particular.

Is the shared skeleton load-bearing, or a template? All six ground, state a predicate, cite principles, name actions, and say when they are wrong. That is either the contract that makes them methods or a shape I imposed on six kinds of work that differ more than it admits. Say which.

Do the phases actually bind to things that exist? Every action and recipe id should resolve to a skill that does what the step claims. compass lint proves the links resolve; it cannot prove the step named the right one.

Note the site build could not run here — three low-memory kills — so the org export evidence comes from the skills build instead. Details in the PR body.

mcraveiro:
Fixed in 4704d49278 — both findings accepted, and both are the kind no check I have could reach.

The citation slip is the more interesting one. Prototype and figure-it-out cited the bare principle-never-block-on-the-human at exactly the two places where the reconciliation is the entire content: deciding whether to ask the operator, and keeping the trail for an unattended run. Citing the un-reconciled upstream rule there would have told an agent to proceed in precisely the cases our conventions carve out. Both now cite the binding. The general lesson is recorded on the story: compass lint verifies id-links, so a rule cited as =principle-x= is just text and nothing notices when it names something that does not exist.

The marker: removed from skill_methods.org, since the hand-written table this task added is the inventory. You are right that the other four are the same defect, and I filed it rather than sweeping it here — a marker is an instruction not to edit a region, so one naming a script that never writes it is worse than no marker. Its acceptance requires the check that would have caught this, demonstrated failing before and passing after, because the reason four sat stale is that lint checks BEGIN/END matching and not whether the named generator targets the file.

On the three challenges: thank you for answering them with evidence rather than agreement. The closed-graph argument for the set and the phase-count variation for the skeleton are both things I could not honestly assert about my own work, and the prototype verdict — watch, retire into the never-block binding if it stays unused — is now recorded as the retirement criterion so it does not get re-litigated from scratch.

You are also right that "the skills build covers the org export" is narrower than "the site build passed". The limitation now says what is actually covered and what is not: the export and the link resolution, not the rendered page.

PR #2059, opened 2026-09-10
[doc] Build compass-helm, the sticky mode

mcraveiro:
## Summary

A method an agent must remember to reach for is the one it forgets on the task that needed it. The mode is entered once and governs every turn until stood down: it declares the System, matches the work to a method, copies that method's phases in verbatim, and routes to actions, recipes and deliberation as they fire.

## Changes

- The blocking criterion first: the matching predicate is specified before the mode ships. Five ordered questions, first yes wins, ordered by how sharply they discriminate rather than how common they are. Question 2 outranks question 5 deliberately, because work that adds a capability due to a misbehaving one is a defect first.
- Two methods matching the same work is settled by that ordering. Two matching different parts is not a tie at all, and figure-it-out sequences them. Nothing matching sends the reader to the limits first, because most unmatched requests are too small for a method.
- The mode is eight steps: decline if the work does not earn it, declare the System, read the principles index, match, copy the phases in verbatim, run them routing outward, delegate as this same mode, close.
- The verbatim copy is its own step because it is the one most easily lost. The failure is not refusing a phase; it is reading a method and writing a bespoke plan that drops the inconvenient ones. Copying first makes the drop visible.
- Declining is step 1. A mode that fires on everything trains people to route around it, and one that cannot say this is not worth my cost gets disabled rather than used.
- compass-helm is a bare noun, admitted to the register with a decision-log entry; compass-mode was rejected as accurate and forgettable.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Build compass-helm, the sticky mode](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_add_helm_mode_skill.html) | 60C7573C-763F-466D-A174-61C3A78616AF |

## Testing

Plan. Specify the blocking predicate, build the mode, wire it into the levels, the catalogue and the architecture's third entry point, then exercise it end to end on real work rather than asserting that it works.

Evidence. docs and python tooling. Exercised end to end: the mode declared s1, matched compass-method-bug-fix on question 2 of its own rule, copied the seven phases in, and ran them on the stale-marker defect — reproduced four markers naming a script that never writes them, traced the cause to compass lint checking only BEGIN/END pairing, fixed the cause with _lint_marker_generators, and proved it with the check failing on exactly those four and passing after. That run also closes the marker task, which is why it lands here. Compass suite: 159 passed, 1 skipped. compass lint clean, all five generator checks pass, bundle rebuilt and the harness loaded compass-helm.

Limitations. The site build still cannot run here; four low-memory kills are captured separately, so the rendered pages are unverified while the org export and link resolution are covered. Two tasks land in one PR, deliberately: the acceptance requires the mode be exercised on real work before the task closes, and inventing a throwaway exercise would have proved less than doing filed work with it. The matching rule is specified but has one run behind it; whether the ordering holds is a question for the next several uses.

### DIFF
--- a/build/scripts/apply_skill_levels.py
+++ b/build/scripts/apply_skill_levels.py
@@ -115,6 +115,10 @@ LEVELS = {
     # Identity: what belongs in the product at all.
     "compass-agile-brainstorm-idea": "s5",
 
+    # The mode. It declares the System rather than belonging to one, so
+    # every level can enter it.
+    "compass-helm": "cross",
+
     # Methods. The level is the level of the work the method sequences:
     # execution for the four that carry a change, cross for one that only
     # reads, s2 for the one that plans work spanning many units.
--- a/build/scripts/generate_skills_catalogue.py
+++ b/build/scripts/generate_skills_catalogue.py
@@ -53,6 +53,11 @@ DOMAINS = {
     "method": ("Methods (=compass-method-=)",
                "How a kind of work is approached, from framing to a "
                "verified end."),
+    # A bare noun rather than a domain: the mode is a posture, not an
+    # operation on a target.
+    "helm": ("The mode (=compass-helm=)",
+             "The sticky posture a session enters, which matches work to "
+             "a method and routes to everything else."),
 }
 
 # Planned-but-not-yet-created skills, shown as prose under each section.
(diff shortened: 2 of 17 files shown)


mcraveiro:
@claude please review. Three challenges, and one procedural call to disagree with if you think it wrong.

The matching rule is the blocking criterion and has exactly one run behind it. Is the ordering right? I put 'observably wrong' above 'behaviour is new' so that a capability added because the current one misbehaves is treated as a defect first. Attack that if the ordering breaks on a case you can construct.

Does the mode actually bind? Every step should route to something that exists and does what the step claims — the attenuation, the principles index, the methods, the close actions.

Is declining real or decorative? Step 1 says the mode can refuse. If nothing in the document would ever cause a refusal in practice, it is theatre.

Procedural: two tasks land in this PR. The acceptance requires the mode be exercised on real work before the task closes, so I ran it on a filed defect rather than inventing a throwaway. If you think that should have been two PRs with the exercise recorded by reference, say so.

mcraveiro:
Fixed in 0a4b943895. The 200-to-207 case is the best finding of the session, and it is one I could not have produced myself: I wrote the rule, so the cases I generate to test it are drawn from the same intuition that wrote it. You had to come from outside for that.

You are exactly right about the mechanism. Without an anchor, "wrongly" floats, and any change of mind about what is correct can be dressed as a symptom — which then borrows reproduce/root-cause/prove for work that has no cause to find. Question 2 now requires the behaviour to have been wrong against a contract already in force, and I put your counter-example directly beside it so a future reader sees why the clause is there rather than trimming it as pedantry. The ordering itself stands, as you say: a crash under load still reproduces and still gets fixed by adding what was missing.

On declining: your framing is better than my question. I asked whether it was theatre; the honest answer is that it is structurally real — two independent routes reach the exit — and empirically untested. That is the same risk as the matching rule, so both are now one filed task rather than two vague intentions. Its acceptance wants six real uses with the matching question recorded, at least one decline, at least one two-method arbitration, and an explicit challenge to any question that never fires across the sample, since an unused question is either missing its cases or is not discriminating.

One thing I can close that you flagged as unverified: you took the harness loading compass-helm on the PR's word, correctly noting you had no way to reproduce the build. This session can confirm it — the skill appeared in its own catalogue after the bundle rebuild, and again after this round's rebuild.

And thank you for naming the cost of the two-in-one-PR call rather than only agreeing with it. That the lint fix can no longer be reverted alone is a real consequence, and it is now recorded on the task instead of living in this thread.

PR #2061, opened 2026-09-11
[build] Enable the compiler warnings the code assumes

mcraveiro:
## Summary

-Werror without -Wall only escalates the few diagnostics on by default, so a switch over an enum class that misses an enumerator has compiled clean here for years: -Wswitch arrives with -Wall. This enables -Wall and -Wextra for GCC and Clang, /W4 beside the existing /WX for MSVC, and clears everything they surface.

## Changes

- A syntax-only sweep of all 3591 translation units found 2153 warnings. 2040 were -Wmissing-field-initializers and its designated twin, both firing on aggregate initialisation that leaves a trailing field to its default, which is idiomatic here. Both are switched off with the counts and reason recorded at the flags; leaving them on would have buried the 113 that named something real.
- 35 -Wsign-compare: Eigen's size() returns a signed Eigen::Index and the quant processes held it in std::size_t. None of those four files touches a std::vector, so the indices now carry Eigen's own type rather than casting at each comparison. Loops counting ticks keep std::size_t, and the one real boundary between an API tick count and an Eigen size gets an explicit cast.
- 9 unused variables in generated table renderers, fixed in cpp_domain_type_table.cpp.mustache rather than the outputs: an entity with no display columns generates a loop whose body is empty. Applied to 74 files by regeneration.
- 4 -Wreorder-ctor, 1 -Woverloaded-virtual where an accessor with zero call sites hid QPaintDevice::metric, and 62 unused parameters, captures, constants and functions.
- domain.cpp is 104641 lines emitted by xsdcpp from ORE's schemas, so its three warnings are silenced on that one file. The fix belongs upstream in the generator.
- The type-discipline binding is corrected: it described -Wswitch as a mechanism when the flag was off, and now describes the one that exists.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Enable the compiler warnings the code assumes](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_enable_compiler_warnings.html) | A849F82C-0AE6-4217-90FC-F17443B3C168 |

## Testing

Plan. Measure before deciding, fix at the cause rather than the symptom, and re-sweep every translation unit after each round rather than trusting the edits.

Evidence. Three full sweeps of all 3591 translation units: 2153 warnings, then 13, then 3. The final three are domain.cpp, proved clean under its suppression and confirmed wired by reconfiguring CMake and reading -Wall, -Wextra, -Wno-missing-field-initializers, -Wno-unused-const-variable and -Werror back out of the compile database for that file. The gate is demonstrated rather than asserted: the same non-exhaustive switch compiles clean under main's -Werror alone and is an error under these flags.

Limitations. Verified by syntax-only compilation of every translation unit, not by a full build and link: this machine has killed every heavy build tonight, which is captured separately. The operator is running a full build independently. Three findings are filed rather than fixed here, each because it is a judgement rather than a warning: three Wt account operations that confirm and then do nothing, codegen drift of 116 files and 24 uncommitted generated files in three components, and the 174 default: labels which suppress -Wswitch wherever they sit, so the new gate protects new code and not those sites. The task's own acceptance asked for that default: sweep; I have argued it out of scope and said so in the Result rather than quietly dropping it.


mcraveiro:
@claude please review. This one touches the build flags for every target, so the things worth challenging are judgements rather than mechanics.

Suppressing two of -Wextra's warnings is a standing decision about what the build will never tell us again. 2040 of 2153 were -Wmissing-field-initializers and its designated twin, firing on aggregate init that leaves a trailing field defaulted. I judged that idiomatic here and not a defect. If you think the suppression hides real risk, say so.

The Eigen::Index change touches numerical code. Verify the retype is behaviour-preserving and that I did not convert a loop that counts something other than Eigen rows — I got that wrong once and had to put three tick loops back.

The template fix regenerated 74 files. Check the one-line change is genuinely all that landed in them.

Verification is syntax-only across all 3591 units, not a full build; the operator is running a full build separately. Three findings are filed rather than fixed, including that the task's own acceptance asked for a default: sweep which I argue is not this task's work.

mcraveiro:
Fixed in 5eea1bda49 — both accepted.

The suppression point is the one worth having. I justified it by the noise ratio, which explains why the warning is off but not what turns off with it. You named the actual cost: adding a field to an aggregate stops erroring at the call sites that do not set it, and they keep compiling with the field value-initialised. That is the sentence someone will read when a missing-field bug slips through, so it is now at the flags in those terms rather than in terms that only describe the noise. The trade stands; the record of it no longer flatters it.

Orphaned comment removed.

On the Eigen verification: thank you for checking it file by file rather than taking the description. That is the one part of this change I could not verify on my own work — I wrote the retype, so my reading of it carries my own assumptions, including the one that already went wrong when a blanket replace caught three tick loops. Independent confirmation that no tick loop was swept up and that the single boundary cast sits in the right place is worth more than my re-reading it a third time.

Same for checking the 74 regenerated files by numstat. Mechanical evidence that every one is exactly one line changed is stronger than the spot-check I offered, and it is the kind of check I should have run myself before claiming it.

Still outstanding and not something either of us can settle here: this is verified by syntax-only compilation of all 3591 units, never a full build and link. The operator is running that separately.

mcraveiro:
Two further commits since the review, and a correction to the evidence I gave in it.

The operator's full build broke on `trade_generator.cpp`: a declared-and-unused `counter`. That is a defect in my own sweep, not just in the code. The checker decided a translation unit was clean by grepping stderr for a `[-W...]` tag, so a unit that failed to compile outright matched nothing and was counted clean. The "2153, then 13, then 3 across 3591 units" figures in this PR measured less than they claimed. The checker now treats a non-zero exit as a failure, and the task's verification section says this rather than the original numbers.

The corrected sweep covers 3949 units, the count having grown with `ores.synthetic`, and flags 73. Of those, 55 fail on a precompiled header the concurrent full build invalidated and 11 on Qt `ui_*.h` and moc outputs that only exist after a real build step — both properties of syntax-only compilation against a build tree rather than diagnostics. Three are `domain.cpp`, reported only because the sweep appends `-Wall` after the compile database's own flags and re-enables what the per-file `-Wno-unused-const-variable` turns off; the suppression is confirmed present in the database entry.

Six were real: an anonymous-namespace `parse_flag` nothing calls, in six `ores.shell` synthetic command files, hand-written rather than generated. Removed, along with the empty anonymous namespace it left in two of them.

The counter fix went to the org source and was re-tangled, so the 28 changed generated files follow from the template rather than being edited in place; drift passes. The unused counter turned out to be worth keeping as a signal: `needs_counter` switches on for any column marked `requires_unique_suffix`, `trade.external_id` carries that mark, but the generator emits an empty string with no suffix. That column is under a unique index, so a second synthetic trade in one process would collide. Marking the declaration `[[maybe_unused]]` unblocks the build and also silences the only thing pointing at this, so it is filed as its own task rather than left to the compiler to keep reporting.

Your point 1 stands unchanged. What is different is that the verification behind the rest of it is now stated at its real strength.

PR #2064, opened 2026-09-11
[agile] Abandon the unique-suffix task on a false premise

mcraveiro:
## Summary

The task was filed on a premise that does not hold. requires_unique_suffix is applied correctly wherever it is set, and the unused counter that prompted the filing is codegen drift in trade rather than a missing suffix. The task is abandoned with its reasoning, and the evidence moves to the drift task that owns it.

## Changes

- Abandon task_fix_unique_suffix_not_applied, recording why the premise was false and how it was checked
- Extend the codegen drift task with trade as a fourth affected component, including why regeneration is not a one-line diff there
- Note that the [[maybe_unused]] added by the warnings task marks drift and should come out when trade is regenerated
- Sync the story row to ABANDONED

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Apply the unique suffix that requires_unique_suffix asks for](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_fix_unique_suffix_not_applied.html) | 4CB973CA-86B7-4476-85F2-0F11626BFB27 |

## Testing

Plan. Docs class: compass lint and the misspell check, since the change touches only org documents under doc/agile.

Evidence. docs; compass lint clean (filetags and generator markers valid); misspell-fixer exit 0. The factual claims were verified directly: grep for ^:unique: across every model returns one column, tenant.hostname, whose generated line carries the counter suffix; regenerating trade removes the counter declaration entirely.

Limitations. No build or test run, as no code changes. Regenerating trade to prove the drift claim was done locally and reverted; clearing that drift is the other task's work.


mcraveiro:
@claude please review. This PR abandons a task rather than implementing it, so the thing worth challenging is whether the abandonment is correct, not whether the code is.

The claim is that `requires_unique_suffix` has no defect. My evidence: `grep -rln '^:unique:' projects/*/modeling/*.org` returns exactly one file, and the one column it declares (`tenant.hostname`) receives `+ "-" + std::to_string(idx)` in its generated output. If that grep is the wrong instrument — if a column can acquire `unique` some other way, through a field group, an index declaration, or an inherited property — then the survey is too narrow and the task should not be abandoned. That is the check I would most like a second pair of eyes on.

Second claim: the unused `counter` in `trade_generator.cpp` is drift rather than a missing suffix. Trade has a uuid primary key, no natural keys, no unique column, so all three disjuncts of `needs_counter` are false. Regenerating removes the declaration and also rewrites the function from field groups to flat fields and from namespace `generator` to `generators`, which would not compile against the current domain type.

Context worth having: I filed this task myself, during the previous task, while the operator's build was broken. I wrote it from the shape of the symptom and never opened the model to confirm `external_id` carried the property the premise depended on. So treat the filing as suspect input rather than as a specification to check the code against.

mcraveiro:
Both accepted; fixed in f1e6e4a705.

The index-declared uniqueness gap is the finding worth having, and I verified it before recording it: the Indexes table in ores.refdata.party.org does carry codename_uniq on codename, and that column's generator emits a literal empty string with the value assigned by the trigger off its sequence. So it is not a counterexample, and the abandonment stands.

What I have changed is how the Notes state the sufficiency. They now say the narrower grep is sufficient today by accident rather than by construction, and name the case that would slip past both the per-column property and this survey: an index-unique plain-string column added later with an ordinary generator. That is the sentence someone will need if this doubt comes back, and it only exists because you went looking for the escape hatch rather than confirming the instrument I offered.

Also cited test_generator_unique_suffix.py in the Notes. Corroborating no defect with an existing test is stronger than the reading I gave, and I should have found it myself before writing prose about the mechanism.

Status rows padded.

PR #2065, opened 2026-09-11
[ci] Wire both unguarded codegen drift checks into CI

mcraveiro:
## Summary

Three kinds of codegen drift exist and only template drift was enforced. Artefact drift had a working checker no workflow invoked; model drift had no checker at all. This wires both, corrects the nightly PR body that claimed a check which did not run, and regenerates the 102 files the gate found on its first run.

## Changes

- Add codegen-drift.yml running check_component_drift.py --all over the drift-free registry, with clang-format installed so generated C++ does not read as drifted
- Add check_model_drift.py, comparing the features a model states against the values its bound profiles fix, and run it as a second job
- Regenerate 102 generated files left behind by the two template changes merged in the previous two tasks
- Correct the nightly-includes PR body to name the check that now runs
- Document the three drift kinds, what guards each and whether it is enforced
- File report_definition, whose uuid-keyed workspace-scoped shape no profile describes, as a named exception with its own task

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Wire the codegen drift checks into CI and close the false claim](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_wire_codegen_drift_checks.html) | 2DF7DE1E-77CA-419E-90E6-E70DF5223314 |

## Testing

Plan. CI class plus codegen tooling: the two new checks must pass in CI, and each must be shown to fail on real drift rather than only to pass.

Evidence. ci + python tooling; check_component_drift.py --all failed on 102 genuinely drifted files and passes clean after regeneration; check_model_drift.py passes with one named exception and was proved to fail by injecting a contradicting feature into a bound model and to pass again once restored; compass lint clean; yamllint clean on both workflows; misspell-fixer exit 0.

Limitations. No full C++ build: the regenerated files are exactly one line each and the only diff bodies across all 102 are the two template marker lines, verified by numstat. The workflows themselves are first exercised by this PR.


mcraveiro:
@claude please review. This PR adds two CI gates, so the things worth challenging are whether they actually catch what they claim and whether the exceptions are honest.

The strongest evidence here is accidental. Running `check_component_drift.py --all` on clean main, before writing any workflow, failed on 102 files. Both causes were template changes I merged in the previous two tasks: the table-iterator marker and the counter marker. Each regenerated the components it was reaching for and left the rest. Both went through review and CI green. That is the case for the gate, and it is also a fair question about my own work in this story.

Two things I would like attacked specifically.

First, clang-format. Regeneration runs the formatter over its C++ output, so if CI's clang-format differs from the one that formatted the committed tree, the check fails on formatting rather than drift. I install the distro package on ubuntu-24.04 on the grounds that `nightly-format.yml` uses the same one, so both agree on what formatted means. My local clang-format is 21 and ubuntu-24.04 ships 18, and my clean run was under 21. If that reasoning is wrong, this job fails on its first real model change and nobody will trust it again.

Second, the exceptions. `check_model_drift.py` carries one named exception, `report_definition`, which is uuid-keyed and workspace-scoped where `uuid-identified-lookup` fixes `has_workspace_id=false` and `workspace-scoped-lookup` fixes `has_uuid_primary_key=false`. I judged that re-binding would move the false promise rather than remove it, and filed it. If that is rationalising a shortcut, say so. An exception list is exactly where a gate goes to die.

On the checker itself: it compares only features the model states, on the grounds that an unstated feature is supplied by the profile and cannot contradict it. If that lets a real class of drift through, that is the design error to find. I proved the check fails by injecting a contradiction into a bound model and passes once restored, rather than only observing it pass.

mcraveiro:
All five accepted; fixed in 36dd98bbd1.

The drawer finding is the one that mattered and you were right that it invalidated the result rather than dented it. Three of the five features the profiles fix were never compared, so 'no drift, one exception' was the parser not looking. The checker now reads every property drawer keyed by its heading path, and reports a model that states the same feature twice with different values instead of letting ordering decide.

Fixing it produced two real findings within seconds, neither of which the first version could see. compute.result states has_change_reason_cache=false and documents it as overriding the profile default; the profile offers no default, its assignments fix values. So the model and the profile disagree about what binding means, which is a bigger question than a binding correction. reporting.report_instance binds fk-scoped-child and declares no parent at all. Both are named exceptions with their reasons and are filed together as task_settle_profile_binding_semantics, because correcting either needs the semantics settled first.

A fourth issue fell out of the fix rather than the review. fk-scoped-child fixes parent_entity_singular to a prose value saying the feature must be set rather than which value it takes, and comparing that literally reported two correct models as drifted. Assignments beginning with 'required' are now presence-checked, which is what the profile says it means.

On the apt asymmetry: accepted and fixed on the other side. nightly-format.yml now refreshes the index too, with the reason at the step. The invariant only holds if both sides hold it, and it was inconsistent of me to state it while relying on one side not to move.

I reproved the check the same way as before, but on a Qt-drawer feature this time, which is the class the original parser was structurally unable to catch. Worth noting that my original proof passed while the blind spot was wide open, which is a fair warning about what a single injected case demonstrates.

PR #2066, opened 2026-09-11
[analytics] Clear the codegen drift and enforce it

mcraveiro:
## Summary

Clears the codegen drift in analytics-cpp and adds it to the enforced registry. refdata needed nothing. scheduler-cpp turned out not to be a drift clear at all: its model warns against regenerating any facet of it, and the warning is correct, so it is filed as a model resync instead of being regenerated into wrong types.

## Changes

- Regenerate analytics-cpp: 23 files changed and 12 committed that had never been, the history-provider registrars and eventing integration tests
- Fix the model's combo_sort_field, which named a text column where sort_key_of takes an int, so the generated dialog did not compile and had been hand-patched to sort on a constant
- Add the eventing dependency the generated tests need to the analytics test target
- Add analytics-cpp to KNOWN_DRIFT_FREE, bringing the enforced set to eight components
- File the scheduler job_definition model resync rather than regenerating it

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Clear the codegen drift in analytics, refdata and scheduler](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_fix_codegen_drift_three_components.html) | A6DB4F4C-D4F5-4913-AFD3-CC791A921517 |

## Testing

Plan. Codegen and C++ class: the component must regenerate byte-identical, and every new or changed C++ file must compile, since 12 of them have never been built.

Evidence. code + python tooling; check_component_drift.py --component analytics-cpp passes clean, and --all passes with analytics-cpp in the registry; all 4 new registrars, all 4 new eventing tests and the regenerated Qt dialog verified to compile syntax-only under the project's own flags; the pre-fix generated lambda verified NOT to compile, which is the evidence the model was the defect; compass lint clean; check_model_drift.py clean; misspell-fixer exit 0.

Limitations. No full build or ctest: the new eventing tests are compiled but not linked or run, and they are integration tests needing a live database. The eventing dependency added to the test target is verified by supplying the include paths by hand rather than by reconfiguring CMake.


mcraveiro:
@claude please review. Two judgements are worth attacking, and one piece of scope I did not deliver.

First, the sort field. The tree had a hand-edited `return 0;` and the generator emits `return t.code;`. I concluded neither was right and changed the model to name `display_order`. My evidence that the generated form is broken is that I compiled it and it fails: `sort_key_of` is `std::function<int(const Entity&)>` and `instrument_code::code` is a `std::string`. The judgement I would like challenged is the replacement. `display_order` is what 24 other dialogs sort by and the column exists on the entity, but it is a behaviour change: the combo sorted by nothing before and now sorts by display order. If that is a UX change smuggled into a drift clear, say so.

Second, scheduler-cpp, which I did not clear. Its model says in capitals not to regenerate any facet of it, because the column shapes do not match the hand-written struct: regenerating emits `std::string` where the code has a `cron_expression` value type and `int` where it has a `bool`. I judged that accepting that regeneration would regress the code to clear a diff, so I filed a model resync instead. The task's acceptance asked for all three components, so this is scope I deliberately did not deliver, and it is a fair thing to push back on. The alternative I rejected was regenerating and then hand-fixing the types back, which reintroduces exactly the hand-edit problem this task exists to remove.

Third, worth checking independently: I committed 12 files that had never been built. I compiled all of them syntax-only, and four needed an eventing dependency the analytics test target did not link, which I added. I verified that by supplying the include paths by hand rather than reconfiguring CMake, so if the CMake change is wrong in a way my shortcut hides, that is where it will be.

mcraveiro:
Checked the one thing you flagged before merge, and it holds up.

display_order does vary in the seed data. refdata_instrument_codes_populate.sql seeds 115 rows with distinct values running 0, 1, 2, 3 and up, so the combo genuinely changes from unordered to ordered rather than being a no-op dressed as a behaviour change. Good question to have asked, because if they had all been 0 the PR body would have been describing an effect that does not exist.

On the registrars not being called outside their own file and test: agreed it mirrors iam rather than introducing a gap, so nothing changes here. Noting it so it is not lost.

One thing your review did not cover, raised by the operator while this was open, and it is a real gap in what I delivered. Clearing codegen drift is not the same as the component passing a component architecture audit, and I claimed only the former. ores.analytics.quant sits as a dotted sibling of ores.analytics when the signed-off layout is nested parts under a composite root, so it should be projects/ores.analytics/quant/. quant is also not in the signed-off folder catalogue, so it would be a finding on sight. That is 120 C++ files referencing the include prefix and 11 CMake files, so it is its own task rather than something to append here.

PR #2067, opened 2026-09-11
[prototype] Add the structure-based trade entry prototype

mcraveiro:
## Summary

Add a throwaway clickable prototype of a trade entry flow built around structures, where a booking screen manages a structure and its components and a companion pricing screen hands the economics across with its Book button. The prototype exists to make the screen arguable with something concrete before anyone writes production code.

## Changes

- Add doc/prototypes/trade-entry/: two pages, one stylesheet and three scripts, with no framework and no build step.
- Add doc/prototypes/index.org, a gallery listing each prototype and linking it, plus a Prototypes entry in the Developer menu.
- Serve the prototype from doc/prototypes/trade-entry/ and deploy the directory from the site build through the existing ores-deploy-web-app.
- Model the versioning rule literally: one mutating function bumps the structure version and writes the audit line, and authorisation does not.
- Close the product-type research and record its catalogue in the task.
- Separate reference data from economics. The structure header carries an envelope strip, and the component panel carries an envelope block grouped under the field-group names of ores.trading.trade: identity, classification, parties, lifecycle and audit. The instrument form carries the instrument only.
- Split the instrument form into tabs named for what the widget holds, so no tab runs past one screen. The premium renders as a sub-leg with its own identifier, currency, amount and payment date. The audit folds into a closed drawer.
- Take the side chip's class from a known pair. The generic form lets the reader type anything into longShort, so the class was the one value that reached an attribute unescaped. The same change gives the chip the colour the stylesheet defines for it.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Trade entry prototype: structure-based booking screen](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/trade-entry-prototype/story.html) | D63FCA11-EE40-438F-ADF7-12C2ED5541D3 |
| Task | [Build the booking screen prototype](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/trade-entry-prototype/task_prototype-booking-screen.html) | EDA1C2B2-C029-4BFC-8DEF-A15E8B369B4A |
| Task | [Make the site build copy and deploy doc/prototypes](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/trade-entry-prototype/task_site-deploy-prototypes.html) | 07E0C977-798A-4801-A623-E294169D6960 |

## Testing

Plan. Run the site build, which publishes the new org pages and deploys the prototype. Check the numbered checks in each task's Verification section by hand. The prototype's own self-check is reachable with ?selfcheck=1 on either page.

Evidence. The site build finished with "Build succeeded." and exit 0, and the graph export wrote 5301 nodes and 31839 links. The deploy log names graph, agile and doc/prototypes/trade-entry, and diff -rq between the prototype source and its deployed copy shows index.html differing only by the injected chrome. The booking screen passes all twenty-one self-check assertions in headless Chrome with ?selfcheck=1, and the pricing screen passes the seventeen it can run and names the four that need booking.js. Both pages render in headless Chrome with no script error, at their new URLs. A move between groups and the full pricing handoff were driven in headless Chrome and both took effect. The gallery renders in headless Chrome, its link resolves to the app, and the Prototypes nav entry appears on the home page, the knowledge hub and the gallery. The misspell check is clean on the changed paths.

The reference-data split is covered by three new assertions, and the second of them was shown to fail: adding a counterparty field back to the generic instrument form reports one failure. The side-chip assertion was shown to fail the same way: restoring the old class line in a scratch copy reports one failure. The three rendered chips carry side-buy and side-sell, the classes the stylesheet defines. Removing the premium, adding it back and editing a field were driven in headless Chrome, and each took the version with it. The layout was read back from a screenshot at 1500 by 1000: both envelopes and the whole of one tab are visible without scrolling.

Limitations. No C++ source, cmake, codegen template or modelling file is touched by this diff, so the full build and ctest were not run; the one non-doc file is Elisp, and the site build above exercises it. The prototype is throwaway: it has no backend, no real valuation and no live market data, so nothing here is covered by automated tests beyond the in-page self-check. A whole-site nav change does not invalidate the org-publish timestamp cache, so a warm worktree must remove build/output/org-timestamps/site:pages.cache before the build; CI is cold and unaffected. The task records this.


mcraveiro:

mcraveiro:
Fixed in 078e4cce56.

1. Unescaped class attribute — accepted. The side chip's class now comes from a known pair: side-buy for a long, side-sell for a short, and nothing for anything else, so the field's own text never reaches the attribute. A self-check holds the class to the pair, and it was shown to fail: restoring the old line reports one failure. The same change closes a defect the review did not name — the chip emitted side-long and side-short, which no stylesheet rule matches, while side-buy and side-sell were defined and unused. The chip carries a colour for the first time.

2. Assertion count — accepted, and corrected before this round. The old text was wrong twice: the count is twenty, not nineteen, and node never runs the check, because the guard reads window.location. The task record and the PR body carry the true count, which is twenty-one with this round's new assertion.

3. Dead select branch in field() — accepted. Removed. selectField() is the helper for that, and field() is only ever given text, number or date.

mcraveiro:

mcraveiro:
Fixed in 03eeb5cf75.

1. The self-check threw on the pricing page — accepted, and confirmed in headless Chrome before the fix: Uncaught ReferenceError: sectionsFor is not defined, with no verdict line printed. A bookingLoaded flag now gates the four checks that read booking.js. The booking screen still runs all twenty-one assertions; the pricing screen runs seventeen. The four it cannot run are skipped by name, never passed, because a skip that reads as a pass is worth nothing. Both verdicts are read back from the console: 'prototype self-check: all passed' on the booking screen and 'prototype self-check: all passed, 4 skipped' on the pricing screen. The gate did not neuter the checks: restoring the old class behaviour reports '1 failed a side chip takes its class from a known pair'. The task doc, the description page and the PR body no longer claim the check runs the same on both pages.

2. The stray fragment on line 191 — not changed. Line 191 is one complete table row and reads end to end. Your own note records that your grep printed the line as '[Omitted long matching line]', so the fragment is the grep's truncation, not the text.

PR #2068, opened 2026-09-11
[analytics] Nest quant under analytics and register the role

mcraveiro:
## Summary

ores.analytics.quant sat as a dotted top-level sibling of the ores.analytics composite, claiming a parentage the tree did not have, and quant was in neither the sub-component catalogue nor the signed-off folder catalogue. It is now the quant part of that composite, with its role registered in both. The NAME_COLLISION exception that had been keeping the audit green is deleted.

## Changes

- Move projects/ores.analytics.quant to projects/ores.analytics/quant, a pure rename of 129 files
- Point the add_subdirectory line at the nested path, beside its api, core and service siblings
- Register quant in the sub-component catalogue's specialist roles, the signed-off folder catalogue, the composite component registry and the group overview
- Delete the NAME_COLLISION exception, which the audit calls part of the fix
- Repoint nine agile documents and the checker docstring that cited the old dotted path

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Nest quant under analytics and register the role](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_nest_quant_under_analytics.html) | 82B0DD7E-4DB1-4F99-AF73-6FBB6C258786 |

## Testing

Plan. Code class: the doc checks cannot see build wiring, so the move must be proved by configuring CMake and compiling across the component boundary, not only by the audit tool passing.

Evidence. code; cmake --preset linux-clang-debug-make configures clean, exit 0 with no CMake errors; all 61 quant translation units appear in the compile database at the nested path and none at the old one; a quant unit compiles, and so does ores.marketdata crm_ingest_bridge.cpp, which includes quant headers across the component boundary and so proves the include prefix survived; validate_docs passes with the exception deleted (43 components); check_component_drift.py --all clean; check_model_drift.py clean; compass lint clean; misspell-fixer exit 0.

Limitations. No full build or ctest run: the change is a directory rename with no content change, verified by configure plus representative compiles rather than by linking every target. The audit's manual checklist was not walked; only its automated half is covered.


mcraveiro:
@claude please review. This came from the operator noticing that clearing codegen drift in analytics said nothing about whether the component has a legal shape, and it did not.

The thing worth checking hardest is my claim that the move is free at the call sites. My reasoning is that a part's CMake target and include prefix are already the fully-qualified name, so moving the directory leaves `ores.analytics.quant` as both the target and the include prefix and none of the 120 including files change. I verified it by configuring CMake (clean, no errors), confirming all 61 quant translation units moved to the nested path in the compile database with none left at the old one, and compiling an `ores.marketdata` consumer that includes quant headers across the component boundary. If there is a consumer shape that breaks and those three checks would not catch it, that is what I want found.

Second, the registration. The audit says an unregistered folder name is a finding in its own right, so I judged that absorbing the component without adding `quant` to the catalogues would trade one finding for another. I added it to the sub-component catalogue's specialist roles, the signed-off folder catalogue, the composite registry and the group overview. Challenge the role description if it is wrong: I wrote that quant is separated from `core` because it carries no database or NATS dependency and is consumed for its mathematics alone, which I took from its own overview rather than from reading all 61 units.

Third, and this is the part I would not want waved through. The `NAME_COLLISION` exception for this had been sitting in `docs_exceptions.txt` describing the exact fix to apply. The audit was passing by suppression, and nothing re-reads that file, so the finding stayed quiet until a human looked. I deleted the entry and filed `task_wire_component_audit_into_ci`, because the underlying problem is that the audit is not wired into CI the way the codegen drift checks now are. If you think the remaining exception lines hide anything of the same kind, say which.

mcraveiro:
All four accepted; fixed in c08b270645.

The consumer list was the real finding and I had it backwards in the worst way. I wrote 'by core' from quant's own component overview rather than from the build, and in a sub-component catalogue that phrase reads as the sibling core in the same composite. ores.analytics.core does not link ores.analytics.quant.lib at all. I checked every target_link_libraries myself before changing the text and your list is exactly right.

Following that through gives a sharper fact than a corrected list, so the catalogue now states it: no part of ores.analytics links quant. Every consumer is in another group. The nesting is still correct — a dotted sibling claimed a parentage that did not exist, and this one is registered — but the placement records where the mathematics belongs by subject rather than an existing dependency, and the entry now says that instead of implying a dependency that is not there.

Three sub-components against a four-row table: corrected. Worth naming what that was, since it is the same class of defect this PR exists to remove. I added the fourth row and left the sentence above it stale in the same commit, which is documentation drift introduced by the change that was fixing a documentation-shape finding.

On the move-is-free claim, your check is stronger than mine and I will take the correction. I showed it held for 61 translation units and one cross-component consumer; you showed it cannot fail, because the target name is a literal and the include path derives from CMAKE_CURRENT_SOURCE_DIR. Evidence that something cannot break beats evidence that it did not.

On docs_exceptions.txt: thank you for answering the question rather than deferring it. NAME_COLLISION being the only structural code, with exactly one entry and that entry now gone, narrows the follow-up task from a hunt to plain enforcement. The MISSING_PUML entries can still go stale silently, which is the same failure mode at lower stakes, and task_wire_component_audit_into_ci covers them.

Noted that you could not run bearings, validate_docs or CMake in the sandbox. I re-ran validate_docs (43 components pass) and compass lint after these edits.

PR #2069, opened 2026-09-11
[build] Register every composite's parts in its own CMakeLists

mcraveiro:
## Summary

projects/CMakeLists.txt listed each part of each composite, so a component's parts were spread across the file. It now adds the component, and the component adds its parts. Codegen generates the composite root rather than leaving it hand-written: the composite kind was the one no archetype claimed, so the single file a composite owns had nowhere to come from.

## Changes

- Add a root CMakeLists to the 19 remaining composites and collapse projects/CMakeLists.txt from 91 entries to 39
- Add the ores.cmake.component.composite_root archetype, declaring component_kind composite
- Declare part order in the model's #+parts:, since alphabetical breaks ores.qt and ores.eventing
- Expose parts to the template context and correct the loader comment that described a composite as generating nothing
- Update the component architecture doc and the cmake.component facet to the new shape
- File the eight composites that have no group model for codegen to generate from

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Register composite parts in their component CMakeLists](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_convert_composites_to_component_cmake.html) | 4B43941A-957B-4FC9-82BD-AB131F541825 |

## Testing

Plan. Code class. A build-wiring change cannot be read for correctness, so the test is that the configuration comes out unchanged: a baseline of every target was taken before the change and compared after.

Evidence. code + python tooling; cmake --preset linux-clang-debug-make configures clean, exit 0, zero CMake errors; target set identical to the pre-change baseline at 93 directories and 3957 translation units, none missing and none new; the generator reproduces all twelve declared composites' roots byte-identically; check_component_drift.py --all clean; check_model_drift.py clean; validate_docs 43 components pass; codegen pytest 180 passed; compass lint clean; misspell-fixer exit 0.

Limitations. No full build or ctest: no compilable source changed, only which CMakeLists adds which directory, and the target-set comparison covers that. The eight composites without a group model keep hand-written roots, so those are not generator-verified.


mcraveiro:
@claude please review. This restructures how every component is registered with CMake, so the risk is a silent build change rather than a visible one.

The claim to attack hardest: the configuration is unchanged. I took a baseline of every directory and translation unit in the compile database before touching anything, and after the change it is identical — 93 directories, 3957 translation units, none missing, none new, and configure exits 0 with no CMake errors. If a target can disappear or change in a way that comparison would not see, that is the failure I want found.

The specific ordering risk is worth checking independently. Parts of a composite were scattered: ores.history had api at line 5 and core at line 23, ores.telemetry spanned 37 lines, and every service part sat far below its siblings. Collapsing means each component takes the position of its first part, so service parts now get added much earlier than before. I reasoned that CMake resolves target names at generate time so link order does not matter, and that the identical target set plus a clean configure confirms it. If there is a case where add_subdirectory order matters for something other than target resolution — a variable set in one directory and read by a later sibling — that would not show up in either check.

On codegen: composite was the one component kind no archetype claimed, and the loader said so explicitly, that a composite generates nothing from either facet by design. So the single file a composite owns had nowhere to come from. The new archetype fills exactly that slot. Part order is declared in #+parts: rather than derived, because alphabetical would put ores.qt's plugins ahead of the headless and api parts they link, and would drop ores.eventing's modeling part. Challenge that if a derived order would in fact work.

Two things I did not do. Eight composites have no group model, so their roots stay hand-written and are filed rather than generated. And I reused ores.eventing's hand-written licence header when writing the new roots, which wraps differently from codegen's — the drift gate caught it before I pushed, which is the second time in three PRs that gate has caught my own work.

mcraveiro:
All five accepted; fixed in 40c7a53db9.

The cross-check finding is the one I asked for and did not find myself, and your framing of why it hides is exactly right: configure cannot fail over a directory it never hears about, and the drift check compares generated output against the same wrong model. Two gates, neither able to see it.

resolve_targets now raises when a composite's declared parts disagree with its hosted sub-components, in either direction. I proved it by injecting both failures into a real model rather than reasoning about it. Dropping quant from ores.analytics reports that a sub-component on disk would stop being built; misspelling it reports twice, as the missing part and as a name that builds nothing. A part naming something that is not a sub-component is still allowed when it carries its own CMakeLists.txt, which is how ores.eventing pulls in its modeling diagram target — that case has its own test so the allowance cannot rot into a hole.

One thing your review surfaced indirectly that I should name. test_container_declaring_composite_passes asserted that a composite generates nothing. That was true while no archetype served the kind, and this PR makes it false. Left alone it would have kept passing for the wrong reason, so it is updated and three tests are added for the new gate.

On deriving part order instead of declaring it: agreed it would be strictly more correct, and I am keeping the declaration. What changed is the cost of getting it wrong — a bad declaration now fails generation instead of quietly building less, which is most of what worried me about declaring it.

On the collapse itself, thank you for checking marketdata, iam and compute byte-for-byte against the old flat list. That is the claim my baseline comparison was making at the aggregate level, and it is better evidence for having been checked per-component by someone who did not write the script.

mcraveiro:
Operator ran a full local build on this branch and it is green.

That closes the gap this PR declared rather than hid. The testing limitations said no full build or ctest, on the grounds that no compilable source changed and the target-set comparison covered which CMakeLists adds which directory. A real build is the check that settles it, and it agrees with the comparison: 93 directories and 3957 translation units, none missing, none new.

Recording it here rather than as a commit, so the evidence is on the PR without spending a CI cycle on a change with no content.

PR #2071, opened 2026-09-11
[codegen] Bind custom C++ types to the headers that define them

mcraveiro:
## Summary

A column could already declare any cpp_type, but the header defining it was listed separately by hand with nothing connecting the two, so the include could be forgotten or left behind when the type changed. A registry now binds them. Eight of the nine registered types are already in use; cron_expression is the ninth, registered for the scheduler resync that motivated this.

## Changes

- Add projects/modeling/cpp_custom_types.org, binding each custom type to its header
- Resolve wrapped forms such as an optional or vector of a registered type to the same header
- Add only missing headers and never reorder, which is what keeps every model predating the registry byte-identical
- Reject a registry binding one type to two headers, since a bare domain:: name is component-relative
- Audit a component before closing its task rather than after — the ordering that let analytics close on codegen drift alone

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Bind custom C++ types to the headers that define them](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/task_bind_custom_types_to_headers.html) | 3B57DD4B-E6CB-48CC-803A-74834270AB9E |

## Testing

Plan. Docs plus python tooling. The registry must both change nothing that exists and actually supply a header a model stops declaring, so it needs two opposite verifications rather than one.

Evidence. docs + python tooling; site build succeeded (exit 0, 5313 nodes, no errors — the five error matches in the log are filenames); check_component_drift.py --all byte-identical, so the registry agrees with every include a human wrote; deleting the asset_class include from feed_binding's model left the generated header unchanged and still including it, proving the registry supplies it; codegen pytest 192 passed; compass pytest 159 passed 1 skipped; check_model_drift.py clean; validate_docs 43 components; compass lint clean; misspell-fixer exit 0.

Limitations. No C++ build: no generated C++ changed, which the byte-identical regeneration establishes. The component audit's manual checklist was not walked, because the only component touched is ores.codegen, a tool component the checklist does not govern.

### DIFF
--- a/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/story.org
+++ b/doc/agile/versions/v0/sprint_25/unify_llm_skills_catalogue/story.org
@@ -112,7 +112,7 @@ Tasks are listed in intended execution order.
 | [[id:C8C1316A-07A5-47E2-9077-FF0664C511AF][Run the component architecture audit in CI]] | BACKLOG | | | validate_docs.py is the audit's automated half and no workflow runs it; its exception file kept a real finding quiet until someone looked by hand. |
 | [[id:4B43941A-957B-4FC9-82BD-AB131F541825][Register composite parts in their component CMakeLists]] | DONE | 2026-09-11 | 2026-09-11 | Eighteen composites still register each part in projects/CMakeLists.txt; the target shape adds the component there and its parts in the component's own CMakeLists. |
 | [[id:9747E4C5-734E-4F75-AD8B-CEB4C9089B76][Give every composite a group model]] | BACKLOG | | | Eight composites have no group-level component_overview.org, so codegen cannot generate their root CMakeLists and their parts are undeclared. |
-| [[id:3B57DD4B-E6CB-48CC-803A-74834270AB9E][Bind custom C++ types to the headers that define them]] | STARTED | 2026-09-12 | | A column can declare any cpp_type but its header is listed separately by hand, so the two can drift apart or the include can be forgotten entirely. |
+| [[id:3B57DD4B-E6CB-48CC-803A-74834270AB9E][Bind custom C++ types to the headers that define them]] | DONE | 2026-09-12 | 2026-09-12 | A column can declare any cpp_type but its header is listed separately by hand, so the two can drift apart or the include can be forgotten entirely. |
 
 * Decisions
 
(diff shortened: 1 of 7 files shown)


mcraveiro:
@claude please review. This is the reusable half of the scheduler resync: a registry binding a custom C++ column type to the header that defines it.

The claim worth attacking is that this changes nothing existing. Regenerating all eight registry components is byte-identical, which I read as the registry agreeing with every include a human wrote by hand. If byte-identity could hold while the mechanism is silently inert — matching nothing, or matching and being discarded downstream — that is the failure to find. I tried to close it from the other side by deleting the `asset_class` include from `feed_binding`'s model and regenerating: the generated header was unchanged and still included it. Those two together are my evidence, and I would like the second one checked hardest, because it is the only proof the registry does anything at all.

Second, the matching rule. I match a registered type name as a token inside the cpp_type string so `std::optional<cron_expression>` resolves, and I reject a hit where the adjacent character is alphanumeric or underscore so `prefixed_cron_expression` does not. If there is a real C++ type spelling that should match and does not, or one that should not and does, that is a bug in the rule rather than the registry.

Third, a design choice worth challenging. A bare `domain::asset_class` is component-relative, not globally unique, so a single global registry is only safe while no two components declare the same short name. I made ambiguity a hard error rather than scoping the registry per component, on the grounds that the corpus has no collision today and an error is honest about the limit. Scoping it properly would be more correct and more machinery. Say if the error is not enough.

One process change rides along, and it is an admission. The audit already listed zero codegen drift as a trigger and said to audit right after the drift story closes — that ordering is exactly what let me close ores.analytics on codegen drift alone and report it as clear. It now says before the task closes, and it is the close-task skill's first step.

mcraveiro:
Both findings accepted; fixed in 4f3cc530be.

The leftmost-occurrence bug is real and your repro is exact. Joining the types into one string and boundary-checking a single position was the mistake, so the fix is not better index arithmetic — each column is now searched separately, and every occurrence within it. The failure is gone by construction rather than patched.

The part worth more than the bug is your note on why my test could not have caught it. test_a_longer_identifier_is_not_a_match only ever put one type in the list, so the hidden-later-occurrence path was never exercised. The test proved the rule against the cases I had already thought of, which is the failure mode it exists to prevent. Your case is now a regression test.

The :: finding I had not seen at all, and I have fixed it rather than documented it. A preceding :: is now a boundary, so a name qualified by another namespace is a different type. Two consequences worth stating: the ores::trading::domain::product_type row becomes load-bearing rather than redundant, which is the right outcome since each spelling in use should earn its row; and the silent-resolution path you describe is closed, where an unrelated type ending in ::product_type would have taken the trading header without the one-type-one-header check ever seeing it, because that check reads the registry's keys and not what matching does at runtime.

On the deletion experiment: agreed it holds and agreed it is on a different axis. It proves the registry acts, and says nothing about which names it matches. I had been treating it as stronger evidence than it is.

Regenerating all eight registry components is still byte-identical under the stricter rule, and the suite is at 195.

PR #2072, opened 2026-09-12
[ore] Sync ORE examples and XSD to v1.8.17.0

mcraveiro:
## Summary

Move the vendored ORE examples and XSD from v1.8.16.0 to v1.8.17.0, and regenerate the C++ domain types from the new schema. The v1.8.17.0 engine package is committed beside the v1.8.16.0 one, and the compute seed registers both versions so provisioning can publish each.

## Changes

- Sync external/ore/examples/ and external/ore/xsd/ from the v1.8.17.0 engine checkout. Both manifest.json files record the provenance. The only deletions are 13 Emacs auto-save files that a previous hand-vendoring committed by accident.
- Regenerate the C++ domain types from the new XSD. Three compile breaks followed, and all three are fixed: commodityOptionData moved Strike into the strikeGroup substitution group; five FX types turned OptionData and BarrierData from repeated to single; sensitivityanalysis.DiscountCurves became optional without changing its name.
- Add xml_example_coverage_tests, which walks all 132 files in Products/Example_Trades and requires a golden round-trip and a test reference for each one. Credit_CallableBond.xml was the only gap; its golden is committed with it.
- Seed the v1.8.17.0 app version and package in the compute seed, so ACME provisioning publishes both engine versions.
- Whitelist the codegen venv in .misspell-fixer.ignore, so a root-level misspell run exits 0 on a clean tree.
- Verification: code + docs + ci; build clean (linux-clang-debug-make); ctest 71/71 passed; round-trip exit 0; cmake-sources and codegen drift clean; site build clean; misspell exit 0.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Update ORE to v17](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/update-ore-to-v17/story.html) | A34F4835-5CA5-4F30-91D6-90975354534C |
| Task | [Sync ORE examples and XSD to v1.8.17.0](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/update-ore-to-v17/task_sync-ore-examples-and-xsd.html) | 25EC2B31-812F-4EB1-8ADE-966DF496D5E8 |

## Testing

Plan. The branch mixes code, docs and ci, so the union of the three check sets ran against commit b772be2e43. The commits after it touch doc only.

Evidence. Build clean on linux-clang-debug-make. ctest: 100 percent passed, 0 failed out of 71, 777.59 s. Round-trip script exit 0. cmake-sources drift exit 0. Codegen component drift: no drift, regenerated output matches the checked-in tree. Site build: Build succeeded, 5318 nodes and 31901 links. misspell-fixer .: exit 0, nothing to replace. The first ctest round failed eight suites on local environment alone: three named schema the worktree database had not applied in 15 days, and five could not reach NATS because the services were stopped for the database recreate. The final round is the one recorded here.

Limitations. The v1.8.17.0 tag on the engine fork is provisional. Upstream had not tagged when this work started, so the tag sits at the sync commit 3d75a6908; the package rebuilds once upstream tags. assets/test_data/domain_roundtrip is deliberately left out of step with the new examples tree, tracked by the backlog item on regenerating or retiring it. The compute seed is verified only inside a rolled-back transaction; no provisioned database or grid run exists yet, which is task ADA8C792. Seven trade types are new in v1.8.17.0 and none has a mapper or a round-trip test, tracked by task 64CB13D8. The CDash experimental run passed 70 of 71 suites; ores.reporting.core.tests failed, that suite is not in this branch diff, and the failure is captured in the product backlog inbox.


mcraveiro:

mcraveiro:
Review round 1 — decisions on the two notes. Both declined, with evidence.

**`make_barrier_group` returns an empty group** (`fx_instrument_mapper.cpp:231`).

Not reachable from real domain data, so no guard was added:

- `projects/ores.sql/create/trading/trading_fx_barrier_option_instruments_create.sql:70-71` constrains `barrier_type <> ''` and `lower_barrier > 0`. A persisted instrument therefore always carries a positive lower level.
- `projects/ores.trading/api/src/generators/fx_barrier_option_instrument_generator.cpp:52-53` is the only other producer, and it hard-codes `barrier_type = "UpAndIn"` with `lower_barrier = 1.2`.

So `has_lower` is true for every instrument that reaches the reverse mapper, the early return never fires, and the old vector-based code and the new helper agree on every reachable input. You were right that the state would be a behaviour change if it occurred — it just cannot.

**`std::stod` on the schema-provided strike string** (`commodity_instrument_mapper.cpp`).

Confirmed inherited, as you said. The new read mirrors `equity_instrument_mapper::extract_strike` (`equity_instrument_mapper.cpp:211-220`) exactly, and the bond mapper follows the same shape. Changing only this copy would leave the three mappers inconsistent, so it stays as is.

Nothing pushed for this round: both notes are declines, so there is no fix to carry.

PR #2073, opened 2026-09-12
[ore,trading] Round trip the bond products and envelope at zero loss

mcraveiro:
## Summary

The bond products and the trade envelope now round trip at zero loss on the mapper path, over every document in external/ore/examples. The losses were never in the xsdcpp-generated binding, whose reader and writer are symmetric. They were in the hand-written container, which flattened xsd::optional into a plain member and so collapsed a stated-but-empty element into an absent one, and which never held most of legData. The container now carries all 18 leg members with a presence flag each, the database carries the shared tables those members need, and the shell carries the verbs that drive the round trip. Testing that path end to end surfaced a second defect: reflect-cpp writes a std::variant untagged and reads it back in declaration order, so every instrument decoded as std::monostate and was silently dropped. The wire now carries an explicit instrument_payload.

## Changes

- Grow the bond container to the full legData shape: 18 members, a presence flag per member, and emit-on-presence in the reverse mappers, so a stated-but-empty element survives.
- Add the shared schedule, option, strike and residue tables, generated from the modeling orgs, and point the mapper's read and write paths at them.
- Make scripts/ore_mapper_roundtrip_diff.py product-aware: the bond products plus the envelope report in scope, the other instrument families report separately as recorded debt rather than failing the run.
- Add the ores.shell ore upload, import and export verbs, plus the .ores script that drives the full XML to database to XML round trip through the service layer.
- Replace the std::variant on trade_export_item with an explicit instrument_payload, because reflect-cpp writes a variant untagged and reads it back by trying the alternatives in declaration order.
- Align bond_issue with the ORE data model: drop the maturity_date and description columns, and carry the option exercise dates as a list in the option block.
- Unify the NATS request timeout defaults behind one constant, and surface workflow step logs in the shell.
- Route the two throwing parses in the bond mapper through the file's own count_of and number_of helpers, so a malformed value no longer throws and costs the whole product block.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Rework the ores.ore bond mapping to the relational model](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_rework-bond-ores-mapping.html) | 212862C4-65C5-470D-99DC-09C23448FE4D |

## Testing

Plan. Code class, with docs, ci and python tooling alongside, so the union of those four classes' checks: full build and ctest on linux-clang-debug-make; the mapper roundtrip diff gate over external/ore/examples; the domain roundtrip snapshot check; the codegen, cmake-sources and shell-script drift checks; the compass pytest suite; and the site build.

Evidence. Build clean on linux-clang-debug-make. ctest 71 of 71 passed, 915.31 sec. The site build succeeded. The mapper gate over 1,566 documents: in scope, bond products and the trade envelope, lost 0 and unexplained 0, exit 0. Codegen drift: no drift. cmake-sources drift: all component_files.cmake up to date. Shell-script drift: tangling 99 scripts left no diff. Doc lint: filetags and generator markers valid. Compass pytest: 129 passed, 1 skipped. misspell-fixer clean outside the vendored venv. The C.3 shell round trip ran live over two documents.

Limitations. Several things are untested. No document in external/ore/examples is a BondFuture trade, so the future paths are corpus-untested; a crafted probe covers them and shows one remaining loss, recorded as task A3DA0BB5, where an element stated with a value equal to its type default is dropped on export. Windows and MSVC are unmeasured, and bond_instrument_data reflects 28 fields against the 19 that already tripped MSVC C1202 in sprint 19. The branch does not merge cleanly with origin/main: two files conflict, the sprint.org bookkeeping and a clang-format rewrap in bond_instrument_mapper.cpp; everything else auto-merges. Two recorded debts are unchanged by this branch: format_number writes 1e+05 for 100000.0, though the value reparses to the same double and the gate counts it classified, and the import's book creation is not request-id-scoped. scripts/ore_domain_roundtrip_check.py still reports a stale committed snapshot and is non-strict, so it gates nothing; the gate that matters is ore_mapper_roundtrip_diff.py. CDash experimental was not run, because it repeats the same suite for visibility only.


mcraveiro:

mcraveiro:
Review round addressed at `78f0810655`. Ten items, five accepted and five declined with reasons.

**1. Accepted.** Both catch blocks in `count_of`/`number_of` now log a `warn` carrying the raw text and the value used. The two helpers are free functions in the file's anonymous namespace, so they cannot reach the class's private `lg()`. They get a file-local `logger_name` and `lg()` accessor instead, matching the pattern used in `publish_from_dq_handler.cpp`. The build caught this on the first attempt: the log calls as originally written did not compile, so the fix is now verified rather than assumed.

**2. Not changed.** Pre-existing, one layer below this work, and outside the bond scope. It is the same silent-drop class, so it is recorded as debt on the task rather than fixed here.

**3. Accepted.** Scope-limit comment added above `map_rate_group`. `legDataType` is an abstract element with eighteen substitution-group members; three of them have a container here, so a document that states one of the other fifteen round trips to an empty element.

**4. Not changed.** Pre-existing and already a stated scope limit. Rule-based call schedules stay out of scope for this work.

**5. Not changed, deliberately.** The loss is real and is tracked as task A3DA0BB5. A test written now would either assert the drop as correct or fail the suite. It belongs with that task's presence-flag change, which is where it will land. Declining this one is a judgement call and it is recorded as such.

**6. Accepted.** `static_assert(std::variant_size_v<trade_instrument> == 9)` now sits next to the calls, so a tenth family fails the build instead of passing silently.

**7. Not changed.** Documented accepted risk: the services restart together and old blobs are re-imported.

**8. Accepted.** The header comment now claims idempotence, which is what the test checks, and it names `scripts/ore_mapper_roundtrip_diff.py` as the field-level check.

**9. Not changed, and confirmed not a defect.** The XSD really does differ behind the shared names. `Payer` is `xs:string` in `bondData` (`instruments.xsd:398`) and `xs:boolean` in `legData` (`instruments.xsd:928`). `paymentLag` is `xs:integer` in `conventions.xsd:104`, but a period string where `legData` uses it (`instruments.xsd:932`, defined at `ore_types.xsd:31`, described in the schema as "either integer or a period 1D, 5M, 10w, 20Y"). The tables mirror the source, so the divergence is intended.

**10. Accepted.** Both defaults in `ores.shell/include/ores.shell/app/request_helpers.hpp` now use `default_request_timeout`, so the unification holds and there is no second literal left to drift.

**Verification at `78f0810655`.** Build clean under `linux-clang-debug-make` with 0 errors and 0 warnings, ctest 71/71, and `scripts/ore_mapper_roundtrip_diff.py` over `external/ore/examples` exits 0 with the bond products and the trade envelope at lost 0 and unexplained 0 across 1,566 documents. One analytics test case needed a local test-database refresh before it passed: the restored database carried a validation that main has since removed, so the local function was stale rather than the test being wrong. No source change was needed for it.

PR #2079, opened 2026-09-14
[build,ci] Turn the red GitHub Actions builds green on all three OSs

mcraveiro:
## Summary

Main's GitHub Actions runs are red on every OS for six independent reasons. This PR carries one fix for each, so the continuous and nightly workflows can go green again.

## Changes

- Rename the ascot fact member to ascot_row in bond_instrument_data, because gcc rejects a member whose name repeats its own type name.
- Rewrite decode_with_legs as the try_one_with_legs and try_each_with_legs helper pair, because clang 19 and AppleClang 16 segfault inside the fold expression the old shape used.
- Disable C4251 for MSVC across the tree, because it promotes the warning to C2220 on every exported class that carries a std::string or a container. The comment beside the flag records that scope and its reason.
- Silence the unused-function warning on the generated domain.cpp for Windows clang, with the matching /wd4505 for MSVC, both scoped to that one generated file.
- Move -performance-avoid-endl from the invalid Warnings key into Checks in .clang-tidy, so every clang-tidy run stops aborting.
- Install the dev packages and Qt6 in the Nightly Include Hygiene job, which failed at configure before any check could run.
- Trim the bond future exporter test document to the three elements the v17 schema keeps on the trade side, so ctest passes once the build fixes land.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Hotfix: CI builds red across all OSs on the payload decode fold and the ascot member](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ci-builds-red-payload-fold-hotfix/story.html) | 8D65A290-92C2-4C02-85DE-88C78A7C66E8 |
| Task | [Implement Hotfix: CI builds red across all OSs on the payload decode fold and the ascot member](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ci-builds-red-payload-fold-hotfix/task_implement-ci-builds-red-payload-fold-hotfix.html) | 1D2C0DD2-EC91-4807-9EFD-1FDFAA3A96DC |

## Testing

Plan. Classify the change as code and ci, then run the union of the class checks on the rebased tree, and reproduce each compiler diagnostic with a local probe before and after the fix.

Evidence. code and ci; on the rebased tree at main 34ed6cf1: the full build under linux-clang-debug-make is clean; ctest passes 71 of 71 suites and the CDash experimental submission succeeds; the ORE domain roundtrip check exits 0; the cmake-sources and codegen drift checks report no drift; the doc lint is clean; the site build succeeds; the seven changed C++ files are clang-format clean under clang-format 21.1.8; a gcc 16.2.0 probe reproduces the -Wchanges-meaning diagnostic on the pre-fix header and compiles clean on the fixed one; a clang 19.1.7 probe reports a compiler bug on the pre-fix file and exits 0 on the fixed one.

Limitations. MSVC and Windows clang cannot run on this machine, so the C4251, /wd4505 and unused-function fixes are predictions from the CI diagnostic text, and a further Windows error may sit behind the first. The Nightly Include Hygiene job dies at configure, so the .clang-tidy repair is proven locally only, by running clang-tidy with the pre-fix and post-fix config. The bond future test document failure is pre-existing on main, and the red builds were masking it.


mcraveiro:

mcraveiro:
Round handled. Per finding:

1. No change. The rename is complete and the Doxygen note records why the name must differ from the type it names.
2. No change. The suppression is per-file and Windows-only, which is the smallest change that clears the diagnostic on a generated file.
3. No code change, and you were right that the scope deserved a plain statement. The tree-wide scope is deliberate. C4251 fires on every exported class that carries a std::string or a container, which is most of the tree, so narrowing the flag to ores.platform would re-break the MSVC build at the next such class. The comment beside the flag records that reason. I have corrected the PR body, which had implied the flag was scoped to ores.platform.
4. No change. The check keeps its leading minus, so it stays off.
5. No change. The document matches the v17 schema, and the sibling mapper suite asserts the relocated terms stay at their defaults.
6. No change. Both nightly steps are byte-identical to the continuous-linux ones.
7. No change. The flagged hunks are pre-existing and formatting is not a build gate here.

No code change came out of the round, so no commit was pushed.

PR #2080, opened 2026-09-15
[ores.compass,ores.ore] Fix the red Continuous Windows and Linux gcc jobs

mcraveiro:
## Summary

Main's continuous builds are red on two independent defects, and this PR carries both fixes. Continuous Windows stops in its Setup Database step with exit 127, because the wrapper runs the venv interpreter by the POSIX name and a Windows virtual environment holds no such file: CPython's venv module writes python.exe there whatever interpreter name creates the environment. The trigger is the runner image's move from Python 3.12.10 to 3.14.7; the wrapper and the workflow are byte-identical across that boundary. Continuous Linux fails in both gcc jobs inside the xsdcpp-generated parser at projects/ores.ore/core/src/domain/domain.cpp, on two gcc-only diagnostics that clang does not raise. The macOS jobs are already cleared by PR #2079, whose run on d780d75eeb is past the point where it used to crash.

## Changes

- compass.sh resolves the venv interpreter once, by probe over python3, python3.exe, python and python.exe, and drives both pip installs and the compass CLI through the resolved path. Linux and macOS keep python3, so nothing changes for them.
- projects/ores.compass/tests/test_venv_python_probe.py lifts the probe out of the wrapper and runs it over four venv layouts plus a venv that holds no interpreter.
- projects/ores.ore/core/src/CMakeLists.txt adds -Wno-implicit-fallthrough and -Wno-unused-but-set-variable for the generated domain.cpp, behind a GNU-only guard, beside the existing per-file suppressions. No generated file is hand-edited.
- Two hotfix stories and their tasks record the root cause, the probes, the local verification and what remains unproven.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Hotfix: Continuous Linux gcc jobs red on two constructs in the generated domain.cpp](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ci-builds-red-gcc-domain-warnings-hotfix/story.html) | AE904D12-B82E-48AA-B270-589B91628953 |
| Task | [Implement Hotfix: silence the two gcc warnings in the generated domain.cpp](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/ci-builds-red-gcc-domain-warnings-hotfix/task_implement-ci-builds-red-gcc-domain-warnings-hotfix.html) | 4B10F3F0-4983-457D-89FC-22E59FD2A788 |

## Testing

Plan. Neither defect reproduces on this host, so Continuous Windows and Continuous Linux are the real proof. Locally: the wrapper probe over simulated venv layouts, a real compass run on Linux, the generated file compiled with g++ from its own compile command, and the class checks for a mixed diff.

Evidence. Mixed class: code, python tooling and docs. clang debug build 100 percent with zero warnings and zero errors; ctest 71 of 71 passed in 897 s; site build succeeded; cmake-sources drift clean; codegen drift reports no drift; compass suite 164 passed, 1 skipped; the domain roundtrip check exits 0; misspell clean on the changed paths. The gcc build compiled domain.cpp with the suppressions in place and raised no diagnostic for it. The macOS run on d780d75eeb is past the point where it used to crash.

Limitations. The Windows job cannot run here, so the wrapper fix is exercised only through the simulated venv layouts and the Linux run. The gcc tree has had no green run since -Wall and -Wextra landed on 2026-09-11, and the local gcc build was stopped before the end of the tree, so further gcc-only fallout may sit behind this error; Continuous Linux is the proof. CDash experimental was not run locally, because the build and test it submits had just run here. The three continuous workflows are scheduled, not PR gates, so this pull request carries no build check of its own; the proof lands on the first scheduled run after the merge, within two hours for Linux and twelve for Windows.


mcraveiro:

PR #2082, opened 2026-09-16
[ores.shell] Split ores.shell into api and application parts

mcraveiro:
## Summary

ores.shell was one component with one CMake target, so a domain adapter could not link the shell plumbing without the REPL host, and codegen had no part to write a generated command unit into. This splits the component into an api part and an application part behind a generated composite root. The api part carries the plumbing that other components link against. The application part carries the REPL host and the command units. The change also runs the recurring component audit against the result. It is the prerequisite for the per-domain adapter parts and for the shell-command codegen wave.

## Changes

- Split ores.shell into an api part and an application part behind a generated composite root. The api part holds the command arguments, feedback, pagination and request plumbing. The application part holds the REPL host, the command units, the config parsing and the script runner, and it carries the ores.shell binary.
- Repoint the build and every consumer at the new targets. ores.qt.application links ores.shell.application.lib, the include paths and CMake source lists follow the new layout, and each test binary sets its own log module name.
- Declare ores.shell a codegen composite. Add its row to the component catalogue and to the known-drift-free registry of the drift gate, so the generated root has a regeneration path and drift coverage.
- Record the Application-layer naming rule in the component architecture model, and rewrite the shell section of the application system model.
- Close the two split tasks and the story, and record the four review findings with their decisions in the task and the component overview.
- Record the measured build-slice peak in the slice drop-in comment. One application translation unit held 6.76 GiB RSS, which crosses the 8 GiB memory.high cap on its own.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Split ores.shell into composable parts](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/split_ores_shell_into_parts/story.html) | A52DCD72-BEF4-4655-9987-A3103B288576 |
| Task | [Split ores.shell into api and application parts](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/split_ores_shell_into_parts/task_split_into_api_and_application_parts.html) | 2915CF39-6936-4942-97C1-F0C11247F11E |

## Testing

Plan. The class is code (C++), docs and python tooling, so the union of their checks runs. Full build and ctest on the linux-clang-debug-make preset, submitted to CDash as Experimental. The ORE domain roundtrip check. The cmake-sources drift check with --all --check. The codegen drift check over the whole known-drift-free registry. The compass pytest suite. A site build for the org and modeling edits.

Evidence. The build step had no errors. The test step passed 72 of 72 tests and 0 failed. Both shell test binaries ran in that sweep. ores.shell.api.tests passed 49 assertions in 16 test cases, and ores.shell.application.tests passed 31 assertions in 10 test cases. The run submitted to CDash Experimental as build 4102474. The codegen drift check over the known-drift-free registry reports no drift, and it regenerates the shell composite root with no difference. The cmake-sources drift check reports all component_files.cmake up to date, including the two new shell part trees. The ORE domain roundtrip check exits 0. The compass pytest suite reports 164 passed and 1 skipped. The site build succeeds. The misspell check over the 210 changed files exits 0.

Limitations. The gcc tree cannot build in this worktree. The build slice sets memory.high to 8 GiB, and one application translation unit holds 6.76 GiB RSS on its own, so the kernel throttles the compiler into uninterruptible sleep and no job count avoids it. The clang preset is the one that fits the cap, and it is the preset the evidence uses. macOS and Windows were not run locally; the parts stay static on Windows too, which is the thing for CI to confirm.


mcraveiro:

mcraveiro:
No code change from this round. The review found no functional issue, and both flagged items are already recorded rather than lost. The part include prefixes and the scripts placement sit under the Recorded entries in projects/ores.shell/modeling/component_overview.org, and the part-generation trap sits under Recorded in the same file's Review round. The first two are open decisions carried in the split task, and the trap belongs to the command codegen wave. CI on this PR is green for lint, drift and model.

PR #2083, opened 2026-09-16
[ores.shell] Land the shell-command facet and the trading part

mcraveiro:
## Summary

The shell's per-entity command units get a codegen facet and an adapter part to receive them. The facet turns one trading entity model into a registered command unit, and the first unit (fx_forward_instrument) lands in projects/ores.shell/trading.

## Changes

- Add the ores.cpp.shell-command facet with its two archetypes, command_header and command_implementation, whose output paths aim at projects/ores.shell/{component}/.
- Declare adapter a component kind the loader understands, and give the two file-list archetypes that kind, so an adapter part receives its file lists and no other archetype.
- Teach the generator the :supplied_by: annotation on entity columns, so the add verb reads one positional per user-supplied column and the model states which those are.
- Create projects/ores.shell/trading as the composite's third part, with its overview, diagram, file lists and tests.
- Point fx_forward_instrument at the facet and generate the first command unit into the part.
- Record the wave in the sprint, the story and the two tasks, and capture the template pages that sit outside the ores.* graph.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Generate shell command units per component](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/shell-command-codegen/story.html) | 7C9B6A2E-F1D1-4BA2-9960-C36359B77B9A |
| Task | [Land the shell-command facet and the ores.shell.trading part](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/shell-command-codegen/task_land-shell-command-facet-and-part.html) | 824C610A-E8A9-4BE2-B8A0-59527458E66B |

## Testing

Plan. Mixed change. Classes and their checks: code (full preset build plus ctest), ci (site build, domain roundtrip, codegen drift), python tooling (compass pytest suite plus drift). The codegen drift gate runs for the shell component; trading-cpp stays outside the known-drift-free registry until the trading estate is reworked.

Evidence. Code class: preset build clean; ctest 73/73 passed in 938 s, including the new ores.shell.trading.tests. ci class: domain roundtrip check exit 0; regenerate_cmake_component_files.py --all --check reports every component_files.cmake up to date; check_component_drift.py --component shell reports no drift after a tangle run. Python tooling class: compass pytest 164 passed, 1 skipped; misspell-fixer reports no proposed change across this branch's files. Site build succeeded over the branch content, with a confirming rebuild running at raise time over the last three prose edits.

Limitations. The hand-written trading command units still carry the shell surface, so the part runs beside them; retiring them is the next task. The facet's delete and history verbs read one key string, so compound-key entities are not generated yet, and the add verb's audit assignments are unconditional, so entities with no audit group are not generated yet. Both gaps are recorded on the follow-up task. Two tree-wide misspell hits sit outside this diff, in the update-ore-to-v17 task docs that quote an upstream identifier; they are already recorded as declined.


mcraveiro:

mcraveiro:
Fixed in 362d5c206a and f5bca20569.

Finding 1. `from_token` gains an overload that takes the column name, every conversion in the implementation archetype passes it, and the construction block now sits in one `try` that reports through `fail(out)` and returns. A malformed token names its field instead of surfacing a raw lexical cast message. The archetype was re-tangled and the unit regenerated.

Finding 2. The test file for the part now has seven cases that drive all three handlers offline. They cover the login guard on add, delete and history, the arity check, a malformed bought_amount that names its field, the empty default party, and a valid token vector that reaches the transport. The follow-up task records that the facet has no tests archetype yet.

Local checks on the pushed head: build clean, ctest 73/73 passed, codegen drift check clean.

PR #2087, opened 2026-09-17
[codegen] Mark generated C++ files as generated

mcraveiro:
## Summary

Generated SQL names its template in an AUTO-GENERATED block; generated C++ said nothing, so nothing told a reader that a file was codegen output — which is how a generated file gets hand-edited. The marker is now emitted for every generated C++ header and implementation. Found while sizing the instrument-identity collapse, where the generated/hand-written split could not be read off the 815 files referencing instrument_id.

## Changes

- Adds the AUTO-GENERATED marker at the render seam in codegen core, naming the template that produced each file
- Guards the marker so it reaches only .hpp and .cpp outputs, not any non-C++ output that carries the C++ licence
- Regenerates the nine drift-free components, materialising the marker in 3557 files

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Mark generated C++ files as generated](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_mark-generated-cpp-files.html) | FF378E87-BDF9-412A-A3CC-B58EC61296B4 |

## Testing

Plan. Python tooling plus code class: compass test suite, the codegen drift gate over the registry, a full build to prove 3557 changed C++ files still compile, and an audit of every added and deleted line in the diff.

Evidence. Compass test suite 164 passed, 1 skipped. Codegen drift gate clean across all nine registry components: regenerated output matches the checked-in tree byte for byte. Diff audit: outside core.py the diff adds only the five-line marker block and deletes nothing, across 3558 files. misspell-fixer reports no candidates.

Limitations. The full build (linux-clang-debug-make) was still running when this was raised; it is the only check that proves the comment lands legally in every file, and the result will be posted on this PR. ctest has not been run — the change adds comments only and no test behaviour is touched. The nine components outside the drift-free registry keep unmarked generated C++ until they can be regenerated without dragging in pre-existing drift; ores.trading gains its markers in the instrument-identity work that follows.


mcraveiro:

mcraveiro:
Round 1 addressed in 45d3e67 and pushed as 50bc28b2ae. Both observations accepted.

1. No unit test for the helpers — accepted, and your reasoning is the part worth recording. The drift gate proves regeneration is stable against the checked-in tree, which stays true if the marker stops being emitted and the tree is regenerated to match, so it cannot catch this regression at all. Seven tests added in projects/ores.codegen/tests/test_generated_marker.py: the marker names its own template, it follows the licence and precedes the body, both .hpp and .cpp are marked, a non-C++ output carrying the C++ licence is not marked, a template with no cpp_license is untouched, _emits_cpp classifies by output suffix, and the block is a closed C comment.

You are right that _emits_cpp excludes nothing today — every template referencing cpp_license ends in .hpp.mustache or .cpp.mustache. The non-C++ test is therefore written against a synthetic template on purpose: it states the intent for the template that does not exist yet, which is the only way to pin a guard that has no live case.

2. Definition order — accepted, _emits_cpp now sits above render_template.

Verification for this round: codegen suite 202 passed, compass suite 164 passed and 1 skipped, drift gate clean across all nine registry components.

On the build: still compiling locally, last at 56 percent. I will post the result here before merging rather than merge on the strength of the other checks.

mcraveiro:
Build result, as promised before merge: the full build passed. That closes the one limitation recorded in the PR body — the marker is legal in all 3557 changed files.

Run by the author rather than by me; my local run was stopped partway and the passing result is theirs.

PR #2088, opened 2026-09-17
[analytics,agile] Drop the redundant tenant flag and close the analytics bind

mcraveiro:
## Summary

The four analytics entity models carried a profile bind and an explicit has_tenant_id set to true. Both profiles already fix that feature to true, so the explicit line said nothing the profile had not said. Removing it leaves the profile bind alone, which is the shape every other bound component carries. The rest of the change records the survey that found the bind already on main, and reconciles the superseded ad hoc task.

## Changes

- Remove the per-entity has_tenant_id from the four analytics entity models; regeneration moves no generated file
- Record the modeling survey in the bind task: four entities bound, no no-match, two non-entity orgs excluded
- Close the bind task and cross-reference the superseded ad hoc task, which stays DONE

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Entity classification and drift baseline: ores.analytics](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-analytics/story.html) | 9ABCB932-9501-4519-BECA-A30A93646770 |
| Task | [Bind ores.analytics entities to profiles; verify zero-diff regen](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-analytics/task_bind-analytics-entities-to-profiles.html) | 09618E7F-552A-445E-907D-B9D7589E382B |

## Testing

Plan. Regeneration is the test for a binding change. The edit deletes a line that the assigned profile already sets, so the pass condition is that no generated file moves. The drift gate regenerates analytics-cpp in place and compares the result against the checked-in tree. The changed documents are published with the incremental site page build. The diff carries no C++ and no generated file, so there is no new compilation for it to prove.

Evidence. Class: docs. The modeling paths add the codegen drift check. check_component_drift.py --component analytics-cpp reports: No drift, regenerated output matches the checked-in tree, and git status is empty afterwards. compass site page over the seven changed files published 7 pages with 0 failures. The misspell check is clean on all seven changed files. Local code review found no critical or important issue.

Limitations. The full site sweep was not run. It is the only run that reads every file, so it is the only one that finds a link which no longer resolves, and it also regenerates the graph data. This change modifies existing pages and adds no link to a newly created document, so the incremental build covers it. The repository-wide misspell check reports pre-existing hits in files this diff does not touch.


mcraveiro:

PR #2089, opened 2026-09-17
[codegen] Remove model properties the bound profile already fixes

mcraveiro:
## Summary

A profile supplies defaults and the generator merges them, so an explicit property line that restates the profile's own value is dead weight. It invites the model and the profile to drift apart, and it makes the profile stop being the single source of truth. This removes the redundant lines across the components in the drift gate's known-drift-free registry, keeps the genuine overrides, and proves every removal output-neutral by regeneration.

## Changes

- Remove 41 redundant property lines across 19 models in analytics, compute, iam, refdata, reporting and synthetic. The set covers has_uuid_primary_key (19 lines), has_pagination (8), has_tenant_id (7) and has_change_reason_cache (7).
- Keep the two deliberate overrides. compute.result sets has_change_reason_cache false against the profile's true, and reporting.report_definition sets has_workspace_id true against the profile's false.
- Keep the structural properties, because profiles hold variability features only and never assign them. business_unit keeps parent_id_column and hierarchy_name_field, and the other hierarchy models keep their equivalents.
- Keep the four has_parent_id lines. The profile assigns the feature into the sql namespace, but the templates and core.py read it at the entity root, so the explicit line is load-bearing. Removing it dropped the hierarchy function, repository method, handler and service over 24 and 8 generated files. The task records this generator defect as a finding.
- Add the Model and profile hygiene section to the component architecture audit, so a later pass on any component checks for this class of finding.
- Scope the sweep to the known-drift-free registry. dq and dq-cpp carry the same redundancy, but the gate cannot prove a removal there, so the sweep leaves them alone.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Entity classification and drift baseline: redundant explicit properties](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-redundancy/story.html) | CFC41F33-DA7A-4D2A-88AA-70D65150DA6B |
| Task | [Remove explicit model properties the bound profile already fixes](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-redundancy/task_remove-profile-redundant-properties.html) | 88ADF884-86BB-4561-B1A7-6B956010EBAD |

## Testing

Plan. Regeneration is the proof, because a profile supplies defaults and the generator merges them at a specific point in its pipeline. Each affected component is regenerated in place on the committed tree, and the generated output is compared byte for byte with what is checked in. A component whose output moves is a finding, not a pass.

Evidence. Docs class (doc/*, *.org, projects/*/modeling/*). Local site build clean (compass site page, 160 files, 0 failures). Codegen drift gate clean: check_component_drift.py --all over all nine registry entries reports 'No drift: regenerated output matches the checked-in tree', and the working tree stays clean afterwards. Misspell check clean over the changed files (misspell-fixer, nothing to replace).

Limitations. No C++ was rebuilt or retested locally, because the drift gate proves the generated tree is byte-identical and no generated source changed. The whole-tree misspell run still exits 1 on pre-existing hits outside this change, in vendored trees and in earlier agile records that quote upstream identifiers by name; an isolated run over this diff's files is clean.


mcraveiro:

PR #2090, opened 2026-09-17
[codegen] Route has_parent_id to the entity root so profile bindings take effect

mcraveiro:
## Summary

A profile assignment for has_parent_id landed in the entity sql namespace while every generator reader looked at the entity root, so a profile-bound has_parent_id was silently inert. This routes the feature to the root, retires the explicit workaround lines the defect forced, and rebinds the eleven flat entities the defect had masked.

## Changes

- Route has_parent_id to the entity root in _FEATURE_NAMESPACE, where has_tenant_id and has_workspace_id already live.
- Drop the four explicit :has_parent_id: true lines from business_unit, counterparty, party and synthetic.folder. Regeneration still emits their hierarchy function, repository method, service and handler.
- Rebind eleven flat entities from self-referencing-hierarchy to uuid-identified-lookup and drop their redundant :has_pagination:. None has a parent column or a hierarchy name field, so the old binding stayed silent only because the namespace defect stopped the feature reaching the reader.
- Correct the documents the defect had misled: the feature page, the feature index, the variability model, and the two codegen meta-model pages. Both UVL feature diagrams are recompiled.
- Record the defect and its resolution in the component architecture audit, and file the recurrence guard as a capture.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Fix the has_parent_id feature namespace so profile bindings take effect](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/has-parent-id-namespace-fix/story.html) | 689F92F1-39F4-4956-B6CC-F105A3C1B2E9 |
| Task | [Route has_parent_id to the entity root and rebind the entities the defect masked](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/has-parent-id-namespace-fix/task_fix-has-parent-id-namespace.html) | E2929BA0-00ED-4AD6-BB20-AC9CB518B3FB |

## Testing

Plan. Docs, models, and one line of python tooling. The change alters no generated output, so the gate is regeneration: run the drift check over all nine known-drift-free components and require no diff. Then the docs-class site build and the compass suite.

Evidence. docs and modeling and python tooling; codegen drift No drift across all nine registry components, regenerated output matches the checked-in tree; site build 33 pages published, 0 failures; compass suite 164 passed, 1 skipped; both UVL diagrams recompile byte-identical to their committed SVG.

Limitations. No C++ build or ctest run. The drift gate proves the generated tree is byte-identical, so there is no compiled behaviour to test. A loader guard that fails when the feature is on without both column properties is not built here, and is filed as a capture.


mcraveiro:

mcraveiro:
Thanks. Nothing to change, so this round carries no new commit.

Two notes for the record.

The drift check did run before the PR was raised, and it passed. It reported `No drift: regenerated output matches the checked-in tree` across all nine registry components, and that result is in the Testing section of the description. The CI `drift` check also passed on this PR, so the evidence does not rest on my local run alone.

One correction. `has_pagination` is not a feature the two profiles match on. `self-referencing-hierarchy` does not set it at all, and `uuid-identified-lookup` sets it to true. It and `has_parent_id` are the only two axes where the profiles differ. That is why the eleven entities carried an explicit `:has_pagination: true` line that they can now drop. The verdict does not change, because the eleven keep pagination through the new profile and the four hierarchy models never had it.

PR #2091, opened 2026-09-17
[codegen] Express the trade schema and compose its struct from N field groups

mcraveiro:
## Summary

Regenerating the trade entity clobbered 28 of its 30 outputs, which is why the identity collapse could not be made in the model. This closes two of the five gap classes: the SQL schema and the N-way domain composition. Both now regenerate with zero deletions. The service layer and the Qt estate follow in a second PR.

## Changes

- Declares the trade schema in its model: six partial indexes, three soft foreign keys and three code validations, none of which needed new grammar
- Adds a loader for the party_id_from_book_id struct feature, which the template has always supported and no entity had ever used
- Adds Domain groups, the N-way generalisation of the domain_identity_group and domain_audit_group pair, so an entity can compose its struct from as many field groups as it needs
- Resolves each group's field types from its field group model, so anything keyed on a column type still sees one through a member

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Make the trade entity regenerable](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_make-trade-regenerable.html) | FBE43DB4-AC3C-4589-8372-E2803C4C44E8 |

## Testing

Plan. Code and python tooling: full build, ctest against a live environment, the codegen and compass suites, and a database recreated from the regenerated schema to prove the generated SQL is equivalent to what it replaced.

Evidence. Full build clean (linux-clang-debug-make, exit 0, through ores.qt.exe). Codegen suite 202 passed; compass suite 164 passed, 1 skipped. Database recreated from the regenerated schema: all six declared indexes present under their original names, both partial predicates intact, and the insert trigger carries the book-derived portfolio, the book validation, the trade-type validation and the status foreign key. Regression check on bond_instrument confirms the 26 entities on the two-slot path are unaffected. trading_trades_create.sql, trade.hpp and trade_table.cpp each regenerate losing nothing: every line the old regeneration dropped is now generated. The six deletions remaining in trading_trades_create.sql are the template's clock_timestamp() replacing a stale current_timestamp, not lost content.

Limitations. ctest was still running against the live environment when this was raised; the result will be posted on this PR. An earlier ctest run was made against a stopped environment and its failures are not evidence of anything. Three of the five gap classes remain open and are recorded on the task: the service layer, the protocol -- which structurally diverges from the generated shape rather than merely extending it -- and the nine Qt files.


mcraveiro:

mcraveiro:
Test result, closing the limitation recorded in the PR body: ctest is 73 of 73, zero failures, against a live environment (23 services running, database recreated from this branch's regenerated schema).

For the record, since the PR body mentions it: an earlier run showed 8 failures, which I described as environmental before checking whether the environment was up. It was not -- every service including NATS was stopped and the database had no trading schema. The conclusion happened to be right but the check came after the claim, so the only run that counts is this one.

Full evidence now: build clean through ores.qt.exe, ctest 73/73, codegen suite 202, compass suite 164 passed and 1 skipped, and the recreated database carrying all six declared indexes with their partial predicates and the full insert trigger.

mcraveiro:
Round 1 addressed in 09c6c4a03a, pushed as 90197e8154. Three accepted, one declined, and one thing you did not raise that turned out to matter more.

1. Domain includes discarded unconditionally — accepted, and it was live rather than hypothetical. trade.org declared a Domain includes block that the new path dropped in silence. The block is dead now that every field arrives through a group, so it is removed, and regeneration stayed byte-identical afterwards, which is the proof it was dead. The loader now refuses the combination outright instead of documenting the limitation, because a silent drop is worth an error.

2. Both mechanisms declared at once — accepted. Refused with a message naming which supersedes which, so the author is told rather than left to trace mustache conditionals. You were right that identity_group would have been silently ignored.

3. Docstring style — declined. Every function in org_loader.py carries a docstring; the comment skill's rule is to match the surrounding code, and converting this one would make it the odd one out in the module. Happy to revisit if the intent is to migrate the whole file.

4. 'Zero deletions' against a diff with six — accepted, the wording was loose. The six are the template's clock_timestamp() replacing a stale current_timestamp. The PR body now says nothing is lost rather than nothing changes.

5. Not raised, and the important one: a column belonging to no group vanishes from the struct entirely, because the struct is the member list. That costs a field on the wire rather than a compile error, so it is the worst of the three silent losses. Guarded and tested. All 18 of trade's columns are covered, which is why this never showed.

Six tests in test_domain_groups.py pin the guards and the resolution. Codegen suite 208 passed, compass suite 164 passed and 1 skipped. Regeneration of trade is unchanged, so the build and ctest evidence already posted still stands.

PR #2092, opened 2026-09-17
[ores.refdata,ores.marketdata] Move classification into refdata catalogs

mcraveiro:
## Summary

Classification stops being code. The ores.marketdata asset_class and series_subclass C++ enums are deleted, the asset class taxonomy becomes the seven classes the domain analysis fixed, and a new series_subclass_code catalog ships end to end. Consumers carry the code string, so no C++ enum mirrors a vocabulary that a table already holds.

## Changes

- refdata.asset_class_code holds the seven-class taxonomy: rates becomes interest_rates and cross_asset is dropped. The fourteen cross_asset instrument codes move onto their real classes, and the create, drop, populate and RLS policy SQL changes with them.
- A new refdata.series_subclass_code ships end to end: the modeling org, the generated domain type with JSON and table IO, the table, repository, entity, mapper, service, protocol, handler, registrar, history field mapper, an eventing integration test, and the Qt client model, controller, detail dialog and MDI window.
- The ores.marketdata C++ enums asset_class and series_subclass are deleted. The domain types carry the code string, and the two entity orgs, generators, mappers, history field mappers and domain tests follow them.
- The Qt dialogs read classification labels from the catalog tables, showing the catalog name and filtering on the code. The market series grid shows the raw code, because its list response carries no display name.
- Market data validates its classification columns against the catalogs through trigger functions that raise SQLSTATE 23503. market_series validates both columns, feed_bindings validates asset_class, and refdata.instrument_code validates asset_class where it is set.
- The DQ populate scripts and the classification knowledge documents move to the new taxonomy.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Classification catalogs as data](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/classification-catalogs-as-data/story.html) | 623FD704-50F8-4FA8-8EED-7E885797B48C |
| Task | [Move classification into refdata catalogs](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/classification-catalogs-as-data/task_classification-to-refdata-catalogs.html) | 643DE9E2-B105-4463-95DC-29D7907C00A3 |

## Testing

Plan. The change set is code class unioned with docs class, so it runs the code checks plus the site build. For code: the full build and the full ctest suite, the domain roundtrip check, the cmake source-list check, and the codegen drift gate over its known-drift-free registry, which covers ores.refdata and ores.marketdata. For docs: the incremental site publish. The drift gate runs with unrelated local work held aside, because that check diffs the whole tree and cannot pass while other workstreams have dirty files.

Evidence. code + docs. Full build clean, exit code 0, re-run after the review. ctest 100 percent passed, 0 tests failed out of 73, in 1082 seconds; ores.refdata.core.tests passed in 262 seconds and every ores.marketdata suite passed. The domain roundtrip check is clean. The cmake-sources check reports every component_files.cmake up to date. Codegen drift is clean for the registry, and the drift and model CI jobs pass. The incremental site publish rebuilt the 15 pages this change touches and reported 0 failures, after rescanning the org-id index over 5345 source files.

Limitations. Three things are untested or knowingly regressed. No test imports a market data file that carries a CORRELATION series, and such an import now throws std::invalid_argument, because the change deleted that row from the ORE vocabulary table; a correlation throws before the batch is written, so the whole file aborts rather than skipping the row. The example corpus carries CORRELATION/RATE and CORRELATION/CDXIG series, so this is a real regression from main, where the row mapped to cross_asset. It cannot be repaired by retargeting the row, because a pairwise correlation belongs to no single asset class, and market_series.asset_class is not null. The correlation analysis is the next PR. Second, the market series grid shows raw catalog codes where it previously showed compiled-in labels. Third, no test drives the Qt windows, so the label lookups rest on the build and on reading the code. The CDash Experimental submission was not run; the suite result above comes from the local run.


mcraveiro:

mcraveiro:
Both substantive findings are correct and I accept them. Thank you - the CORRELATION analysis is precise, including that the throw happens before the batch is written so the whole file aborts rather than skipping the row.

Blocking, CORRELATION. Verified as described: the vocabulary table lost its row, classify_series_type throws inside find_or_create_series, and the corpus does carry CORRELATION/RATE and CORRELATION/CDXIG. It is not repairable by retargeting the row, because cross_asset was dropped for exactly the reason you give - a pairwise correlation belongs to no single class - and market_series.asset_class is not null. On the maintainer's call this merges with the regression recorded as known debt, and the correlation analysis is the next PR. That work picks between deriving the class from the correlation's operands and making asset_class nullable.

Non-blocking, grid labels. Also correct. The grid returns the raw code and never reads the catalog, and the PR description claimed otherwise; that description is now corrected. The maintainer's direction is that the enum catalogs we added need a human readable description for UI purposes, so the fix is to give the catalogs that description and have the grid display it. Recorded as follow-up rather than fixed here.

Nitpicks. All three are real and recorded: the stale ores.marketdata.api.puml references to the deleted enums, the hardcoded asset class literals, and the loadLookups default that would silently keep index 0 if a code were absent from the catalog. None are fixed in this PR.

Merging with these recorded rather than fixed, per the maintainer's call.

PR #2094, opened 2026-09-17
[reporting,codegen] Bind the reporting entities to profiles

mcraveiro:
## Summary

The ores.reporting catalogue entry joins the codegen drift baseline. Every entity is bound to a Variability Profile or recorded as a genuine no-match, and risk_report_config, the last hand-written entity with no org model, is modelled and generated. reporting joins KNOWN_DRIFT_FREE, so the drift gate covers it from now on.

## Changes

- Bind report_definition to uuid-identified-lookup with its explicit has_workspace_id override, and report_instance to fk-scoped-child with parent_entity_singular: report_definition.
- Model risk_report_config and generate its API domain type, core entity, mapper, repository, and SQL.
- Refuse an entity column the generated struct cannot express: a column matching no entity-template type flag now raises, unless the model declares :sql_only:.
- Move the sql_only guards on the domain class and generator templates into the org sources they are tangled from, so both .mustache files are reproducible.
- Delete the retired non-temporal SQL archetype, drop three false AUTO-GENERATED headers, and regenerate the ER diagram.
- Regenerate the report_instance Qt controller. It is generated only; the build does not include it yet.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Entity classification and drift baseline: ores.reporting](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-reporting/story.html) | 3E49EAF7-0545-4A4C-9DD8-34608FC41293 |
| Task | [Bind ores.reporting entities to profiles; verify zero-diff regen](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/entity-classification-drift-reporting/task_bind-reporting-entities-to-profiles.html) | 789D58B7-5282-41E0-879E-F5052AB295C6 |

## Testing

Plan. Run the union of the change classes' checks locally: docs, code, ci, and python tooling. The branch point classifies doc/*.org and projects/*/modeling/* as docs, the C++ and SQL under projects/ores.reporting, projects/ores.qt, and projects/ores.sql as code, projects/ores.codegen/library/templates/* as ci, and projects/ores.codegen/src and tests as python tooling. CI's codegen-drift workflow runs the same two drift checks on this PR.

Evidence. Full build --preset linux-clang-debug-make: exit 0. ctest: 73 of 73 passed, 0 failures. ctest check_component_drift.py --all: exit 0, reporting No drift: regenerated output matches the checked-in tree. check_model_drift.py --summary: exit 0, 137 of 262 models bound, 2 known exceptions. regenerate_cmake_component_files.py --all --check: exit 0. ore_domain_roundtrip_check.py: exit 0. compass pytest: 164 passed, 1 skipped. misspell-fixer: 3 hits, all pre-existing, none in a file this change touches. build --direct site: exit 0. Environment: the database was recreated twice, the second time because the rebase brought main's retaxonomised asset classes and the new series_subclass_code catalog; ores.refdata.core.tests and ores.marketdata.core.tests both pass now.

Limitations. ores.qt/reporting is regenerated but no CMakeLists.txt builds it, so the Qt output is not compiled or tested. This is deliberate step 1 of a two-step plan; step 2 brings the Qt templates to parity and deletes the hand-written copies. Three reporting entities stay hand-written as recorded exceptions: risk_report_config_books and risk_report_config_portfolios, where the junction archetype emits version plus four audit columns unconditionally, and report_input_bundle, where the domain entity archetype emits valid_from and valid_to into the primary key unconditionally but the table is immutable. The index rename on ores_reporting_risk_report_configs_tbl leaves the old index in place on a database that is upgraded rather than recreated; compass db recreate drops the table with its indexes. The ctest suite ran as index-range chunks rather than one run, because three other worktrees on this host tripped the memory guard on long runs; the chunks cover tests 1 to 73 with no gap and no repeat, and every chunk reported 100 percent tests passed.


mcraveiro:

mcraveiro:
Both findings handled.

1. Sentinel round-trip - not changed. Confirmed pre-existing and repo-wide: this change touches neither cpp_domain_type_mapper.cpp.mustache nor any caller of it, and every nullable numeric column in the tree shares the convention. Your reading of the trigger is right: a caller that sets the field to exactly 0 persists NULL. Recorded as finding 8 in the task Review table, declined, with its own change named as the fix.

2. var_quantiles - fixed in 7db8f17b87. Nothing in C++ reads the field, and the task records why: sqlgen cannot bind numeric[]. But your point about checking downstream was the right one - ores.reporting.api.puml is hand-maintained, not generated. Its CMakeLists only renders the png, so the diagram still declared the member. It was the one divergence across its 36 members. The member is removed and the png regenerated with the same command the cmake target uses; the image keeps its width and loses one text row. Recorded as finding 9, accepted.

mcraveiro:
Correction to the sha in my reply above: the branch was rebased onto main to clear a merge conflict, so the var_quantiles fix is now fee9c8999b rather than the sha I quoted. All four CI checks are green on that head.

The conflict is worth noting: it was in projects/ores.lisp/src/ores-build-skills.el, where main had independently fixed the same stale-skills defect with ores/prune-stale-skills. That mechanism is better than the one this branch had -- it prunes after a successful publish, where this branch emptied the directory before it -- so the resolution keeps main and drops the local purge. What this branch still adds is the exit code: the publish script messaged on failure and then fell through, exiting 0. Both paths now set the exit code.

PR #2095, opened 2026-09-18
[marketdata] Extract the series classifier and give a series its asset classes

mcraveiro:
## Summary

Lift the ORE-vocabulary mapping out of import_service.cpp into a standalone, testable classifier, complete the vocabulary so every series type the ORE corpus carries classifies, and replace the single market_series.asset_class column with a junction table so a pairwise correlation can name both of its classes.

## Changes

- Move the ORE-vocabulary table out of import_service.cpp into ores.marketdata.core/classification/series_classifier.{hpp,cpp}. The signature is the market_series natural key, series_type, metric and qualifier, so a test reaches it with no database, NATS or import pipeline.
- Complete the vocabulary. The eight series types the corpus carries that the old table threw on now classify, so an import file that carries one no longer aborts before a batch is written. A pairwise correlation takes its classes from the two operands in its qualifier, by the factor-class prefix ORE puts on them, so a cross-class correlation yields two classes.
- Replace market_series.asset_class with ores_marketdata_market_series_asset_classes_tbl, a tenant-scoped junction generated from a new ores.codegen.junction org. Dropping the column took its check row, its generator and its insert-trigger validation with it. Junction insert-trigger validations are new, and org_loader now parses that section for a junction as it already did for a domain entity.
- Cover the corpus rather than a hand-written file list. A test walks external/ore/examples, pushes every distinct series key through try_classify and asserts the failure list is empty, with anti-vacuity floors underneath. The subclass catalog grew from 15 codes to 18 and three descriptions were amended to match the codes they describe.
- Take the consumer port with it. import_service writes one junction row per code the classifier returns, and feed_ingest_loop, the DQ publish SQL, the domain tests and six Qt files follow.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Extract the series classifier and cover the whole ORE corpus](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/complete-market-series-classification/task_extract-series-classifier-and-cover-the-corpus.html) | BF395A0B-C738-43B7-B4FA-DB32988DCAE8 |

## Testing

Plan. code and docs and codegen templates, so the union of the code and docs classes. Full build and ctest, the corpus classifier suite, the roundtrip check, the cmake-sources and codegen drift checks, the compass python suite, the site build, and a database recreate for the schema change.

Evidence. Build green at 100% with no errors and no warnings on linux-clang-debug-make. ctest 73 of 73 passed in 1126s, with ores.marketdata.core.tests at 697 assertions across 191 cases and all four corpus cases among them. compass db recreate -y -k clean in 1m25s with both the series and junction rows present. regenerate_cmake_component_files.py --all --check reports all component_files.cmake up to date. ore_domain_roundtrip_check.py exits 0 and its two input trees are untouched by this change set. Compass python suite 164 passed with 1 skipped. Site build exit 0. Misspell reports three hits, all the upstream identifier LOAD_PROFILE_1_EXPLICT in records outside this change set, already declined in two earlier review rounds.

Limitations. check_component_drift.py --all exits 1 in this worktree, and every file it flags is an uncommitted edit belonging to another workstream, namely the two agile stories, doc/llm/memory, the marketdata modeling subdirectory wiring, and base64_converter.cpp. No file in this commit is among them, and marketdata is in the check registry and regenerated clean apart from that excluded line. No automated test reaches the Qt windows, so the RateCurvesMdiWindow filter change is verified by compile, link and inspection only. The corpus walk covers classification but not the ORE-to-oresmd round trip, per the scope ruling. The FIXING rule is unexercised because the corpus carries no FIXING series key.


mcraveiro:

mcraveiro:

PR #2096, opened 2026-09-18
[codegen] Generate the trade service layer

mcraveiro:
## Summary

Regenerating trade clobbered its handler, service, repository, mapper and protocol, which is why the identity collapse could not be made in the model. They now come from the model. Closes the remaining gap classes from the previous PR except the Qt estate, which follows.

## Changes

- Adopts the generated protocol shapes, extending only where the entity had a reason: a vector-valued save, a node-scoped list that paginates, and the handler's http base url
- Moves the bespoke handler, repository and protocol bodies into the model through the paste points the templates already had
- Replaces the hardcoded identity. member path with the prefix each column carries, across six templates and four places in core.py, collapsing the duplicated branches it had required
- Widens ores.history's history_version_source concept to admit a grouped audit, the one place the flat-struct assumption had escaped codegen
- Adds the trading-record profile and binds trade to it: tenant and workspace scoping for transactional trading data, leaving party derivation and struct composition to the model so an instrument can adopt it unchanged
- Makes the seeded FSM faithful to the trade lifecycle note and moves the transition guard into the insert trigger, where the version select already holds the row lock
- Lets the eventing seeder reach a parent in another component: it always resolved them, but the includes and namespaces assumed the child's component

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Regenerate the trade service layer and Qt estate](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_regenerate-trade-service-layer.html) | 577C2F64-3394-44A0-9AD9-FBD61143B8CA |

## Testing

Plan. Code class: full build, ctest against a live environment with all services up and a database rebuilt from the committed scripts, the codegen drift gate over the registry, and the two python suites.

Evidence. Full build clean (linux-clang-debug-make, exit 0). ctest 73 of 73, exit 0, against 23 running services and a database recreated from the committed schema and seeds. Codegen drift gate clean across all nine registry components, including the shared eventing template change, whose only effect elsewhere was to qualify a currency generator call in two refdata tests. Codegen suite 208; compass suite 164 passed, 1 skipped. FSM verified in the database: eight transitions matching the lifecycle diagram, and the insert trigger carrying the guard.

Limitations. trade's own eventing test is generated but switched off in the profile, which records why: the seeder now reaches refdata for the book and portfolio, but the seeded book's party is not active when the book is written. No instrument binds the new profile yet. The trade history provider is composed but unexercised until the Qt history dialog drives it. trading remains outside the drift registry.


mcraveiro:

mcraveiro:
All five blocking findings confirmed. I verified each against the tree rather than taking them on faith, and every one reproduces:

1. stamp() no-ops. handler_helpers.hpp:119 gates on 'requires { obj.tenant_id; }', a direct-member check, and the generated call is stamp(t, ctx_) on the split struct. origin/main had stamp(t.identity, ctx_) and stamp(t.audit, ctx_). Confirmed.
2. execution_timestamp: 2 occurrences on origin/main, 0 now. Silently dropped in both directions. Confirmed.
3. Generator emits activity_type_code 'New' and a random status_id; the seed only defines codes like new_booking. Confirmed, and the model carries the literal.
4. product_type::unknown mapped to nullopt on origin/main and now always renders a string the Postgres enum does not have. Confirmed.
5. Four seeded transitions -- capture, execute, unexpire, uncancel -- are referenced by no activity type. I queried the database: draft is unreachable and neither ending is reversible, contradicting the seed's own comments. Confirmed.

Your closing question is the important one, and the answer is worse than 'coverage is read-only'. Nothing writes a trade at all: no caller of save_trade or save_trades in any test, and the files touching trade_repository belong to trade_type and trade_id_type. Trade writes have no coverage, which is why 73 of 73 is green.

And the eventing test I generated would have been the first real trade write. It failed, and I switched it off. Finding 3 is very likely part of what it was reporting -- an invalid activity code and a random status_id fail validation before the seeding chain matters -- so I read one error message, attributed it to the seeded book's party, and disabled the test instead of reading it properly. That was the wrong call: the test was the only thing telling me the truth, and turning it off is what let items 1 to 4 through.

Working the fixes now, starting with stamp, since that one defeats the tenant boundary and makes the audit stamps client-spoofable. I will not re-run the suite as evidence of correctness for these; the fix has to be proven by a test that actually writes a trade.

mcraveiro:
Fixed in c5b9446fb8. All five blocking findings and four of the five smaller items accepted; one declined with a reason below.

Blocking. (1) The service now stamps each domain group. While fixing it I found the batch loop had no braces, so even the corrected stamp would have covered only the first group. (2) The mapper carries execution_timestamp both ways. Root cause was narrower than a missing fallback: there was no branch at all for an optional string domain field against an optional timestamp column, so the field fell through silently. Added the conversion, str() on the way out and emplace on the way in. I kept the domain type a string rather than making it a time point, because the group's other temporal fields are strings and changing it would have pulled four hand-written ores.qt files into a PR that deliberately leaves Qt alone. (3) The generator emits new_booking and leaves status_id nil. (4) A default-constructed enum maps to nullopt. (5) Added draft_capture, execution, maturity_reversal and cancellation_reversal; I queried the machine afterwards and all eight transitions are now named by at least one activity.

Smaller items. FSM resolution is extracted into ores_trading_resolve_trade_transition_fn, so the two branches keep only their differing guards. Seed items carry parent_component_include and parent_component_core of their own, and the stale docstring is corrected. trade_status_service and fetch_fsm_transitions are deleted. The end-of-line comments moved above their members. I declined the redundant row read: the row is already locked by the version select, so it costs one index probe, and removing it needs the shared version-select template to accept entity-specific columns, which is more grammar than the saving earns.

You asked whether the tests exercise a real write. They did not, and fixing that is what found the rest. I re-enabled the trade eventing test rather than leaving it off, and it exposed five further defects that no existing test could have caught:

- status_id was validated before the trigger resolved it, so every trade write failed before reaching the transition logic meant to set it. Same ordering problem your item 3 describes, and it would have defeated the item 3 fix on its own.
- trade_identity_field_group.org never declared workspace_id even though the committed header carried it. The header was ahead of its model, so regenerating the group dropped the field.
- The eventing seeder gave a party only to a direct parent. An ancestor one step deeper, the portfolio a book hangs off, was written with a generated party_id and rejected. That is your cross-component seeding item, one level further down.
- portfolio_id was cross-checked against the book's parent rather than derived from it. The book settles both party and portfolio, so the trigger now derives both and ignores what was sent.
- The generated round-trip test amended with the booking activity, which names a machine-starting transition and is correctly illegal on a row that already has a state. The model now declares the activity an amendment carries.

Evidence. ores.trading.core.tests passes, and the full suite is 73 of 73 against a recreated database with services up. I also checked the written rows directly rather than trusting a green run: two versions, new_booking resolving to live and then amendment, modified_by stamped, execution_timestamp persisted, and both party_id and portfolio_id equal to the book's. Codegen drift is clean across all nine registered components and the codegen suite is 208 of 208.

One gap I am not claiming coverage for: item 4 is correct in the generated mapper but is not exercised, because the trade generator always emits a real product_type, so the unknown-to-null path never runs in the test.

mcraveiro:

mcraveiro:
Follow-up on evidence. The full suite is now 73 of 73 against a database recreated from the rebased tree, with all 23 services up. My earlier reply reported 73 of 73 from before the rebase; this is the same result re-established on the merged tree, so the number is no longer inherited from a stale branch.

Two things were in the way and are worth recording.

CI's drift check failed after my first push, and it was right. main had added the series_subclass_code entity to refdata after I branched, and CI checks the merged tree. My local drift run was clean only because that entity did not exist in my worktree, so the braced batch-stamp loop never reached it. A repo-wide codegen change cannot be proven drift-free against a stale branch; rebasing first was the missing step. Fixed in its own commit, and the gate is clean on the rebased tree.

Recreating the database then failed on main itself, not on this branch: 'OreStudio Code Generation Methodology' was defined in solvaris_methodology_populate.sql and was deleted with the slovaris dataset, but acme_dataset_populate.sql and ten refdata and synthetic seed scripts still name it. That is now fixed and merged separately as #2099, so this branch is rebased on a main that can build its own database again.

PR #2100, opened 2026-09-18
[codegen,sql] Add the image and ip2country dataset archetypes

mcraveiro:
## Summary

Completes the ores.sql.populate archetype set, so crypto, flags, ip2country and FPML can leave the legacy bash generators.

The nine generic archetypes cover the six-file metadata core of a dataset. They do not cover the artefacts. Crypto icons and country flags need an image archetype. IP geolocation needs an artefact plus a TSV import. FPML needs 22 per-entity datasets with an artefact each, and that is the next task.

The branch also fixes a defect in the generator itself. Mustache escapes values for HTML by default, and render_template took that default, so every value a template emitted reached the generated file as an HTML entity. No committed file shows the damage, because no committed payload carries an apostrophe or an ampersand. The image archetype would have shown it first: images_generate_sql.py, which it replaces, escaped quotes by hand.

The review round closed the rest of that gap. Stopping HTML escaping is not the same as making a value safe in SQL, so both artefact templates now double the apostrophe in every value they quote, and both builders select the manifest source their dataset names rather than the first one that carries the key they need.

## Changes

- Retire the flag archetype. It took its rows from manifest.json, which made a metadata document double as an artefact payload. The image_artefact archetype replaces it.
- Add the image_artefact archetype and its mustache. They emit the artefact rows for crypto icons and country flags, and read their entries from images.json.
- Add the ip2country archetype and its mustache. They insert the artefact rows and load the ranges from a TSV file at run time.
- Let an archetype's data_source field name several payloads. The generator splits the field on whitespace, validates every path, and passes the extra files as additional model paths.
- Stop mustache escaping values into generated files. render_template now builds a Renderer with a no-op escape, because pystache's module-level render ignores its escape argument and silently keeps HTML escaping on.
- Add a test that pins the escaping rule. A description carrying an apostrophe and an ampersand must reach the output verbatim.
- Move the ISO generator from src to external/iso, beside the dataset it generates, and fix its stale output. It emitted a schema switch and metadata calls that no committed populate file uses.
- Add the provenance fields the four manifests were missing: origin_code, nature_code and treatment_code, plus as_of_date on FPML and methodology_file on ISO.
- Escape the values the two artefact templates quote. The template writes the surrounding quotes, so an apostrophe in a description or a dataset name would close the literal early and the generated SQL would not parse. Both builders now double it.
- Select the manifest source a dataset names as its methodology, not the first source carrying the key the builder needs. external/crypto lists two sources with a data_dir, so only their order kept the choice right. A dataset that names no source now raises instead of guessing.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Migrate remaining ~100 entity models from JSON to org](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/migrate-remaining-entity-models-from-json-to-org/story.html) | 5FFB587D-D85A-4B51-8C80-0F8D485754A7 |
| Task | [Add the dataset archetypes the four-dataset migration needs](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/migrate-remaining-entity-models-from-json-to-org/task_add_dataset_archetypes.html) | 9B900D8C-8F98-4733-96AF-2C838370ED9C |

## Testing

Plan. Ran the class checks a mixed codegen and SQL change calls for, and the full ctest suite to see whether anything outside codegen moved:
- the codegen unit suite
- the compass unit suite
- template tangle drift
- the cmake source-list regeneration check
- the component drift check across all components
- the ISO generator round trip, which must leave the tree clean
- the site build and the full build
- misspell across the tree
- the full ctest suite

Evidence. Every check passes except the items listed under Limitations.
- Codegen unit suite: 239 passed, 1 failed. The failure is test_a_colliding_earlier_column_does_not_hide_a_later_real_one in test_custom_type_headers.py. It reproduces on origin/main, at 8036dc81fd, in a clean detached worktree with nothing from this branch applied. Four of the passing tests are new in the review round and cover the quote escaping and the source selection.
- Compass unit suite: 166 passed, 1 skipped.
- Tangle drift: none. Regenerating every registered component produces byte-identical output, so the escaping fix changed no committed file and nothing committed depended on the escaping. Re-run after the review round changed the two artefact templates, still no drift.
- cmake source lists: all up to date.
- Codegen component drift across all components: no drift.
- ISO generator: reruns clean and leaves git status empty, so the committed ISO output is reproducible from its manifest and its methodology.
- Site build: succeeded, exit 0.
- Full build in linux-clang-debug-make: reached 100%, exit 0.
- misspell: five EXPLICT hits, all in doc/agile/versions/v0/sprint_25, all present on origin/main, none in a file this branch touches.
- ctest: 73 suites, 73 passed, 0 failed, in 1105 seconds. The three suites that failed on the stale schema, ores.refdata.core.tests, ores.trading.core.tests and ores.marketdata.core.tests, all pass on the recreated database.

Limitations. The test database in this worktree was 6 days stale, built from b772be2e43-d, which predates the commits that add the trade envelopes table and the calendar column on bond issues. The first ctest run therefore failed three core suites for reasons unrelated to this branch, which touches no schema file. The database was then dropped and recreated from this revision, and the run above is the one on the recreated database. The branch was later rebased onto main, 8036dc81fd, to clear a merge conflict in core.py, so that run predates the rebase. Main has touched modeling and populate files since, so this database is stale against the rebased branch and the ctest evidence above describes the pre-rebase tree.

The new archetypes are unexercised. No payload JSON exists for crypto or flags yet, so nothing consumes them and no generated output was diffed against the committed SQL. Two acceptance criteria on the task stay unverified: that the image archetype reproduces the images_generate_sql.py output, and that the ip2country archetype reproduces its committed output. The codegen tests cover the loader and the tangle, not the SQL that the templates emit.

FPML is not started. Its per-entity table spec is the next task.

The ISO manifest contradicts its own methodology. The manifest still declares source_system WIKIPEDIA and license CC BY-SA 3.0, and names two Wikipedia extraction methodologies. The new methodology.txt says the country and currency rows were entered by hand from the standards, with no source file. The generated iso_methodology_populate.sql therefore carries both claims at once, and a reader cannot tell which is true. This was raised with the user and left open, because the answer decides what the licence may assert.


mcraveiro:

mcraveiro:
Round handled. The review names three findings. Two are accepted, one is declined.

**Finding 1, SQL quoting in the artefact templates.** Accepted, fixed in ecda0b3952. Both builders now double the apostrophe before the template sees the value, and both templates gained a regression test. I probed psql before choosing the escape. `\set f 'a''b'` yields `a'b`, so the `\set` default in the ip2country template takes the same escape as the SQL literals. The same gap is present in `sql_dataset_populate`, `sql_catalog_populate` and `sql_methodology_populate`. It is pre-existing, and none of those three archetypes runs on this branch, so the fix stays out of this PR.

**Finding 2, source selection.** Accepted, fixed in ecda0b3952. Both builders now select the source whose `name` matches the dataset's `methodology`, and a dataset that names no source raises instead of guessing. Every manifest in the tree sets that link. The crypto and ip2country test fixtures were missing it, so they gained it.

**Finding 3, `model_key` shadowing.** Declined. This repeats finding 2 of the previous round. The first key of `payload_paths` is the model itself, and the outer binding has no reader after the loop, so a rename would be churn with no behaviour change.

The ISO licence contradiction is still open and needs a decision. It is recorded in the PR description and in the task notes.

mcraveiro:
The rebase onto main rewrote the commit SHAs this round's fixes landed in, so the reply above names a commit that is no longer in main. The changes themselves are unchanged and merged. Their new homes are:

- `c353fceca9` — the quote escaping in both artefact templates, and the source selection that follows the dataset's methodology link.
- `acdcdb7048` — the task doc's review round and the SQL-quoting fix.

The rebase resolved one conflict, in `core.py`, where main's TypeScript support and this branch's escaping fix both added a block after `_emits_cpp`. Both are kept.

PR #2101, opened 2026-09-18
[codegen] Let the entity guard see the nullable enum the template carries

mcraveiro:
## Summary

Trade cannot be regenerated on main. The struct-coverage guard added in dfa62d1a87 refuses a nullable scoped enum, on the premise that no type flag claims one and the column is therefore dropped from the entity struct. The entity template does claim it, and trade.product_type is exactly that shape: nullable so an unset product type reaches the column as NULL, because the Postgres enum has no member to spell it. This teaches the guard the flag it was missing, and abandons the Qt estate task.

## Changes

- Add render_is_enum to _ENTITY_STRUCT_FLAGS: the template pairs it with an is_enum inversion to emit an optional string, which is the member trade_entity.hpp has carried all along
- Rewrite the guard's test, which asserted the same wrong premise; it now generates the column and checks the optional member is present
- Abandon the trade Qt estate task, recording what the investigation found: bond_instrument is five foreign keys with no economics, and the containment that would let codegen follow bond_issue, bond_leg and bond_leg_rate is not modelled

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Regenerate the trade Qt estate](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/data-oriented-trading-model/task_regenerate-trade-qt-estate.html) | 5A01EEBC-3955-4B1D-8490-698F1DCF824E |

## Testing

Plan. Prove the refused shape does reach the struct, confirm trade regenerates again, and check no other component moves.

Evidence. docs and python tooling; codegen suite 211 passed, drift clean across all nine registered components, compass suite 166 passed 1 skipped, site page on the two changed org pages 2 files 0 failures. Generated the guard's own probe column and read the struct back: the optional string member is present. compass codegen regenerate for trade at address ores now completes.

Limitations. test_custom_type_headers.py::test_a_colliding_earlier_column_does_not_hide_a_later_real_one fails, and fails identically with these changes stashed, so it is pre-existing on main and not addressed here. trading is excluded from the drift gate, which is why this reached main unnoticed; that exclusion is unchanged.


mcraveiro:

mcraveiro:
Fixed in 98463b86dc. Both nits accepted.

1. The module docstring said nine type flags; the tuple holds ten. Counted it rather than trusting either number, and corrected the docstring.

2. The Status table's State cell. This one was my miss, and your diagnosis is exactly right: the abandon commit matched on STARTED, which that cell never held, so it rewrote the narrative fields and left State reading BACKLOG. Now ABANDONED, consistent with Now, Next and the story row.

Thank you for tracing the template rather than taking the body's word for it. Your reading matches what I found, including the point that the guard still refuses what it exists to catch: an array column falls through is_nullable_numeric and is_simple and is still rejected, so this widens the guard by exactly the one shape the template handles and no more.

On your 'not verified' note, the suites do run here: codegen 211 passed, compass 166 passed 1 skipped, drift clean across the nine registered components, and site page on the changed org pages clean. The one codegen failure, test_custom_type_headers.py::test_a_colliding_earlier_column_does_not_hide_a_later_real_one, is pre-existing on main -- it fails identically with these changes stashed -- and is called out in the PR's limitations rather than fixed here.

PR #2102, opened 2026-09-18
[ores.ore,ores.marketdata] Make the series vocabulary data

mcraveiro:
## Summary

The ORE series vocabulary becomes data. How a series key splits into a series and a point, and what asset class it carries, are now rows in managed tables rather than lists compiled into the binary, so a series type ORE adds later, or one a user brings, needs an inserted row rather than a rebuild. Both tables are seeded from the corpus, copied into every new tenant, and covered by a test that walks the examples rather than listing them by hand.

## Changes

- [ores.ore] How a key splits is a table. ores_ore_series_key_shapes_tbl carries the thirty-five rows the compiled table held plus the eight the corpus carries and it did not know, and series_key_registry becomes a value built from those rows and passed in by const reference.
- [ores.marketdata] What a key classifies as is a table. ores_marketdata_series_classification_rules_tbl reproduces the compiled map row for row, so the swap changes no answer the classifier gave before. Two columns carry what a single class column could not: asset_class_source separates a literal rule from a correlation whose classes are read from the two operands its key names, and the key holds the metric as well as the type so GENERIC-MD is classified by the instrument inside it.
- Both readers take their table by const reference and never reach for the database, so each is testable with no database, no NATS and no pipeline. An empty table throws and names itself; a type with no row folds its whole key into the qualifier, loses nothing, and still reconstructs verbatim, so an uncatalogued type never takes an import down.
- Both tables enable row level security through the :rls_tenant_isolation: flag, and the tenant provisioner copies each seed into a fresh environment so a new tenant can import. The review round found the two tables lacking the policy their siblings carry and fixed it before the branch left.
- Tests walk the corpus through both readers and assert the classifier's type set equals the table's in both directions. The eight added rows drop the census of distinct series from 1,612 to 1,117.
- One commit rides along from other sessions, parked verbatim with no edits of mine because the rebase refused a dirty tree. Its one code change guards the base64 converter's OpenSSL calls against a length an int cannot represent, and drops a BIO_free_all that ran before a throw while the catch block freed the same pointer. The rest is bookkeeping belonging to other stories: the oresmd native market data identity story and its two task records, the entity classification drift marketdata story and its task, one memory record, three product backlog captures, and the market data model design task. Two components also gain the ER diagram artefacts their siblings already track, being a CMakeLists.txt, a puml source and a png render under projects/ores.ore/modeling and projects/ores.marketdata/modeling.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Register the unmapped series types in the key registry](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/complete-market-series-classification/task_register-the-unmapped-series-types.html) | 65572B8E-55B8-4988-AE63-A25EBAFD65CD |

## Testing

Plan. docs plus C++ plus codegen templates, so the union of the three classes. Full build and the suites the change reaches; the ORE round trip; the cmake-sources and codegen drift checks; and the site build, which is the only check that renders the pages.

Evidence. code, docs and ci. Full build clean. ores.marketdata.core.tests passed in 31.43 s and ores.ore.core.tests in 535.41 s, both re-run after the rebase, both reporting no failure and no skip; the ore run matters most because a test tenant is provisioned fresh on every run, so it proves the provisioning copy fills both tables with the policy switched on. The row level security fix is verified at the database as well as in generated SQL: a catalogue query reports both new tables with row level security on and one policy each, matching ores_marketdata_market_series_tbl. The cmake source lists report All component_files.cmake up to date; the template tangle is clean; misspell-fixer exits 1 on four hits, all pre-existing and none in this diff; the ORE round trip exits 0; compass lint is clean; the site build over the rebased tree reports Build succeeded with 5458 nodes and 32667 links. The codegen drift check reaches a clean verdict: it first flagged one file, series_classification_rule_service.cpp, which the rebase left stale against a template commit that landed on main after this branch was cut, and regenerating it as bdfdf164d3 brought the check to No drift.

Limitations. Two. The full ctest sweep was not run: the instruction was to re-run only the suites that had failed rather than the whole set, so the suites this change cannot reach were not exercised locally. The site build is long and memory-hungry, and the harness low-memory guard has killed it on this machine before, a known intermittent defect with its own backlog capture; the run over this tree reported success, but a fresh worktree can need more than one attempt.


mcraveiro:

mcraveiro:
Not changed for either observation. Both are now recorded in the task Review table.

First, the wildcard row. The seed settles the shadowing question: 43 rows cover 43 distinct series types, one row per type. Forty-two carry the empty metric and exactly one, the EQUITY_OPTION row for GENERIC-MD, carries a metric, so no type holds both a metric-specific row and a wildcard. One correction to the mechanism as described: the constructor does not reject a duplicate (series_type, metric) pair. It assigns into its map, so a duplicate would silently overwrite the row before it. That case is unreachable rather than rejected, because the read filters on valid_to being the maximum, which leaves exactly one live version per natural key.

Second, the guard stays. It keeps on_tick from depending on an invariant that start() establishes, which you allowed as deliberate defence. Removing it would trade one std::optional test for a cross-function assumption about call order. Two corrections to the reference, neither changing the substance: the file is projects/ores.marketdata/service/src/app/feed_ingest_loop.cpp rather than the core/src/service path, and the guard is at line 409.

PR #2105, opened 2026-09-18
[codegen] Repurpose the presentation drawer into an ores.ts.ui facet

mcraveiro:
## Summary

The TypeScript screens need the metadata the Qt templates read, so the entity models keep their drawer under the name Presentation, and a new ores.ts.ui facet projects it into one TypeScript module per entity. The Qt templates and the facet that carried them are gone.

## Changes

- Rename the Qt drawer to Presentation in all 109 entity models, with the Columns and Icon columns sub-tables, and move the twelve variability features from the qt namespace to presentation.
- Delete the ores.cpp.qt facet, its nine archetype orgs, the nine tangled .mustache files, and the Qt scaffolding in generate.py.
- Add the ores.ts.ui facet and its ui_meta archetype, emitting projects/ores.typescript/src/{component}/ui/{entity}_ui.ts.
- Add core.ui_meta_projection, which derives every member from the drawer or the entity tables, and omits a member the model does not carry. _ui_lookup resolves the collection one hop; _ui_column_style implements the mapping table from the request.
- Withhold the facet from a domain entity or junction whose drawer carries no column table, in generate.resolve_targets, before the per-archetype enabled overrides.
- Add the hand-written projects/ores.typescript/src/ui-contract.ts, the only hand-written TypeScript in the change.
- Emit 90 modules, one per entity that carries a column table: refdata 63, synthetic 7, compute 6, iam 5, analytics 4, reporting 4, marketdata 1.
- Add projects/ores.codegen/tests/test_ui_meta_projection.py, and clean up the Qt references the facet removal left in the docs.
- The parent story stays open: its acceptance asks for the ores.qt and ores.wt.service trees to be gone, which is PR 2.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Retire Qt and emit the UI metadata to TypeScript](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/retire_qt_and_emit_ui_metadata_to_typescript/story.html) | 6C5EE3F6-0DE6-48C2-9325-21D818C7FD6B |
| Task | [Repurpose the presentation drawer into an ores.ts.ui facet](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/retire_qt_and_emit_ui_metadata_to_typescript/task_repurpose-the-presentation-drawer-into-an-ores-ts-ui-facet.html) | 0DC6D0CB-D420-49B6-8751-43598610FEA4 |
| Environment | festive_dijkstra | |

## Testing

Plan. Regenerate all nine registered components and compare the output byte for byte against the checked-in tree; compile the emitted TypeScript against the contract with tsc --noEmit; build the system and run the full test suite; run the C++ and SQL roundtrip; build the documentation site; run the codegen and compass suites.

Evidence. Classes: docs (projects/*/modeling/*, *.org, doc/*), ci (library/templates/*), python tooling (projects/ores.codegen/*), and the catch-all for the emitted TypeScript. Full build clean on linux-clang-debug-make; ctest 73 of 73 passed; roundtrip clean; cmake-sources drift --all --check reports every component_files.cmake up to date; codegen drift check_component_drift.py --all reports no drift and no untracked file; model drift reports no model drift, 2 known exceptions; compass suite 166 passed, 1 skipped; codegen suite 253 passed; npx tsc --noEmit clean under strict and noUncheckedIndexedAccess; site build succeeded. The change touches no C++ source, so the build and the suite run against a compiled tree this branch does not alter.

Limitations. The ores.qt and ores.wt.service trees stay: PR 2 removes them, so nothing here proves the build works without Qt. The emitted keys are not translated; the English, Portuguese and French catalogues live in Volga. Ten Qt: prose subsections and the codegen pages describing the retired facet still name Qt; PR 2 removes them with the trees. misspell-fixer . exits 1 on its prefilter list, all ten matches outside the change. The CDash experimental upload was not run; it is visibility only.


mcraveiro:

mcraveiro:
All three accepted.

1. Fixed in d2c2beb151. The pointer is gone from all four models. There is a fourth file the finding did not name: projects/ores.dq/modeling/ores.dq.synthetic_fx_spot_config.org, same deleted section, same dangling reference. The rationale each sentence gives stands without the table.

2. Fixed in ce89d33e6c. Two cases in test_ui_meta_projection.py drive generate_from_model on an entity with no detail-field table, one keyed on name and one on code. No existing case reached the auto-default block, since _project hands the projection its fields. Both cases fail when key_is_name is forced false, so the guard bites.

3. Fixed in 39a12f245c.

Checks for this round: codegen suite 255 passed, codegen drift no drift, model drift no model drift with the 2 known exceptions, compass suite 166 passed and 1 skipped, site pages 7 published with 0 failures.

PR #2108, opened 2026-09-19
[web] Integrate the Volga TypeScript UI as ores.web

mcraveiro:
## Summary

Adopt the Volga TypeScript prototype as the projects/ores.web component, generate its TypeScript from ores.codegen instead of copying it by hand, retire projects/ores.typescript, and launch the interface from compass services start on a port inside the environment block.

## Changes

- Add projects/ores.web with its four npm workspace packages, renamed to the @ores scope and the ORES_WEB_*/ORES_NATS_* variables.
- Retarget the ores.ts.ui and ores.ts.protocol facets at the component, and delete projects/ores.typescript and the hand-copy script.
- Move the hand-written ui-contract.ts out of the generated tree into packages/web/src.
- Register ores.web.service with the new :runtime: node and :entry_point: model properties, and render a Type=simple Node unit in systemd_generate.
- Add ORES_WEB_PORT (offset 2, the retired Wt slot) and ORES_WEB_SESSION_SECRET to the environment, and generate an mTLS certificate for the service.
- Build the component from CMake, and record the TypeScript technical space in the modeling docs.
- Ride-along fixes found while verifying: the Wt role assignment that made `compass db recreate` fail on main, and the removal of the remaining dead Wt database, environment and service residue (`.env` format 22).

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Integrate the Volga TypeScript UI as ores.web](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/integrate-volga-web-ui/story.html) | BAAFD7B2-A37F-4ED8-A9F2-E2A997E3D488 |
| Task | [Adopt the Volga prototype as projects/ores.web](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/integrate-volga-web-ui/task_adopt-volga-as-ores-web.html) | 57323233-C2BA-459A-BC19-94D4FDA0849A |

## Testing

Plan. Unit-test the new Node unit renderer, the service-registry properties, and the broker and environment resolution in the BFF. Typecheck, test and build the component. Run the codegen drift gate and the documentation validator. Smoke-test the built BFF against the running environment.

Evidence. A full `compass db recreate` followed by Acme Corporation provisioning was run to verify the SQL cleanup; the fleet reports 24/24 running and the interface login returns HTTP 200 with the party picker populated. npm run typecheck, npm test and npm run build pass for projects/ores.web, 73 tests across the four packages. check_component_drift.py --all reports no drift in the generated tree. validate_docs.sh reports that all 42 components pass. The built BFF listens on 21402, serves the bundle and the API, logs 'Service ready', selects eager_maxwell from ORES_ENV_NAME, returns a JSON 404 under /api, and answers a bad credential with 401 through IAM over mTLS. The systemd renderer and service-registry loader tests pass.

Limitations. compass services start deploys the units and starts the fleet through --use-busctl, which reaches the systemd user manager over the D-Bus session bus. Plain systemctl cannot connect from inside the bubblewrap sandbox: the sandbox's user namespace leaves the peer credentials unrepresentable (SO_PEERCRED reports pid 0), so the manager closes the private socket. A full services start was not run to completion because the fleet was already running, so the readiness wait had no new 'Service ready' lines to find; the unit deployment, the start call and the transport are verified separately. The browser verification was not run: it needs a Playwright browser download and a seeded account. The committed generated TypeScript is limited to the components the drift gate covers, so regeneration runs per KNOWN_DRIFT_FREE slug, not with --all.


mcraveiro:
@claude review this

mcraveiro:
All four findings accepted and fixed.

1. ORES_WEB_SESSION_SECRET — removed in b53f7a686f. The variable, the config field, and every comment claiming a signing behaviour are gone. You were right that signing adds nothing here: the cookie carries a 32-byte random identifier and the session store keys on its SHA-256, so the secret was required configuration that did nothing.

2. busctl quoting note — corrected in 830bbabfd3. Object paths are quoted like strings and signatures; only integers and booleans print bare. Confirmed against a live call, which returns o "/org/freedesktop/systemd1/unit/dsh_2dweb_2eservice". The module docstring did not make the claim; the _tokens docstring did, and it now states the mix the grouping argument rests on.

3. Rate limiter — bounded in 3d5f1841fe. A new key at the tracking cap sweeps every key whose attempts have all aged out, and evicts the oldest-attempt key if that frees nothing, so the map is held at the cap. Two tests cover both paths.

4. CORS and sameSite — documented in 2846da169e. The comment records that the allow-list is for same-site origins, that a cross-site caller would pass the hook and arrive without a cookie, and that sameSite has to widen before such an origin is added.

On your 'not verified' note: I ran the suites you could not. ctest, npm test (bff 32, web 9, wire-protocol 34), npm run typecheck, and the compass pytest suite (208 passed, 1 skipped) pass at this commit.

PR #2110, opened 2026-09-19
[codegen] Model every IAM NATS message and generate both targets

mcraveiro:
## Summary

IAM renders 138 protocol messages across 16 headers, but only six of them had a TypeScript twin. Five models switched the TypeScript facet off because their fields referenced C++ types with no org model: account, account_version, login_info, permission, role, session and session_statistics are hand-written structs. This PR gives every IAM message an org model and renders both targets from it, with no facet switched off and no hand-written struct on the wire.

## Changes

- Story and task scaffolding for the work, scoped to IAM for this PR: the story carries the same problem across every drift-free component, one task per component.
- An entity model now renders its protocol as TypeScript: a new ores.ts.domain facet emits the domain interface, _ts_type projects ores::<component>::domain::<entity> onto it, and the loader derives the standard CRUD message list once so both twins render from it. C++ is byte-identical.
- Junction models gain the same treatment, which removes the account_party opt-out.
- A memory records the codegen hygiene rule: generate one entity at a time through an address, never a whole component or --all.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Generate the IAM protocol messages as TypeScript](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/codegen_ts_protocol_coverage/task_implement_codegen_ts_protocol_coverage.html) | 9EE6D3B0-6C66-4DA5-B438-6F4C36192A5B |

## Testing

Plan. Codegen unit tests, the TypeScript typecheck over the generated tree, and a regeneration check that the C++ output is byte-identical.

Evidence. pytest projects/ores.codegen/tests -> 267 passed. npm run typecheck -> clean across all four workspaces. Regenerating every drift-free component's C++ leaves git diff -- '*.hpp' '*.cpp' empty. The generated IAM protocol twins were parsed and compared against the C++ headers: message names, ordered field names and subjects match exactly.

Limitations. IAM still has four opted-out protocols whose types are hand-written and unmodelled; they are the next task slice. The check that fails on an opt-out or on a header without a twin is not written yet. The generated TypeScript is typechecked but not yet exercised by the browser against a running fleet.


mcraveiro:
@claude review this PR

PR #2112, opened 2026-09-20
[codegen,sql] Honour tenant isolation in the junction template

mcraveiro:
## Summary

The market series asset class junction carried neither the row level security its sibling marketdata tables have nor a drop script. The first had a codegen cause: the junction create archetype never read the rls_tenant_isolation flag, so no junction in the tree could be tenant isolated by flag. This teaches the archetype the trailer the domain entity archetype already emits, opts the one model in, and gives the junction the drop file it lacked.

## Changes

- The junction create archetype emits the tenant isolation trailer, gated on rls_tenant_isolation, mirroring the domain entity block: enable row level security plus the permissive name_tbl_tenant_isolation_policy, with the system-tenant widening the domain template also carries.
- The market series asset class junction opts in. It is the only junction that gains a policy: the iam account party junction already carries one in iam_rls_policies_create.sql, so nothing is duplicated.
- A hand-written drop file drops the policy, rule, trigger, function and table, and marketdata_drop.sql wires it in ahead of market_series, which the junction references.
- Regenerating marketdata changed that junction create file alone, +14 lines. Component drift --all reports no drift, model drift none, populate references resolve, and the schema validator drops from 2 warnings to 0 (DROP_001 and RLS_001 on this table).
- Live database: after compass db recreate, relrowsecurity is true and the policy exists; as the marketdata service role a session claiming the system tenant sees both rows, one claiming a foreign tenant sees none, one claiming no tenant sees none, and a cross-tenant update is rejected.
- The drop chain was executed, not just read: marketdata_drop.sql inside a transaction dropped all seven marketdata tables, junction included, and the rollback restored them.
- Full preset ctest: 71 of 72 passed, including all four marketdata suites and ores.database.tests.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Complete the market series asset class junction](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/complete-market-series-classification/task_complete-the-market-series-asset-class-junction.html) | 7E064C21-14EA-401C-A73D-35EA1743EED7 |

## Testing

Plan. Prove the template honours the flag, confirm the emitted SQL is what the tree would have had if the flag had always been read, then exercise the table and its drop path against a live database.

Evidence. Schema validator: 2 warnings before, 0 after. Drift gates: no component drift, no model drift, populate references resolve. Regeneration touched that junction create file only. compass db recreate exited 0 and the FOMC populate still wrote both system-tenant junction rows under the new with check. Live RLS probe from the non-owner marketdata service role: 2 rows for the system tenant, 0 for a foreign tenant, 0 with no tenant, and a cross-tenant update rejected. The marketdata drop chain ran to zero tables inside a rolled-back transaction. Full preset ctest: 71 of 72, the one failure re-run green in isolation.

Limitations. One suite failed on the full run. ores.refdata.core.tests was terminated mid-test with truncated XML inside write_book_publishes_nats_changed_event, which is the eventing integration flake family sprint 25 already tracks rather than an assertion failure; re-run alone it passed in 262s. The run took 3506s against 1634s for the previous full run, on a host under load above 10. No C++ source changed and the component drift gate confirms the regenerated tree matches the committed one, so no fresh full build was run.


mcraveiro:
@claude review pr

mcraveiro:
@claude review PR

mcraveiro:
Both findings handled in 6891d0409a.

**1. Policy-name length guard.** Agreed, and left deferred. It is already tracked in this branch as `doc/agile/product_backlog/inbox/junction-rls-policy-name-has-no-length-guard.org`, so it will not be lost; the fix is a loader-side `rls_table_base` for junctions, derived next to `sql_name_base`.

**2. `:rls_party_isolation:` on a junction.** Accepted and fixed. The archetype now states that row level security here is opt-in and tenant-scoped only, and that the party flag belongs to the domain entity path. The note sits outside the tangled block, so re-tangling leaves every mustache byte-identical — verified by re-running the tangle and diffing `library/templates`, which shows the prose alone.

`compass lint` is clean on the new head. CI is re-running.

PR #2113, opened 2026-09-20
[codegen] Model every IAM wire type and render both targets

mcraveiro:
## Summary

Every IAM NATS message now comes from an org model and renders to C++ and TypeScript from that one source, with no facet switched off. IAM had eleven hand-written wire types whose messages could not be projected; account, login_info, permission, role and session are entity models now, account_version and session_statistics are message payloads, and the five :ores.ts.protocol.enabled: opt-outs are gone. The work also produced a current-state entity shape, a drift check that writes nothing, and five generator defects that only compiling and running the tests could find.

The same work then reached the one IAM wire type that no entity model could describe. account_party is a junction, and its messages were an operation model speaking entity-shaped CRUD vocabulary. The junction is now a model type of its own and renders the protocol itself, so the operation model and its nine messages are gone.

## Changes

- The five IAM migrations, each with the generator fix it needed kept as a separate commit: login_info (current-state shape, inet entity projection), permission (no-audit temporal, a cross-tenant soft-close defect), role (a phantom permission_codes member replaced by a standalone message), session (a compound timestamp key), account and account_version (sql_only in the mapper).
- An entity model now describes its table; a joined or denormalised shape is a message. role's permission_codes was the case that settled it, and the codes are served by a get_role_permissions message keyed by role id with no entity embedded.
- check_component_drift.py gains --dry-run: it renders into a throw-away root and reports what would change or be created, writing nothing. An --all run over a drafted model had overwritten sixteen hand-written files, which is what prompted it.
- A coverage gate fails when a protocol header in a component with committed TypeScript output has no twin, or when any model carries :ores.ts.protocol.enabled:.

### Junction protocol

The junction renders four verbs, and the pair that reads like entity CRUD is the whole reason the shape is what it is:

- get_<junction>, paged and unscoped, plus count_<junction>;
- get_<junction>_by_<side>, paged, returning a <junction>_view so the caller sees the row and what it points at;
- save_<junction> and delete_<junction>, both batch over key vectors;
- replace_<junction>_by_<side> and count_<junction>_by_<side>, both sides, offered only when the model asks for them.

link, unlink and exists are gone. save and delete already say what link and unlink meant, and exists was a membership query that the by-side count answers.

- projects/ores.iam/modeling/ores.iam.account_party_junction.org declares :list_by: and :replace_by: and switches the hand-written handler and registrar off; account_party_messages.org is deleted.
- The account_party handler, service and registrar stay hand-written, because the workflow-command wrapping lives there. A workflow step supplies its tenant through the X-Tenant-Id header and the handler scopes the context. That wrapping is orthogonal to which model owns the protocol.
- projects/ores.codegen/modeling/junction_meta_model_protocol.org records the design and what the first comparison against the operations found.

Three behavioural differences, stated rather than left to be found:

- the unscoped read pages at 100 where the hand-written handler returned every row, and reports the true total;
- the by-side read returns view wrappers rather than bare rows;
- the subjects move to the generated spelling, iam.v1.account-parties.* becoming iam.v1.account_parties.* and by-account becoming list_by_account_id. Every consumer sources the subject from the generated struct, and nothing outside the repository addresses them.

### Review follow-up

Three findings from the review on 68815431b5, each its own commit:

- **A junction batch delete silently truncated a mismatched batch.** The handler walked the two key vectors with `i < left.size() && i < right.size()`, so a pair of different lengths deleted up to the shorter length, dropped the rest and replied success. Fixed in cpp_nats_handler.hpp, so every junction gets it, plus the hand-written IAM handler whose generated facet is off. A new test renders the handler and asserts the guard. This one is worth a second look at review time because it was in a template, not in IAM.
- **`_protocol_entity_name` guessed a junction's singular with `rstrip("s")`.** `str.rstrip` strips every trailing `s`, not one. The fallback was dead code, which is why it was removed rather than tightened: the loader now refuses a junction without :name_singular: instead of guessing, and the two readers that guessed take the declared value directly.
- **`GET /api/v1/sessions/statistics` returned 200 with an empty list.** Both paths lied: the aggregate is a Timescale-licence continuous aggregate and the table the old read queried has no `create` in any schema, so a caller could not tell "no sessions in range" from "not implemented". The handler now answers 501 with the reason. No in-repo consumer calls it.

Two further observations were answered rather than changed, with reasons in the review thread: `replace_by_*` churns a version per row on every call, which is the same cost every `write` already pays because the insert trigger does the bitemporal bookkeeping; and the codegen has no watch or daemon mode for `_operation_protocol_owners`' cache to go stale under.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Generate the IAM protocol messages as TypeScript](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/codegen_ts_protocol_coverage/task_implement_codegen_ts_protocol_coverage.html) | 9EE6D3B0-6C66-4DA5-B438-6F4C36192A5B |

## Testing

Plan. Build, then the IAM and shell suites with the fleet up, the drift gate in both scopes, the codegen pytest suite, and the TypeScript typecheck over the generated tree.

Evidence. compass build green; ctest -R (iam|shell) 100% passed out of 6; ctest -R (http.api|compute) 5/5; drift clean for --component iam and --all, both dry-run; the twin coverage gate green over 19 headers; pytest projects/ores.codegen/tests 349 passed; npm run typecheck and the component build clean. The restored role permissions were confirmed through a scripted shell session, not by assertion.

Limitations. The HTTP login rejects the shell credentials in this environment, so the new REST route was verified for registration and matching but not by an authenticated call. The licence-gated Timescale retention policy is not generated; it is inert under the Apache licence this environment runs, and it is why session statistics cannot be served. session_view is declared but unreferenced, and the generated notify trigger now fires on the session heartbeat update path. The junction protocol is adopted only where a junction with a committed TypeScript output exists today, which is IAM; the hand-written junction layers in the other components are a separate pass. One environment note, unrelated to this PR: ores.marketdata.service fails its readiness wait on a restored database that is behind HEAD, so 22 of the 23 fleet units come up.


mcraveiro:
@claude review this PR

mcraveiro:
Addressed all three findings in three commits on top of `68815431b5`.

**1. Silent truncation on mismatched key-vector lengths** — `7220d03af5`

Fixed in the template, `cpp_nats_handler.hpp`, so every junction gets it. The handler now compares the two lengths and answers `bad_request` naming both before the loop:

```cpp
if (req-><left>s.size() != req-><right>s.size()) {
    BOOST_LOG_SEV(...) << msg.subject << " rejected: key vectors differ in length, "
                       << req-><left>s.size() << " and " << req-><right>s.size();
    error_reply(nats_, msg, ores::service::error_code::bad_request);
    return;
}
```

Two things worth noting beyond the finding:

- IAM's `account_party_handler.hpp` is hand-written — its generated facet is off so the workflow-command wrapping survives — so the guard is written there by hand too. A template-only fix would have left the one junction this PR is about unfixed.
- `compute`'s `app_version_platform_handler.hpp` is the only other generated junction handler with the loop, and it is regenerated. The drift gate confirms no other output moved.

New test `test_the_batch_delete_refuses_mismatched_key_vectors` renders the handler and asserts the guard is present and the `&&` loop is gone.

**2. `rstrip("s")` in `_protocol_entity_name`** — `7c65ab2cde`

You were right that it was dead code, and that is why I removed it rather than tightened it. The fallback was a wrong answer waiting for the first junction that omitted the field, so the loader now refuses such a model outright:

```
<junction>.org: a junction must declare :name_singular:. The singular names the
generated header, the C++ types and the protocol messages, and no rule derives
it from the plural name safely.
```

`_protocol_entity_name` and `core.py`'s output-path resolution now take the declared value directly, so a missing one is a loud error rather than a plausible-but-wrong name. All nine junctions in the tree already declare it. New test `test_a_junction_without_name_singular_is_rejected`; the two affected fixtures in `test_generate_component_mode.py` gained the field.

**3. Undisclosed behaviour change in `handle_get_session_statistics`** — `d9aba6a15d`

Documenting it was not enough once I looked: both paths lied. The daily aggregate is `ores_iam_session_stats_daily_vw`, created only under the Timescale licence, and `ores_iam_session_stats_tbl` — what the hand-written read queried — has no `create` in any schema, so the per-account call threw on every request while the aggregate branch returned a bare `{}`. A 200 with an empty vector cannot be told apart from "no sessions in range".

The handler now answers **501** with the reason. The endpoint has no in-repo consumer — the shell's statistics command goes over NATS, and the route is registered but otherwise unreferenced — so nothing breaks.

**On the two minor observations**

- `replace_by_account` churn: leaving it, deliberately. `write` always inserts through the mapper and lets the insert trigger do the bitemporal bookkeeping, so *every* write of an unchanged row already records a version. Diffing inside `replace_by_*` would make that one path quietly cheaper than the rest and would need a read-then-compare the repository does nowhere else. Replace is a full resync and the cost is the same cost every write pays.
- `_operation_protocol_owners`' `lru_cache`: safe. There is no watch or daemon mode — `grep -rn "watch\|daemon" projects/ores.codegen/src/codegen/*.py` finds none — and each codegen invocation is one process over one modelling directory, which is what the docstring says.

**Verification** (fleet up, all after the fixes): `compass build` green; `ctest -R "iam|shell"` 6/6; `ctest -R "http.api|compute"` 5/5; `pytest projects/ores.codegen/tests` 349 passed; `check_component_drift.py --all --dry-run` reports no drift; the twin coverage gate green over 19 headers; `compass lint` valid.

One note on the fleet, since it is the only non-green thing nearby: `ores.marketdata.service` fails its readiness wait with a pre-existing `permission denied for table ores_ore_series_key_shapes_tbl`, the restored database being well behind HEAD. 22 of 23 units came up and it is unrelated to this PR.

mcraveiro:
## Rebased onto `origin/main`, database recreated, full suite run

The branch was 23 commits behind `origin/main`. It is now rebased onto it (40 commits, no conflicts) and force-pushed as `087498f653`. `git rev-list --left-right --count origin/main...HEAD` is `0 41`.

The rebase was not cosmetic. Two of the incoming commits change the database, and neither had reached this environment:

- `5557444cdc` grants `marketdata_service` SELECT on `ores_ore_*`.
- `9b07e08d21` teaches the junction create archetype to honour `:rls_tenant_isolation:` and opts in the market series asset class junction.

### What the drift actually was

The environment's database had been provisioned from `34e64c0451` at 10:54 on 2026-09-19. The grant landed at 17:28 that day and the RLS trailer at 01:51 the next day. Neither was ever applied, because the schema is applied by a script that runs once at provisioning and generated grants are not migrations.

Two consequences, both confirmed against the live database before the recreate:

- **`ores.marketdata.service` was dead.** It aborted on startup with `permission denied for table ores_ore_series_key_shapes_tbl`. The role had SELECT on 7/7 `ores_marketdata_` tables and 89/89 `ores_refdata_` tables but 0/1 `ores_ore_` tables, and no service role had any privilege on that table. Reproduced directly with `set role ores_eager_maxwell_marketdata_service`.
- **One junction of nine had no tenant isolation.** `ores_marketdata_market_series_asset_classes_tbl` had `relrowsecurity = false` and zero policies, while the other eight junctions each had RLS and at least one policy. Every generated query in `market_series_asset_class_repository.cpp` filters `tenant_id = ctx_.tenant_id()`, so I found no cross-tenant read through the service path — what was missing is the second layer the other 222 tenant-scoped tables carry.

All six checks were green throughout, on both the old and the new head. Nothing gates on the `drift` line `compass db status` prints.

### After the recreate, on the rebased tree

```
marketdata SELECT on ores_ore_series_key_shapes_tbl : t
junction RLS, all nine:
  ores_compute_app_version_platforms_tbl          rls=t policies=1
  ores_dq_badge_mappings_tbl                      rls=t policies=2
  ores_dq_dataset_bundle_members_tbl              rls=t policies=2
  ores_iam_account_parties_tbl                    rls=t policies=1
  ores_marketdata_market_series_asset_classes_tbl rls=t policies=1   <- was f / 0
  ores_refdata_currency_currency_groups_tbl       rls=t policies=1
  ores_refdata_party_counterparties_tbl           rls=t policies=2
  ores_refdata_party_countries_tbl                rls=t policies=2
  ores_refdata_party_currencies_tbl               rls=t policies=2
```

`compass db status` now reports `drift: current`, built from `0dc039fdce`. `ores.marketdata.service` reaches `Service ready.` with no errors in its log, having previously aborted at the same point.

### Tests, fleet up, on the recreated database

- Full `ctest`: **72/72 passed**, 0 failed, 1603 s.
- `pytest projects/ores.codegen/tests`: 349 passed.
- `check_component_drift.py --all --dry-run`: no drift.
- `check_protocol_twin_coverage.py`: green over 19 headers.
- `compass lint`: valid.
- CI at `087498f653`: all six checks green.

One process note, captured as a memory in `doc/llm/memory/recreate_is_the_only_db_migration.org`: recreate from the branch under test, and rebase first, because a branch behind `main` recreates the old schema. My first recreate built from the pre-rebase head and reproduced exactly that — the grant landed and the junction stayed un-isolated.

PR #2114, opened 2026-09-20
[codegen] Render the whole refdata wire estate to both targets

mcraveiro:
## Summary

Refdata reported drift-free while nine junction models were excluded from generation, so 168 of their files had never been reconciled and none of its types reached the UI. This removes the exclusion, brings the junctions to the IAM shape, models the last hand-written protocols, and commits the TypeScript tree, so every refdata protocol header now has a twin and two gates police the component.

## Changes

- Drop the junction exclusion from refdata's catalogue entry, so discover_models and the drift gate cover all nine junctions
- Move the nine junctions' profile binding into the canonical * Flags drawer, declare has_tenant_id and give each a :list_by: read
- Make a file-level #+profile: keyword fail loudly instead of being silently ignored, and add ores.ts.protocol to the junction messaging gate so a junction renders both twins or neither
- Regenerate the nine junction C++ and SQL stacks, and delete the two bespoke calendar-assignment paths that collided with the generated protocol
- Model asset_class, calendar_adjustment and calendar_materialisation as operation (message-only) models, replacing five hand-written protocol headers
- Honour :read_only: across the protocol, service, handler and sub-registrar, and stamp the junction save from the request context
- Commit the 151-module TypeScript tree and project hierarchy_node onto a shared HierarchyNode interface
- Delete four dead hand-written domain headers and refresh compute's app_version_platform service for the shared template change

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Generate every protocol message as TypeScript](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/codegen_ts_protocol_coverage/story.html) | 18F49E14-F89E-4E01-B4B7-33928F3B094A |
| Task | [Generate the refdata protocol messages as TypeScript](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/codegen_ts_protocol_coverage/task_implement_codegen_ts_refdata_coverage.html) | 07BCBE53-701D-4517-B5B2-C707DA6B7DF2 |
| Environment | swift_curie | |

## Testing

Plan. Run the full build, the refdata and ores.ore suites, the codegen suite, the TypeScript typecheck and web suites, the drift gate over every space, the protocol-twin coverage check, the site build, and the doc, CMake-source, populate-reference and roundtrip checks.

[...]

PR #2115, opened 2026-09-20
[codegen] Close the refdata codegen follow-ups

mcraveiro:
## Summary

PR #2114 rendered the whole refdata wire estate to both targets but left four findings open, and its review round found five more. This branch closes all nine: an entity model can declare protocol messages, a junction can suppress its write surface without making its repository read-only, the junction batch save validates every row and opens one transaction, and every permission the generated handlers check is now seeded and policed by a gate.

## Changes

- Let an entity model declare protocol messages in a top-level * Messages section, render them in C++ at the protocol header's paste point and in its TypeScript twin, and move party's and counterparty's composite_as_of pairs off their paste blocks
- Split a junction's repository write surface from the verbs a client reaches with :client_read_only:, so calendar_date keeps the repository write the materialisation service needs and loses its client save and delete
- Validate every row of a junction batch save before its single write, as the single-row save and every entity batch save already did
- Derive a junction's wire_write_enabled outside the C++ drawer guard, so a junction with no drawer cannot silently lose its whole C++ write surface while its TypeScript twin keeps the verbs
- Reject :client_read_only: on a domain_entity, field group or operation, and reject a model that both declares * Messages and feeds the protocol paste point, because neither has a reader and both render as though declared
- Seed the 137 refdata permissions the generated handlers check, give the Operations role the four currency junctions (a currency's calendars, issuing countries, groups and pair conventions); the three party junctions stay with the service roles, and add check_handler_permissions.py plus its CI job so the next new resource cannot regress
- Restate the party and counterparty model prose and the client-read-only template note in the present tense

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Let an entity model declare extra protocol messages](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/codegen_ts_protocol_coverage/task_give_entity_models_declared_protocol_messages.html) | 5BF445FD-80CB-4E8E-9C79-E5FD074AE9FA |

## Testing

Plan. Run the full build, the codegen suite, the refdata and ores.compute suites against the recreated database, the TypeScript typecheck and web suites, the drift gate over every space, the protocol-twin, populate-reference and handler-permission checks, the site build, and the doc, CMake-source and roundtrip checks.

Evidence. Code and ci class. Full build green (linux-clang-debug-make); codegen suite 369 pass; refdata suites 4/4 pass (349 tests) and ores.compute suites 4/4 pass (82 tests) against the recreated database; drift gate reports no drift over all spaces; protocol-twin coverage 96/96 with no opt-out; handler-permission check reports 137 distinct codes seeded across the refdata handlers, generated and hand-written alike; populate references resolve; cmake-sources --check up to date; validate_docs OK for 42 components; site build succeeded. A local review of the branch found the batch-validation gap, the missing wire_write_enabled derivation, the silent :client_read_only:, the double-rendered messages and the unseeded-permission gap; each is fixed and pinned by a test.

Limitations. The handler-permission check covers only the components whose seeds are complete, which today is refdata alone; compute, dq, iam, marketdata, ore, reporting, synthetic and trading still check about two hundred codes no seed defines, a gap wider than this change and left to their own tasks. The :read permissions and the suppressed junctions' :delete permissions stay seeded although no handler checks them, because the catalogue names a resource's action matrix rather than the enforced set. The four remaining drift-free components still carry hand-written wire types, so the story stays open. The full ctest run was not repeated for this branch; the refdata and compute suites cover the regenerated services.


mcraveiro:
@claude Please review this PR against the repository's code review checklist. It is the follow-up to #2114: it closes the four findings that PR deferred and the five a local review of this branch raised. Please focus on whether the junction batch save's row-by-row validation matches the single-row path, whether deriving wire_write_enabled outside the C++ drawer guard can change any existing junction, whether the new loader guards can reject a model that is currently valid, and whether the handler-permission check's coverage claim holds.

mcraveiro:
Both points addressed in 8cb4909a96 and 92303dd7a9.

The .cpp blind spot was real: handler_headers() now globs messaging/*_handler.hpp and messaging/*_handler.cpp. Refdata's hand-written calendar_materialisation_handler.cpp checks refdata::calendars:write, which the generated calendar_handler.hpp also checks, so widening the glob left the gate green; test_a_hand_written_cpp_handler_is_covered pins the discovery so the file cannot drop out again. The summary line also counted (file, code) pairs, so it reported 138; it now reports the distinct count, 137.

The role-grant wording was wrong and the PR body is corrected: only the four currency junctions join the Operations role, because its description is currency management and account viewing. The three party junctions stay with the refdata service roles.

mcraveiro:
@claude review pr

PR #2116, opened 2026-09-20
[web] Drive the IAM entity screens from one declaration

mcraveiro:
## Summary

The entity shell was presentational and nothing drove it: an entity supplied its own list, detail and history pages, each wiring the same hooks with different field names. This branch makes the model the only per-entity artefact. Codegen now emits a web declaration and a BFF route descriptor per entity, three shared containers do the wiring once from that declaration, and a route factory turns a declaration into an entity route set, emitting only the routes its capabilities allow. tenant_type is the first entity through it, and the only place it is named is its descriptor reference in the registry. Five IAM entities register through the generic BFF route factory, which previously had no callers.

## Changes

- Generate the TypeScript wire protocol for the components that project and had no protocol output yet — analytics, compute, dq, marketdata, ore and reporting, 80 modules. Fix the generator defects that blocked it: a hand-written utility type gained a projection and an import, and the web facet is now withheld wherever the UI facet is withheld, which had produced six declarations of pure syntax error.
- Add a generated web declaration and a generated BFF route descriptor per entity, both from the same model, so the route, the API path, the key and the subjects are stated once and cannot be restated differently by a screen.
- Drive the shared entity screens from the declaration: three containers replace the per-entity wiring, and a route factory emits list, create, detail, edit and history only where the entity declares them.
- Register the five IAM entities that have a declaration through the generic BFF route factory, so each stops needing four handlers written out by hand.
- Report bootstrap mode before a login is attempted, with its own error code, rather than rejecting credentials in a deployment that has no accounts. This follows the Qt client, which checked bootstrap status before reaching the credential form.
- Local review fixes: a generated route is now withheld when the request it would send is keyed differently from the URL parameter, which had made a delete report success while matching no row and a history fail to decode; the generic save stamps an empty recorded_at as the country route already did; the list route validates offset and limit; and the web facet is opt in per model.
- Review round: the generated route segment is kebab-case, matching every other registry route, so tenant_type serves at `/iam/tenant-type`; and the BFF route descriptor's collection doc example now matches the snake_case the codegen actually emits.
- Gate scoping: the codegen gates take their components from one shared list, `COMPONENTS_UNDER_TEST` in `component_registry.py`, and it names IAM alone. The twin gate scoped itself by committed output, so committing a first twin for a component enrolled it and then failed on the hand-written protocol headers that have no org model to project from; every other component is now explicitly to-do. The permissions gate keeps its own list, because IAM's generated handlers check thirteen permission codes that `iam_permissions_populate.sql` does not seed.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Implement every IAM entity in the web client](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/web_iam_entities/story.html) | 131C14D1-31D1-46EE-B697-B52EE3612F5F |
| Task | [Drive the entity screens from one declaration](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/web_iam_entities/task_entity-shell-from-declaration.html) | FEA81310-1D51-489D-8877-9362B74C191F |

## Testing

Plan. Rebuild at HEAD, recreate the database from the current sources, bring the fleet up, and probe the routes over HTTP. Route registration is observable without a session: an unregistered route answers 404 and a registered one answers 401.

Evidence. Full build exit 0 and 24/24 units running; database recreated from HEAD in 1m05s with drift current. Codegen 397 passed, wire-protocol 36 passed, web 11 passed, typecheck 0 errors. The twin gate reports all 19 IAM headers twinned with no opt-outs, the drift gate reports no drift over IAM's ten addresses, and `compass lint` passes. CI is green on all six checks. Live probes confirm tenant and account_contact_information serve list and save with delete and history withheld (404), while tenant_type, account_type and tenant_status serve all four (401 unauthenticated).

Limitations. The entity protocol defines no single-record read, so the shell loads a record from the list under a ceiling; a collection larger than that ceiling would answer not found for a record past it. Writes cannot be exercised end to end because the deployment is in bootstrap mode and refuses every login, and the bootstrap wizard that would provision it does not exist yet. The BFF generic save does not stamp tenant_id or performed_by; the server owns them in stamp(), and the wire request still carries them, which is the subject of follow-up analysis. The drift and twin gates now verify IAM only: every other component is explicitly to-do until it joins `component_registry.py`, so their generated output is unverified in the meantime. One commit in this branch carries a Task-ID that does not exist in doc/agile, and one commit message overstates its own module count; both were left uncorrected under the never-amend rule.


mcraveiro:
@claude review PR

mcraveiro:
Both findings applied in b4d443ba7e.

Route segment: web_declaration_projection now derives the segment through _ui_kebab, so tenant_type routes at /iam/tenant-type. The entity name, the label-key prefix and the API collection keep snake_case; only the address changes. registry.test.ts pins the wired entity's segment so the convention cannot drift back, and the codegen projection test expects book-status. Verified: 397 codegen tests, 11 web and 36 wire-protocol tests, zero typecheck errors across four workspaces, drift gate clean.

Collection doc: the example is now tenant_types, matching /api/tenant_types beside /api/countries.

Your read on the irony was right — the test was using the value the apps own convention would produce as its mismatch case. It now uses tenant-types. The template tangle and no-drift checks were re-run after regeneration.

PR #2118, opened 2026-09-21
[doc] Specify the canonical NATS entity protocol

mcraveiro:
## Summary

ORE Studio had no specification for its common entity NATS messages: the page every other page linked to as the protocol reference was a subject namespace index, the message shapes lived in a codegen archetype, and the only written statement of the response convention sat inside a closed task. This branch writes the specification, gives the messaging pages a hub, and records the rule that its summaries had to be rewritten twice to satisfy.

## Changes

- Add the NATS entity protocol specification: the subject grammar, eight orthogonal verbs, the request and response envelopes with a closed set of outcomes, keys, pagination and filtering, the write model, the concurrency precondition, the event contract, the security rules, versioning, and the reserved namespace for domain operations.
- Make batching coherent. Single and batch are paired, so no entity is batch-less; an element carries its own precondition while the batch stays one transaction; a batch read answers one entry per requested key in the order asked; and batch and page bounds are part of the protocol.
- Read versions and subtrees as collections rather than verbs. Versions become a read-only sub-resource addressed by a composite key, so a single version is directly addressable for the first time and the version range is a filter; a subtree becomes the relation-scoped list with a subtree scope, so the reply is entities and the caller assembles the tree. The verb set falls from ten to eight.
- Add the Message Queue structure note and wire it into the knowledge index, both in the reading-order table every structure note appears in and in a new Infrastructure section.
- Record the knowledge-page summary rule as a memory, after the hub summary was rejected twice: once for narrating mistakes and describing the cluster, and once for stating facts about one page instead of placing the subject.
- Rename the wire format page to NATS Wire Format and update the four pages that linked it by its old title.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Scaffold story: Define and adopt the canonical NATS entity protocol](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/canonical_nats_entity_protocol/task_scaffold_canonical_nats_entity_protocol.html) | 07056F20-C09B-4F6F-B550-C52E99A7A2D9 |

## Testing

Plan. Read the specification as a stranger and check every clause: each either states a rule a reviewer can apply or is mechanically checkable, which is what the conformance task will encode. Rebuild the site and read the published pages rather than the org sources, since that is what a reader sees.

Evidence. compass lint passes. The site rebuilds with zero failures and every changed page serves 200 on the preview, including the specification, the hub, the knowledge index and the code review checklist. Both summaries were re-read on the published HTML.

Limitations. No code is generated from the specification yet, so conformance is stated rather than enforced: the conformance gate is its own task and until it lands nothing fails when a generated message departs from the document. Every component other than ores.iam still speaks the shapes the specification replaces. And the backlog captures the protocol absorbs are named in the task notes but not yet re-pointed, which the reconciliation task owns.


mcraveiro:
@claude review PR

mcraveiro:
Both findings applied in 2457c371b4, on the branch that carries the IAM adoption, since this PR had already merged.

Events: the example was five segments while the grammar states that every subject has exactly four, and the sentence under it claimed the example was four. An event is now {component}.v1.{resource}_events.{action} -- four segments, the resource segment naming the events collection in the {resource}_{collection} form the sub-resource rule already uses, and the action in the last slot. The Subject grammar section now states the request and event forms side by side, and the verb-suffix rule is scoped to request subjects, because an event reports what happened rather than performing an operation.

get_many: the key set now carries the batch bound of 1000, stated by reference to the Batching section, so the conformance check has a number to enforce rather than a phrase to interpret.

Your read on the archetype was right, and worth recording here: ores.cpp.protocol.protocol_header.org still emits .list/.save/.delete/.history/.hierarchy with success/message responses. Changing it to this specification is the codegen work that starts now.

PR #2119, opened 2026-09-21
[database] Remove self-referential namespace using-directive

mcraveiro:
## Summary

The Windows MSVC build fails in ores.database: database_info_mapper.cpp and database_info_repository.cpp open namespace ores::database::repository and then pull in that same namespace with a using-directive. MSVC reports C4515 and /WX promotes it to the error C2220. The directive is redundant, so removing it changes no behaviour. Both files are generated from the database codegen component, so the templates carry the same defect.

## Changes

- Remove the self-referential using-directive from database_info_mapper.cpp and database_info_repository.cpp.
- Guard the four template sites that emit the directive with component_is_database, so the database component omits it and every other component keeps byte-identical output.
- Expose component_is_database on entity models in ores.codegen core.py.
- Re-tangle cpp_domain_type_mapper.cpp.mustache and cpp_domain_type_repository.cpp.mustache from their .org sources.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Hotfix: database_info mapper and repository use their own namespace](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/database_namespace_self_reference/story.html) | DE9D2DD4-9DB6-48E1-946C-6CAB97D931C1 |
| Task | [Scaffold story: Hotfix: database_info mapper and repository use their own namespace](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/database_namespace_self_reference/task_scaffold_database_namespace_self_reference.html) | A0393C39-A5FA-4D47-8CD9-6D3DB811A2F4 |

## Testing

Plan. Build ores.database.lib, regenerate a non-database entity, and render the guarded template for the database component.

Evidence. compass build ores.database.lib is green. Regenerating ores.refdata.asset_class_code leaves no diff, so non-database output is unchanged. Rendering the guarded template with component database omits the directive, while refdata and assets still emit it. The codegen test suite passes (397 passed).

Limitations. The C4515 warning itself is MSVC-only, so this host cannot reproduce it. Verification is by construction: the self-referential line is gone, the remaining usings are unaffected, and the fix matches a compiler-independent bug.


mcraveiro:
@claude review PR

mcraveiro:
Fixed in 34edc74c6e — added projects/ores.codegen/tests/test_repository_namespace_guard.py. It renders both template families twice: for a database component it asserts the self-referential using-directive is absent and the enclosing namespace is ores::database::repository, and for another component it asserts the directive is still emitted. I also confirmed the test fails when the guard is removed, so it pins the seam rather than the current tree. Full codegen suite: 402 passed.

PR #2120, opened 2026-09-21
[iam,codegen,shell] Adopt the canonical NATS entity protocol in IAM

mcraveiro:
## Summary

Converts ores.iam to the canonical entity protocol and proves the generated shell surface against the running fleet. Codegen emits the specification's message set for every entity, junction and operation; the shell commands generated from it cover every verb; and the recipes that exercise them are now generated from the same models rather than written by hand.

## Changes

- Codegen emits the canonical message set per resource, and every message carries its subject as a constant: 21 IAM protocol headers with TypeScript twins.
- Junctions render shell command units like entities. account_party gains nine commands, including the delete that makes revoking an association possible.
- A new codegen facet generates a literate shell recipe per model, one section per command, each exporting its own script into the shell library: 20 documents, 124 scripts.
- The shell recipe inventory is generated and gated; it held 97 entries for 122 recipes before this change.
- 23 raw subject literals became generated constants, and two operations that nothing served were closed: account permissions is implemented over NATS, session statistics is removed.
- reset reset-system is declared destructive in its own model, so an unattended replay can refuse it rather than discover the fact by running it.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Adopt the canonical protocol in ores.iam](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/canonical_nats_entity_protocol/task_adopt_the_protocol_in_iam.html) | D9A4776C-AB42-4461-8F2F-AD077C20716E |

## Testing

Plan. Build the whole tree and run the full ctest suite locally. Then start the service fleet and run every generated shell command against it, one script per command, classifying each reply as answered or not.

Evidence. Build clean at 100%. ctest 73 of 73 targets passed (1808s). 560 codegen tests pass. Component drift, model drift, protocol twin coverage, handler permissions and populate references all clean. Fleet up and serving. All 124 generated IAM commands have a recipe; 123 were run and all 123 answered (65 ok, 49 domain refusals, 9 replies in another shape).

Limitations. reset reset-system is deliberately not replayed, because it hard-deletes non-system tenants. Generated scripts send type-correct sentinel values that address no row, so a write is proved to reach the handler and be refused on its merits, not to mutate correctly. The ores.web half of the verification is PR 2. One subject literal remains in ores.nats, where referencing IAM's constant would invert the layering.


mcraveiro:
@claude review PR

mcraveiro:
Round handled in `0597a51525` and `482facf711`. Six accepted, one declined. One correction.

**1. `:destructive:` message** — accepted. It names the three values that mean yes.

**2. `event_action` dead code** — accepted, removed. Confirmed dead across the whole tree, not just `.hpp`/`.cpp`: it appears only in its own header and in a test *name*. It was added by this branch (`d6701b43b6`), never wired and never tested. I chose removal over typing because making the action field an enum is a change to the event protocol — JSON serialisation, the generated `*_event` types, the registrars — and not a cleanup to smuggle into this PR.

**3. `read_for_cache` prose** — accepted, but the premise was off, and it is worth knowing why: the `{{#read_for_cache}}` section is still live. `ores.refdata.party` and `ores.refdata.currency_pair_convention` both set the flag, so it is vestigial rather than dead — it now emits the two includes the removed handler method needed, and no generated handler uses them. The prose says that, and says plainly that removing the flag is a change to those models and their generated headers rather than a line deleted here. I did not do that in a PR scoped to IAM.

**4. Inventory script on a missing index** — accepted. It names the path it looked for; a test pins it.

**5. `ts_protocol` duplication** — declined. The three blocks mirror the three blocks in the C++ protocol template, which are already separate, and they are expected to diverge: an operation has no key record and no versions sub-resource. A partial would hide which block a change lands in.

**6. `gather()` three times** — accepted. It walks once.

**7. Stale namespace in the doc example** — accepted. The generated type lives in `ores::iam::messaging`.

On your process caveat: you noted the sandbox could not run `compass`, `git fetch` or `gh`, so this was a static read. That is exactly why finding 3's premise slipped — checking it against the tree took one grep for `read_for_cache: true` in the models. Worth flagging for whoever reads this next: the review is useful, but its "this was removed" claims need a grep before they are acted on.

Verification for this round: build clean at 100%; 561 codegen tests pass; `ores.eventing.api.tests`, `ores.eventing.core.tests` and `ores.iam.api.tests` pass; component drift, model drift, twin coverage, handler permissions and populate references clean; and all six doc-lint steps pass, including the skills type tables.

PR #2123, opened 2026-09-22
[sql] Wire the IAM current-state notify triggers into the aggregators

mcraveiro:
## Summary

SQL schema validation has been failing on main since at least 2026-09-21 with four WIRE_001 warnings, which keeps Continuous Linux and the Nightly Build red. The login_info and session notify trigger scripts exist but are not reachable from the create and drop aggregators. These are the only two :current_state: true entities in ores.iam and the only two whose generated triggers were never wired in; the other nine IAM notify triggers are already included. The aggregators are hand-maintained files, so codegen emitted the trigger scripts and nothing added the includes.

## Changes

- Add the iam_login_info_notify_trigger_create.sql and iam_sessions_notify_trigger_create.sql includes to create/iam/iam_create.sql.
- Add the matching drop includes to drop/iam/iam_drop.sql, each immediately after its own table's drop.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Story | [Hotfix: IAM current-state notify triggers are not wired into the aggregators](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/iam_notify_triggers_unwired/story.html) | 54897BD4-A38C-4271-AA3B-CE79C43663E3 |
| Task | [Scaffold story: Hotfix: IAM current-state notify triggers are not wired into the aggregators](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/iam_notify_triggers_unwired/task_scaffold_iam_notify_triggers_unwired.html) | B2B26F3E-59F2-437B-BE0F-4F03245347B0 |

## Testing

Plan. Run the SQL schema validator that CI runs, and confirm it reports no warnings.

Evidence. validate_schemas.sh reports 0 warnings and Validation Passed, against 4 warnings before the change. This is the exact script and flag set (--validate-only --strict) that Continuous Linux and the Nightly Build run, so the check is the same one that was red.

Limitations. The validator is static; it proves reachability, not that the scripts run clean against a live database. WIRING is the only property it checks, and it was the property failing in CI.


mcraveiro:
@claude review PR

mcraveiro:
Fixed in 1515844bfd — swapped both pairs in projects/ores.sql/drop/iam/iam_drop.sql so the notify trigger drop precedes its own table drop, matching every other entry in the file. The validator still reports 0 warnings and Validation Passed. Recorded in the task Review table.

PR #2127, opened 2026-09-22
[refdata,web,codegen] Adopt the canonical protocol in the refdata web layer

mcraveiro:
## Summary

The refdata web layer now speaks the canonical protocol. The BFF route factory and the browser's generic screens were built against the protocol before the canonical verb set, so this fixes that shared layer once and then moves the country screens onto the generated declaration. It also completes the shared pieces the IAM web layer needs: canonical envelopes, protocol-derived capabilities, a write record every entity can decode, and the flag the generated metadata now expresses.

## Changes

- The BFF route factory builds the canonical request for every verb and answers from result.outcome; a create states must_not_exist and an amend must_match_version.
- The generated route descriptor states the write record's members and per-member defaults, the intent fields, and whether the list and versions requests take an as-of or a filter.
- A route the natural-key path segment cannot drive is withheld, so the single-record read joins the delete and versions reads; no route fills a key record member its request does not carry.
- The web declaration's history capability is read from the derived versions pair rather than the retired history_message_type property.
- An entity that declares a flag states its image in the generated UI metadata, and the shared detail screen renders the picker and saves the chosen image.
- The country model opts into ores.ts.web: the route descriptor and web declaration are generated, and the hand-written schema, api client, four subjects, five routes and the bespoke screens are deleted.
- The technical spaces document names the two generated TypeScript facets it was missing, and the refdata api CMake list drops the umbrella header the conversion removed.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Adopt the canonical protocol in the refdata web layer](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/canonical_nats_entity_protocol/task_adopt_the_protocol_in_the_refdata_web_layer.html) | 0388AB16-48B0-494B-8373-8ABF695C76E2 |

## Testing

Plan. Run the codegen suite and every codegen gate, then the web typecheck, build and suites; build the site.

Evidence. Codegen 573 passed; drift gate clean for iam and refdata; protocol twin coverage 98/98; model drift clean; handler permissions seeded; populate references resolve; CMake component files up to date; shell recipe inventory current; web typecheck and build clean; BFF 32, browser 16 and wire-protocol 36 tests pass; site page build 184 files, 0 failures.

Limitations. No test drives the BFF route factory itself, so the canonical envelopes are pinned through the descriptor projection tests and the typecheck rather than by a route-level test; the screens were exercised by the typecheck, suites and site build, not against a running service. A write member that names another row and that the form does not show takes no value rather than a minted one, so an entity whose screen must create such a row has to carry the field in its presentation drawer. The country list's as-of query survives at the BFF, but the shared list screen does not yet offer the control.


mcraveiro:
@claude review PR

PR #2130, opened 2026-09-22
[codegen,web] Serve the screen sets a read-only entity declares

mcraveiro:
## Summary

calendar_date is a materialised junction whose protocol derives reads and no writes. The route projection demanded a list, a save and a removal together, so the facet emitted a declaration and no route, and the model opted back out -- which left a real screen unbuilt. This serves a screen set from the verbs the model actually derives.

## Changes

- The route projection needs the list alone and states the write record, the save subject and the intent only when the model derives them.
- The factory registers the save route when the descriptor carries one, and the descriptor's write members are optional.
- The declaration's create and edit affordances follow the same verbs, so a read-only entity offers no form the service has no handler for.
- The route archetype admits junctions, which its projection always handled; the declaration had already admitted them, and the two must agree or a screen has no route.
- calendar_date opts in again: it now has a list screen with its own words, and iam and refdata hold 69 declarations and 69 routes with no orphans.

## Traceability

| Artefact | Link | ID |
|----------|------|----|
| Task | [Add screens for every refdata entity](https://orestudio.github.io/OreStudio/doc/agile/versions/v0/sprint_25/canonical_nats_entity_protocol/task_add-screens-for-every-refdata-entity.html) | D2DBA1FB-DA28-4D67-8267-551D3EE801FD |

## Testing

Plan. Run the codegen suite and gates, then the web typecheck, build and suites.

Evidence. Codegen 575 passed (a read-only entity case added); the drift gate clean for iam and refdata; web typecheck and build clean; BFF 47, web 23 and wire-protocol 36 tests pass; 69 declarations and 69 routes, no orphan.

Limitations. The read-only screen is list and read only: calendar_date has no save, delete or versions route because its protocol derives none, and its key is compound so no single-record route exists. It was exercised by the typecheck, the suites and the drift gate, not against a running service.


mcraveiro:
@claude review PR

